PR #67, opened 2019-11-25
Add custom range split generation logic based on primary key of table.

abhijeetvg:
*Description of changes:*
Adds a custom split generation logic based on table primary key. Creates basic equal ranges to generate expected number of splits. This logic works best when data is not skewed. 

This is a first version so has some limitations:

1. Number of splits is static, will be taken as input later.
2. Supports Integer splitter only. Can be expanded later to other data types.
3. Does not work for composite primary keys.

### DIFF
--- a/athena-jdbc/src/main/java/com/amazonaws/connectors/athena/jdbc/splits/SplitterFactory.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/connectors/athena/jdbc/splits/SplitterFactory.java
@@ -22,6 +22,7 @@ package com.amazonaws.connectors.athena.jdbc.splits;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
+import java.util.Optional;
 
 /**
  * Creates splitter depending on split column data type.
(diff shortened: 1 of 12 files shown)


athena-jdbc/src/main/java/com/amazonaws/connectors/athena/jdbc/splits/SplitterFactory.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public Splitter getSplitter(final String columnName, final ResultSet resultSet, final int maxSplits)
+            throws SQLException
+    {
+        int type = resultSet.getMetaData().getColumnType(1);
+        switch (type) {
+            case Types.INTEGER:
+                return new IntegerSplitter(new SplitInfo<>(new SplitRange<>(resultSet.getInt(1), resultSet.getInt(2)), columnName, type, maxSplits));
+            default:
+                throw new RuntimeException("No splitter found for type " + type);
```
atennak1: Should return Optional.empty() so consumers don't need to know which types are supported or catch-swallow exception.

Issue #70, opened 2019-11-28
ThrottlingInvoker to conservative with cloudwatch-metrics


avirtuos:
When testing the cloudwatch-metrics connector I noticed it was too easy for me to write a query that would result in thrashing against cloudwatch getMetricData(...) API. This was surprising to me because ThrottlingInvoker was indeed kicking in and reducing the TPS to 1 for each running Lambda. However, it seems as though in aggregate Athena's minimum concurrency when peak congestion control active was still high enough to overwhelm the typical cloudwatch-metrics account limits. As such this task is to:

1. Reproduce the issue and confirm the above behavoir.
1. Use lower (bigger back off, less frequent calls) for the ThrottlingInvoker used by cloudwatch-metrics. It currently uses the default ThrottlingInvoker config. Don't change the default, just use the builder in the cloudwatch-metrics connector to build a custom one for cw metrics.
1. Update the README for the connector to recommend using a lower default concurrency in the Lambda console for this connector so as to override Athena's minimum concurrency.
1. Update our docs to have a better example query which is less likely to catch dozens of individual metrics and thus produces lower concurrency.
1. We should also consider changing how we form splits so that we make fewer calls to GetMetricData by pulling up to 30 metrics in each call as recommended by CW team

atennak1:
https://github.com/awslabs/aws-athena-query-federation/pull/71 address 1, 2, and 5.  3 and 4 are probably no longer be necessary so resolving this.

PR #71, opened 2019-12-02
fixing ThrottlingInvoker to update state before throwing and making l…

atennak1:
…ess/bigger splits in CloudWatch Metrics connector to improve performance

*Issue #, if available:* 70

*Description of changes:* 2 primary changes:  
1. ThrottlingInvoker was throwing back to Athena before updating its state.  This was preventing effectual throttling because Athena's congestion control was still insufficient to reduce enough calls to GetMetricData which has a fairly low limit (50 TPS).  Fixed it to throw *after* updating its state.
2. GetMetricData can take up to 100 "metrics", where each "metric" is uniquely identified by a quad of Namespace, List<Dimension>, MetricName, Statistic.  So to reduce the number of GetMetricData calls I also adding batching of metrics instead of one-by-one submission.  Also made MetricsRecordHandler use a ThrottlingInvoker with tweaked settings to be more aggressive with throttling.

*Testing:* tested in my account with various queries.  Most queries took less than 1/4th the time.


avirtuos:
Related to #70

athena-cloudwatch-metrics/src/main/java/com/amazonaws/athena/connectors/cloudwatch/metrics/MetricsRecordHandler.java
```diff
@@ -90,7 +93,7 @@
     private static final Table METRIC_DATA_TABLE = new MetricSamplesTable();
 
     //Used to handle throttling events by applying AIMD congestion control
-    private final ThrottlingInvoker invoker = ThrottlingInvoker.newDefaultBuilder(EXCEPTION_FILTER).build();
+    private final ThrottlingInvoker invoker = ThrottlingInvoker.newDefaultBuilder(EXCEPTION_FILTER).withInitialDelayMs(140).withIncrease(20).build();
```
sethtdenney: I'd recommend creating constants for these two throttling constants with a comment about how they were derived.

athena-cloudwatch-metrics/src/main/java/com/amazonaws/athena/connectors/cloudwatch/metrics/MetricsRecordHandler.java
```diff
@@ -192,19 +195,21 @@ else if (field.getName().equals(DIMENSION_VALUE_FIELD)) {
     private void readMetricSamplesWithConstraint(BlockSpiller blockSpiller, ReadRecordsRequest request, QueryStatusChecker queryStatusChecker)
             throws TimeoutException
     {
-        Split split = request.getSplit();
-        List<Dimension> dimensions = DimensionSerDe.deserialize(split.getProperty(DimensionSerDe.SERIALZIE_DIM_FIELD_NAME));
+//        List<Dimension> dimensions = DimensionSerDe.deserialize(split.getProperty(DimensionSerDe.SERIALZIE_DIM_FIELD_NAME));
```
sethtdenney: This line should be removed instead of simply commented out.

athena-cloudwatch-metrics/src/main/java/com/amazonaws/athena/connectors/cloudwatch/metrics/MetricsRecordHandler.java
atennak1: done

athena-cloudwatch-metrics/src/main/java/com/amazonaws/athena/connectors/cloudwatch/metrics/MetricsRecordHandler.java
atennak1: done

Issue #78, opened 2019-12-11
Handle Glue table name coercion


atennak1:
When Glue Crawler crawls a table like a DynamoDB table, it will replace hyphens in the the table name with underscores.  Also it is likely that Glue will not allow hyphens during manual table creation.  Our current implementations of GlueMetadataHandler in all the current connectors currently assume the table name will be the same between Glue and the data source (allowing for case insensitivity).  I can think of three ways to address this:

1. Crawled tables maintain the original name of the table in the table's StorageDescriptor->Location property.  We can leverage this if present and in a known form (like an AWS ARN).
2. In the Table "Resolver" that some connectors have, we can explicitly fall back to replacing underscores with hyphens if a table lookup fails.  This can get messy and expensive if there's a mix of hyphens and underscores in the actual table name.
3. We can add an environment variable that can be used to provide a manual mapping of Glue table name to target table name.

I'm leaning toward a combination of 1 & 3.

avirtuos:
Instead of #3 I'd recommend a table property in glue.

houmark:
Hey guys,

My feedback here (I was hit by this).

This will affect all AWS Amplify (and possibly also AWS AppSync) generated DynamoDB tables, as Amplify creates multiple environments and a unique identifier on each table in addition to its name. Each part is divided by a hyphen.

If option 1 could be done first, then I think this will solve most of these situations. When I created the Glue crawler against my table, it automatically set the ARN of the DynamoDB table in the Location property, and that seems to be a safe mapping for the connector.

Then if someone wants or needs to override it, option 3 with a table property could be good, but I'm not sure when that would be needed if option 1 is implemented.

atennak1:
Merged with https://github.com/awslabs/aws-athena-query-federation/commit/d3b25cb6bef865f51cfd5e19e9d9512b29f48015

Issue #81, opened 2019-12-12
DynamoDB Connector does not handle Glue BigInt column correctly


atennak1:
I crawled a table containing a year column which the Glue crawler decided was a `bigint`.  I see the following stacktrace in CloudWatch:

```
Error while processing field year: java.lang.RuntimeException
java.lang.RuntimeException: Error while processing field year
	at com.amazonaws.athena.connectors.dynamodb.DynamoDBRecordHandler.lambda$readWithConstraint$1(DynamoDBRecordHandler.java:185)
	at com.amazonaws.athena.connector.lambda.data.S3BlockSpiller.writeRows(S3BlockSpiller.java:174)
	at com.amazonaws.athena.connectors.dynamodb.DynamoDBRecordHandler.readWithConstraint(DynamoDBRecordHandler.java:146)
	at com.amazonaws.athena.connector.lambda.handlers.RecordHandler.doReadRecords(RecordHandler.java:191)
	at com.amazonaws.athena.connector.lambda.handlers.RecordHandler.doHandleRequest(RecordHandler.java:157)
	at com.amazonaws.athena.connector.lambda.handlers.CompositeHandler.handleRequest(CompositeHandler.java:135)
	at com.amazonaws.athena.connector.lambda.handlers.CompositeHandler.handleRequest(CompositeHandler.java:100)
Caused by: java.lang.RuntimeException: Unable to set value for field year using value 1933
	at com.amazonaws.athena.connector.lambda.data.BlockUtils.setValue(BlockUtils.java:330)
	at com.amazonaws.athena.connector.lambda.data.Block.offerValue(Block.java:175)
	at com.amazonaws.athena.connectors.dynamodb.DynamoDBRecordHandler.lambda$readWithConstraint$1(DynamoDBRecordHandler.java:176)
	... 6 more
Caused by: java.lang.ClassCastException: java.math.BigDecimal cannot be cast to java.lang.Long
	at com.amazonaws.athena.connector.lambda.data.BlockUtils.setValue(BlockUtils.java:288)
	... 8 more
```

This is because the ItemUtils.toSimpleValue util that is used to convert from AttributeValue to Java object returns a BigDecimal for any DDB Number attribute.

atennak1:
Handled for top level (non-nested) DDB attributes with https://github.com/awslabs/aws-athena-query-federation/commit/d3b25cb6bef865f51cfd5e19e9d9512b29f48015

avirtuos:
[...]

PR #86, opened 2019-12-20
Adding source table name and column name mapping for Glue handlers

atennak1:
*Issue #, if available:* https://github.com/awslabs/aws-athena-query-federation/issues/78

*Description of changes:* Adding source table name and column name mapping for Glue handlers.  Data sources such as DynamoDB allow more characters and upper case in table names while Glue does not, meaning that it may not be possible for customers to make a Glue table with the same exact name
as their data source's table.  This allows customers to add a table property in Glue with the source table name.  This can also extract the source table name from an ARN populated in the Glue table's 
StorageDescriptor.Location property (Glue crawlers populate this).  It also enables a table property for column mapping.

I adapted the Dynamo connector to use this.  The other connectors expose table names directly from the source (DocDB and HBase) or don't have sources tables (Redis).

*Testing*: Unit tests and empirical testing with an uploaded Dynamo connector and a crawled table with different name than the source.

### DIFF
--- a/athena-dynamodb/README.md
+++ b/athena-dynamodb/README.md
@@ -28,11 +28,15 @@ question.  These properties are automatically set if you use Glue's DynamoDB Cra
 the columns and types that the Crawler discovered.
 
 1. **dynamodb** - String indicating that the table can be used for supplemental meta-data by the Athena DynamoDB Connector. This string can be in any one of the following places:
-    1. in the table properites/parameters under a field called "classification" (exact match).
+    1. in the table properties/parameters under a field called "classification" (exact match).
     2. in the table's storage descriptor's location field (substring match).
     3. in the table's storage descriptor's parameters under a field called "classification" (exact match).
 2. **dynamo-db-flag** - String indicating that the *database* contains tables used for supplemental meta-data by the Athena DynamoDB Connector.  This is required for any Glue databases other than "default"
 and is useful for filtering out irrelevant databases in accounts that have lots of them.  This string should be in the Location URI of the Glue Database (substring match).
+3. **sourceTable** - Optional table property/parameter that defines the source table name in DynamoDB.  Use this if Glue table naming rules prevent you from creating a Glue table with the same name as
+your DynamoDB table (e.g. capital letters are not permitted in Glue table names but are permitted in DynamoDB table names).
+4. **columnMapping** - Optional table property/parameter that define column name mappings.  Use this if Glue column naming rules prevent you from creating a Glue table with the same column names as
+your DynamoDB table (e.g. capital letters are not permitted in Glue column names but are permitted in DynamoDB column names).  This is expected be be in the format `col1=Col1,col2=Col2`.
 
 
 ### Required Permissions
--- a/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
+++ b/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
@@ -229,7 +229,8 @@ public class DynamoDBMetadataHandler
                 return super.doGetTable(allocator, request);
             }
             catch (RuntimeException e) {
-                logger.warn("doGetTable: Unable to retrieve table {} from AWSGlue in database/schema {}", request.getTableName().getSchemaName(), e);
+                logger.warn("doGetTable: Unable to retrieve table {} from AWSGlue in database/schema {}",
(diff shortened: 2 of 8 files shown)


avirtuos:
Maybe we should add this to the GlueMetadataHander instead so that all
connectors that use Glue can use a table property to override the
tablename?

atennak1:
@avirtuos I did :) Sorry if the description was misleading

avirtuos:
Ugh, I can't read lately. :)

houmark:
@atennak1 This looks like something I should be testing :)

I'll try merging this fix with the other fix I am already testing, and provide feedback.

Thank you!

atennak1:
Yes please do. Thanks @houmark !

houmark:
I had a chance to test this last night. I did a merge of this branch into the existing fix I had for the empty structs, so this test was based on both of these in once.

While I no longer get errors, there's something going on with selecting from some fields in the glue based table. I have some sensible columns called `firstname` and `lastname`, and those columns return empty on any query. This seems to affect most of my columns, but not all. I tried sorting on a string/date column, which normally works, but it also seems to return the correct result.

In addition I did a `count(*) as total` with a `GROUP BY` and that returned zero rows, while the non-glue based table works perfectly.

When the 404 error was gone, I was sure it would just work, as to my understanding, the glue connector/mapping, is just a mapping, but for some reason things go terribly wrong.

Just to make sure, I did a re-crawl with my existing crawler, but the results were the same. Checking the schema, the fields exist and are strings in the mapping (they show up as varchars in Athena).

Since I'm getting no errors on this one, but just empty columns, I'm unsure on how to proceed here. Anything you need, let me know.

atennak1:
Thanks for testing @houmark ! Does the casing of the columns in Glue match
DynamoDB? I noticed that glue does not allow capital letters in column
names. If the casing does not match then the connector currently assumes
that there is no data for that column. If this is the case for you I'm
thinking we may need to add some capability for case insensitive
matching/mapping.

houmark:
@atennak1 Yes, that makes total sense. My columns/fields in DynamoDB are `camelCase` and when using Glue they are lowercased in the Schema, and shows up in the column list as lowercased, but data can't be fetched in a query. I just tested now, and the columns that has only lowercase in DynamoDB are working correctly and fetches data, while the camelCased ones return nothing but also doesn't give any error.

With the interference schema it will show the column names camelCased but it does not really seem to care if I use lowercase or camelcase in the query.

atennak1:
@houmark I've pushed a commit to handle the column problem as well.  Give it another try.  This will require you to add a table property called `columnMapping` that looks like `mycolumna=myColumnA,mycolumnb=myColumnB`. This seemed like the best way to handle this without hurting performance.

houmark:
Thanks for the quick update. I've tested the column mapping, and had half success half failure.

I mapped some of the camelCased fields, some were string/varchar and seems to work, but I also have some bigint fields which I also mapped, but I now get an error when trying to query those fields. They are in the columns/fields list in both Glue and Athena and it seems to be happening specifically to all bigint fields. It also doesn't matter if the field is queried all lowercase or camelCased:

```
Your query has the following error(s):

GENERIC_USER_ERROR: Encountered an exception[java.lang.RuntimeException] from your LambdaFunction[arn:aws:lambda:us-east-1:ano:function:myfunction] executed in context[S3SpillLocation{bucket='my-athena-spill', key='athena-spill/d928cd63-dd3f-4a21-9574-ano/ano-cea6-41e7-a57d-ano', directory=true}] with message[Error while processing field totalRevenue]

This query ran against the "default" database, unless qualified by the query. Please post the error message on our forum or contact customer support with Query Id: d928cd63-dd3f-4a21-9574-c01612726202.
```

So close :)

atennak1:
Ah that looks related to https://github.com/awslabs/aws-athena-query-federation/issues/81 . I don't expect it to be a difficult fix but for now you can get around it by making those columns `decimal` in Glue.

atennak1:
@houmark it was easy enough so I pushed a quick commit for the bigint problem.  But it will only handle top-level bigints (not nested in arrays or structs).

houmark:
Brilliant! That works! I'm still testing all fields, but it seems like this maps everything out correctly.

Not really related, but it may actually be best for me to change those fields to decimal, because I have money values saved as cents, with the purpose of dividing by 100. When doing that with the interference schema, I get a decimal and the correct decimals. When it's a `bigint`, I lose the decimals and I'll probably have to `CAST` to decimal to keep them anyways, so I'm considering what's the best option.

avirtuos:
@atennak1 Can you take a look at the merge conflicts? I've approved the changes so you can resolve conflicts and merge at your discretion.

Issue #89, opened 2020-01-07
Address Log4J vulnerability


atennak1:
https://github.com/awslabs/aws-athena-query-federation/network/alert/athena-federation-sdk-tools/pom.xml/log4j:log4j/open

avirtuos:
We do not make us of the vulnerable TcpSocketServer and UdpSocketServer appenders that extend the vulnerable AbstractSocketServer. I've resolved the alert on the repo itself but will keep this issue open to track removing the dependency entirely.  Here is the associated issue in the open source project; https://issues.apache.org/jira/browse/LOG4J2-1863

avirtuos:
looks like 2.9.0 is the version we need to switch to @atennak1

atennak1:
Merged https://github.com/awslabs/aws-athena-query-federation/pull/91

PR #91, opened 2020-01-08
Upgrading to Log4j 2 to address Log4j vulnerability

atennak1:
*Issue #, if available:* https://github.com/awslabs/aws-athena-query-federation/issues/89

*Description of changes:* Upgrading to Log4j 2.13.0 to address vulnerability.

*Testing*: mvn dependency:tree no longer shows Log4j 1.x.  Deployed and tested DDB connector and verified no regression.


avirtuos:
Did we test this by running tests for all connectors, the ask, Athena, and @seth's stuff? I vaguely recall log4j2 caused issues for something. Why not just bump to the newer log4j 1?  

Alternatively we could change our pom so that only connectors have this dependency and not the ask itself.

atennak1:
It's my understanding that this vulnerability is not addressed in log4j v1.

I was also concerned that transitive dependencies may rely on log4j v1 APIs but I took a look at the existing dependency tree before modifications and luckily for us all of them used JCL or Slf4j.

I suppose it's possible that future connectors may have the log4j v1 dependency, in which case I believe they may be able to bridge to Slf4j or log4j v2.

In any case I do think pushing down the log4j v2 dependencies down to the connectors and out of the SDK is the right move for cleanliness anyway so I will make that change.

avirtuos:
approved, feel free to merge and test when you are ready.

atennak1:
Tested all connectors by deploying them and running validator script and actual queries.

PR #99, opened 2020-01-30
Setting startFromHead flag for correct pagination in CloudwatchRecord…

atennak1:
…Handler

*Issue #, if available:*

*Description of changes:* A customer reported that their query was returning incomplete data.  After investigation, I discovered that we were not using the GetLogEvents API correctly.  The documentation states that in order to use nextToken, startFromHead must be set to true.  I asked them to test this fix and all data was being returned.

https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_GetLogEvents.html#API_GetLogEvents_RequestParameters


avirtuos:
Nice find!  Though, I'm VERY interested in why "Using this token works only when you specify true for startFromHead." is the case for that API. It seems so error prone...as evidenced here.

PR #119, opened 2020-02-27
Adding datetime support from glue table

soojinj:
*Description of changes:*
Currently we do not support date and timestamp types from schema registered in glue. The changes include 
1. supporting those types coming from glue schema
2. converting string coming from dynamodb with the glue schema's date and timestamp to local date time
3. default timezone and customized time string format can be added to glue table metadata


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/SchemaBuilder.java
```diff
@@ -279,6 +280,14 @@ public SchemaBuilder addMetadata(String key, String value)
         return this;
     }
 
+    public SchemaBuilder addMetadata(String key, String value, Set<String> excludeKey)
```
avirtuos: as a rule, every public method needs java doc and most private methods that are non-trivial do too. 

What is excludeKey's meaning?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
```diff
@@ -330,12 +333,18 @@ protected GetTableResponse doGetTable(BlockAllocator blockAllocator, GetTableReq
...
-            table.getParameters().entrySet().forEach(next -> schemaBuilder.addMetadata(next.getKey(), next.getValue()));
+            table.getParameters()
+                    .entrySet()
+                    .forEach(next -> schemaBuilder.addMetadata(next.getKey(), next.getValue(), excludeKey));
         }
 
         //A column name mapping can be provided to get around restrictive Glue naming rules
         Map<String, String> columnNameMapping = getColumnNameMapping(table);
+        Map<String, String> dateTimeFormatMapping = getDateTimeFormatMapping(table);
+        Map<String, String> datetimeFormatMappingWithColumnName = new HashMap<>();
```
avirtuos: instead of multiple maps, maybe we can create a column object that holds the info we care about?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
```diff
@@ -349,7 +358,11 @@ protected GetTableResponse doGetTable(BlockAllocator blockAllocator, GetTableReq
             if (next.getComment() != null) {
                 schemaBuilder.addMetadata(mappedColumnName, next.getComment());
             }
+            if (dateTimeFormatMapping.containsKey(next.getName())) {
+                datetimeFormatMappingWithColumnName.put(mappedColumnName, dateTimeFormatMapping.get(next.getName()));
+            }
         }
+        populateDatetimeFormatMappingIfAvailable(schemaBuilder, datetimeFormatMappingWithColumnName);
```
avirtuos: can't we do this inside the above loop instead of looping over the columns again?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
```diff
@@ -449,4 +462,24 @@ protected static String getSourceTableName(Schema schema)
         }
         return ImmutableMap.of();
     }
+
+    protected static Map<String, String> getDateTimeFormatMapping(Table table)
```
avirtuos: why protected? isnt this info available on the Schema object we return from getTable?

avirtuos: also, if we do keep this, it needs javadoc

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/examples/ExampleRecordHandlerTest.java
```diff
@@ -83,7 +83,7 @@
 
 public class ExampleRecordHandlerTest
 {
-    private static final Logger logger = LoggerFactory.getLogger(ExampleMetadataHandlerTest.class);
+    private static final Logger logger = LoggerFactory.getLogger(ExampleRecordHandlerTest.class);
```
avirtuos: nice catch!

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandlerTest.java
```diff
@@ -69,7 +70,7 @@
 
 public class CompositeHandlerTest
 {
-    private static final Logger logger = LoggerFactory.getLogger(ExampleMetadataHandlerTest.class);
+    private static final Logger logger = LoggerFactory.getLogger(CompositeHandlerTest.class);
```
avirtuos: nice catch!

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandlerTest.java
```diff
@@ -78,10 +79,15 @@
...
     private Schema schemaForRead;
 
+    @Rule
+    public TestName testName = new TestName();
+
     @Before
     public void setUp()
             throws Exception
     {
+        logger.info("{}: enter", testName.getMethodName());
```
avirtuos: ohhh, i like this. can we do this! is this why you were able to remove the other log statements way above?

soojinj: Yep! @Before and @After runs before each test cases, so thats why i removed it in the other tests above, but i did add the logging lines to setup and teardown methods.

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBRecordHandler.java
```diff
@@ -198,6 +217,17 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
         logger.info("readWithConstraint: numRows[{}] numResultRows[{}]", numRows, numResultRows.get());
     }
 
+    private Map<String, String> getDateTimeFormatMapping(Schema schema)
```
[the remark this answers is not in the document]
soojinj: added in the next revision

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
soojinj: Right now, the customer provides a mapping of column names (ex col1=Col1 where col1 is glue name and Col1 is ddb name). I thought it was better for the customer to do the same for date format mapping (i.e. col1=someformat instead of Col1=someformat) since col1 is glue name and the mapping is on glue table metadata.

this *could be* available on the schema object as is, but Ive added excludeKey in addMetadata is to prevent the un-normalized mapping to be added to metadata when we go through them. If i keep the raw map in the metadata, it would be on the schema object but it will have glue column name, not the normalized ddb column name.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
soojinj: I am "aggregating" key-value in the map as a single string and adding to schema metadata in populateDatetimeFormatMappingIfAvailable. the Normalizing of the column name happens within the loop

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
soojinj: Im not sure what you mean by creating a column object. dateTimeFormatMapping holds the raw map from the table, and datetimeFormatMappingWithColumnName is something GlueMetadataHandler builds based on dateTimeFormatMapping with "normalized" column name

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/SchemaBuilder.java
soojinj: This is created to prevent the map of raw colum name to date format from being added to the schema metadata. instead we normalize the map using column name mapping and add that to the schema metadata.

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/constants/DynamoDBConstants.java
```diff
@@ -37,4 +37,7 @@ private DynamoDBConstants() {}
     public static final String NON_KEY_FILTER_METADATA = "nonKeyFilter";
     public static final String EXPRESSION_NAMES_METADATA = "expressionAttributeNames";
     public static final String EXPRESSION_VALUES_METADATA = "expressionAttributeValues";
+
+    public static final String DATETIME_FORMAT_MAPPING_PROPERTY = "datetimeFormatMapping";
+    public static final String DEFAULT_TIME_ZONE = "defaultTimeZone";
```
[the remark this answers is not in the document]
soojinj: done

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBRecordHandler.java
```diff
@@ -162,6 +173,14 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
                         value = DDBTypeUtils.coerceDecimalToExpectedType((BigDecimal) value, fieldType);
                     }
+                    if ((fieldType.equals(Types.MinorType.DATEMILLI) || fieldType.equals(Types.MinorType.DATEDAY))
+                            && (value instanceof String || value instanceof BigDecimal)) {
+                        ZoneId defaultTimeZone = ZoneId.of(recordsRequest.getSchema().getCustomMetadata().getOrDefault(
+                                DEFAULT_TIME_ZONE, UTC));
+                        String customerConfiguredFormat = dateTimeFormatMapping.getOrDefault(nextField.getName(), null);
+                        value = DDBTypeUtils.coerceDateTimeToExpectedType(value, fieldType,
+                                customerConfiguredFormat, defaultTimeZone);
+                    }
```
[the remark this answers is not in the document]
soojinj: Ive addressed this

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTypeUtils.java
```diff
@@ -206,4 +208,36 @@ public static Object coerceDecimalToExpectedType(BigDecimal value, Types.MinorTy
...
+                    case DATEDAY:
+                        return DateTimeFormatterUtil.bigDecimalToLocalDate((BigDecimal) value, defaultTimeZone);
+                    default:
+                        return value;
+                }
+            }
+            return value;
+        }
+        catch (IllegalArgumentException ex) {
+            logger.error(ex.getMessage());
```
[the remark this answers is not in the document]
soojinj: we can throw illegalArgumentexception then and there. it would only throw if any of the date/datetime values fails to parse with any of the custom or supported format. As it is malformed data, we could have the query fail? I continue passing it as string and set vector as string but it would fail on any filtering and such anyways

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/DateTimeFormatterUtil.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public static LocalDateTime bigDecimalToLocalDateTime(BigDecimal value, ZoneId defaultTimeZone)
+    {
+        Date dateTransformedValue = bigDecimalToDate(value);
+        return dateTransformedValue.toInstant()
+                .atZone(defaultTimeZone)
+                .toLocalDateTime();
+    }
+
+    private static Date stringToDate(String value, String customerConfiguredFormat)
```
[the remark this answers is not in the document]
soojinj: do you think its better to force the customers to provide a custom date/datetime format mapping if any of the columns are of type date/datetime?

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBRecordMetadata.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.arrow.vector.types.pojo.Schema;
+
+import java.time.ZoneId;
+import java.util.Map;
+
+import static com.amazonaws.athena.connectors.dynamodb.constants.DynamoDBConstants.DATETIME_FORMAT_MAPPING_PROPERTY;
+import static com.amazonaws.athena.connectors.dynamodb.constants.DynamoDBConstants.DEFAULT_TIME_ZONE;
+import static com.amazonaws.athena.connectors.dynamodb.constants.DynamoDBConstants.UTC;
+
+public class DDBRecordMetadata
```
avirtuos: java doc for the class

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBRecordMetadata.java
```diff
@@ -0,0 +1,75 @@
...
+{
+    private static final Splitter.MapSplitter MAP_SPLITTER = Splitter.on(",").trimResults().withKeyValueSeparator("=");
+
+    private Map<String, String> dateTimeFormatMapping;
+    private ZoneId defaultTimeZone;
+
+    public DDBRecordMetadata(Schema schema)
```
avirtuos: java doc for all public methods

athena-federation-sdk/pom.xml
```diff
@@ -142,6 +142,11 @@
             <version>1.10.19</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.9</version>
+        </dependency>
```
avirtuos: What are we using from commons-lang3? shoot me a chime message about this since we will need to ensure this plays nice with anything that uses the SDK.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/DateTimeFormatterUtil.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.commons.lang3.time.DateUtils;
+
+import java.math.BigDecimal;
+import java.text.ParseException;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.Date;
+
+public class DateTimeFormatterUtil
```
avirtuos: java doc for all public classes / methods

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/DateTimeFormatterUtil.java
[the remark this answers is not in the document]
avirtuos: i think we can figure it out ourselves similar to schema inference. For example, after you read the 1st row you know what the format is for the rest of the rows. We don't need to rely on exception catching to re-detect the format of each row. Let me know if a quick chat would be easier to clarify this.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/SchemaBuilder.java
```diff
@@ -279,6 +280,22 @@ public SchemaBuilder addMetadata(String key, String value)
...
 
+    /**
+     * Adds the provided metadata to the Schema, unless the provided key is one of the values in excludedKeys
+     *
+     * @param key The key of the metadata to add.
+     * @param value The value of the metadata to add.
+     * @param excludeKeys set of strings that should be excluded from being added to the metadata map
+     * @return This SchemaBuilder itself.
+     */
+    public SchemaBuilder addMetadata(String key, String value, Set<String> excludeKeys)
```
avirtuos: This method strikes me as fairly specific / odd. What would be an example of an excludedKey? and why would we be attempting to addMetadata for the same key twice?

avirtuos: Ok, i see where it is used further down. My advice is to prevent SchemaBuilder from allowing the same key to be set multiple times. You may need to add a set or something else to track keys since the Arrow object might not be helpful for this. Then, in the DDB connector or Glue handler, set the keys you want to treat as reserved first. Then have the addMetadata function throw 'DuplicateMetadtaException' if a key gets set more than once.

athena-federation-sdk/pom.xml
soojinj: Im using this to parse date/datetime

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/DateTimeFormatterUtil.java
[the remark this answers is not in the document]
soojinj: Ive changed to cache the first matching format and use that going forward for all the subsequent records. I am returning null for any string that is not of the cached format with logging indicating so.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/SchemaBuilder.java
soojinj: its not that I want to add the same key twice (which is already not possible as metadata is ImmutableMap). When we read in table parameters, we currently copy everything field to the schema metadata. I want to change this so that we exclude some keys from being copied over. I then later add a "normalized value" for the skipped key to metadata.

avirtuos: I think my view is still the same, I think this method makes using SchemaBuilder more confusing and would recommend we handle this outside the builder. Would that be possible or painful? I'm not sure I've seen this usecase in any other connector yet which is also why I'm apprehensive to change this in the SDK.

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTypeUtils.java
```diff
@@ -199,10 +200,17 @@ public static Object coerceValueToExpectedType(Object value, Field field,
...
-            String customerConfiguredFormat = recordMetadata.getDateTimeFormat(field.getName());
+            String dateTimeFormat = recordMetadata.getDateTimeFormat(field.getName());
+            if (value instanceof String && StringUtils.isEmpty(dateTimeFormat)) {
+                logger.info("Date format not in cache for column {}. Trying to infer format...", field.getName());
+                dateTimeFormat = DateTimeFormatterUtil.inferDateTimeFormat((String) value);
+                if (StringUtils.isNotEmpty(dateTimeFormat)) {
+                    logger.info("Adding datetime format {} for column {} to cache", dateTimeFormat, field.getName());
+                    recordMetadata.setDateTimeFormat(field.getName(), dateTimeFormat);
+                }
+            }
```
avirtuos: Thanks, keep in mind that this is likely to get refactored a bit when we move this connector to the extractor framework.

avirtuos:
looks good to me, please update your branch and also update the readme to include this new feature then I'll give a ship-it on the PR.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/DateTimeFormatterUtil.java
```diff
@@ -0,0 +1,190 @@
...
+            try {
+                DateUtils.parseDate(value, datetimeFormat);
+                logger.info("Inferred format {} for value {}", datetimeFormat, value);
+                return datetimeFormat;
+            }
+            catch (ParseException ex) {
+                continue;
+            }
+        }
+        logger.error("Failed to infer format for {}", value);
```
atennak1: warn since we fallback?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/DateTimeFormatterUtil.java
```diff
@@ -0,0 +1,190 @@
...
+     * @param dateFormat date format to be used to parse
+     * @return date object that is parsed from raw value
+     */
+    private static Date stringToDate(String value, String dateFormat)
+    {
+        try {
+            return DateUtils.parseDate(value, dateFormat);
+        }
+        catch (ParseException | IllegalArgumentException ex) {
+            logger.error("Encountered value {} that is not consistent with inferred date/datetime format {}" +
```
atennak1: warn since we fallback?

and is there any value in logging the exception message or stacktrace?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/DateTimeFormatterUtil.java
```diff
@@ -0,0 +1,190 @@
...
+                            " - will skip in result",
+                    value, dateFormat);
+        }
+        return null;
+    }
+
+    /**
+     * trnasforms big decimal value (epoch milli) to date
```
atennak1: typo

athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseMetadataHandler.java
```diff
@@ -93,7 +93,7 @@
     private static final String DEFAULT_HBASE = "default_hbase";
     //The Glue table property that indicates that a table matching the name of an HBase table
     //is indeed enabled for use by this connector.
-    private static final String HBASE_METADATA_FLAG = "hbase-metadata-flag";
+    public static final String HBASE_METADATA_FLAG = "hbase-metadata-flag";
```
atennak1: minor: package-private

athena-hbase/src/test/java/com/amazonaws/athena/connectors/hbase/HbaseMetadataHandlerTest.java
```diff
@@ -223,19 +224,54 @@ public void doGetTable()
...
+        assertEquals(4, getTableResponse.getSchema().getFields().size());
+
+        Field col1 = getTableResponse.getSchema().findField("Col1");
+        assertEquals(Types.MinorType.INT, Types.getMinorTypeForArrowType(col1.getType()));
+
+        Field col2 = getTableResponse.getSchema().findField("Col2");
+        assertEquals(Types.MinorType.DATEDAY, Types.getMinorTypeForArrowType(col2.getType()));
+
+        Field col3 = getTableResponse.getSchema().findField("Col3");
+        assertEquals(Types.MinorType.DATEMILLI, Types.getMinorTypeForArrowType(col3.getType()));
```
avirtuos: I'm confused how we are supporting this from HBase, how do we go from HBase data format to dateday or datemilli? Specifically HbaseSchemaUtils.coerceType(...) has no code for handling Dateday and datemilli so its not possible to query these types. My suggestion is to rollback changes to any connector other than DDB unless you have seen these other connectors successfully query these types from their real source (e.g. not a unit test). Or is this just a new unit test which happened to have these types?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/DateTimeFormatterUtil.java
soojinj: I dont think so, i think the message shows exactly what the input was and what format it was trying to use.

athena-hbase/src/test/java/com/amazonaws/athena/connectors/hbase/HbaseMetadataHandlerTest.java
soojinj: removed from metadata Handler. we are not supporting them in record handler for hbase and redis

Issue #21, opened 2019-11-16
[FEATURE] Add support for Timestampw/TZ


avirtuos:
Add support for Timestampw/TZ to the SDK, Athena, and relevant connectors (e.g. jdbc)

PR #130, opened 2020-03-24
[ISSUE 21] Adding timezone support

soojinj:
*Issue #, if available:*
https://github.com/awslabs/aws-athena-query-federation/issues/21

*Description of changes:*
SDK side of changes to support timezone


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
```diff
@@ -42,6 +42,7 @@
     BIT("boolean", Types.MinorType.BIT.getType()),
     VARBINARY("binary", Types.MinorType.VARBINARY.getType()),
     TIMESTAMP("timestamp", Types.MinorType.DATEMILLI.getType()),
+    TIMESTAMPTZ("timestamptz", Types.MinorType.TIMESTAMPMILLI.getType()),
```
atennak1: may not be possible but just a (possible dumb) thought: what if we treat timestamp as timestamptz? In other words, replace DATEMILLI with TIMESTAMPMILLI everywhere. What are the implications? If possible, does it simplify or complicate?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/BlockUtils.java
```diff
@@ -455,7 +481,8 @@ public static int copyRows(Block srcBlock, Block dstBlock, int firstRow, int las
             for (int i = firstRow; i <= lastRow; i++) {
                 FieldVector dst = dstBlock.getFieldVector(src.getField().getName());
                 src.setPosition(i);
-                setValue(dst, dstOffset++, src.readObject());
+                Object normalizeObject = normalizeObject(src.readObject());
```
avirtuos: this adds an extra comparison on every every cell, can we do this in a way that we don't need to check every cell's type?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/BlockUtils.java
```diff
@@ -464,6 +491,24 @@ public static int copyRows(Block srcBlock, Block dstBlock, int firstRow, int las
         return rowsCopied;
     }
 
+    /**
+     * We will convert any types that are not supported by setValue to types that are supported
+     * ex) (not supported) org.joda.time.LocalDateTime which is returned on read from vectors
+     * will be converted to (supported) java.time.ZonedDateTime
+     * @param o1 object with unsupported type
+     * @return another object with supported/converted type
+     */
```
avirtuos: why can't we just make setvalue handle this so we avoid the extra type checking and branching?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/SupportedTypes.java
```diff
@@ -36,6 +36,7 @@
     BIT(Types.MinorType.BIT),
     DATEMILLI(Types.MinorType.DATEMILLI),
     DATEDAY(Types.MinorType.DATEDAY),
+    TIMESTAMPMILLI(Types.MinorType.TIMESTAMPMILLI),
```
avirtuos: i thought the task was to add support for TIMESTAMPMILLITZ, wont it be weird if we add TIMESTAMPMILLI (no TZ) and tell people it has TZ awareness? What am i missing

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/TimeZoneKey.java
```diff
@@ -0,0 +1,324 @@
...
+
+import static java.lang.Character.isDigit;
+import static java.lang.Math.abs;
+import static java.lang.Math.max;
+import static java.util.Locale.ENGLISH;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Class for representing timezone to be used in encoding datetime value and timezone value to a single long
+ * a copy from Presto TimeZoneKey
```
avirtuos: if this is an exact copy you might need to copy attribution too. Do we really need it to be an exact copy?

Also, we need to be careful about what apache arrow type we are using for this. If a customer actually wants to use TIMESTAMPMILLI that seems to no longer be possible because we are now defining timestampmilli as something else.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
```diff
@@ -114,6 +114,8 @@
     public static final String DATETIME_FORMAT_MAPPING_PROPERTY = "datetimeFormatMapping";
     // Table property (optional) that we will create from DATETIME_FORMAT_MAPPING_PROPERTY with normalized column names
     public static final String DATETIME_FORMAT_MAPPING_PROPERTY_NORMALIZED = "datetimeFormatMappingNormalized";
+    // Table property (optional) that we expect to contain any types to be overridden for a column, ex) timestamp with tz
+    public static final String TYPE_OVERRIDE_MAPPING_PROPERTY = "typeOverrideMapping";
```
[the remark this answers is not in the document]
avirtuos: +1 we need clear documenation on this. I'm also unclear if we really need this since Glue doesn't validate what you put into the type field. its just a string i believe.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
avirtuos: This mapping seems wrong to me, we are redefining what TIMESTAMPMILLI means, why not map it to TIMESTAMPMILLI With TZ?

athena-federation-sdk/src/main/resources/com/amazonaws/athena/connector/lambda/data/zone-index.properties
```diff
@@ -0,0 +1,2237 @@
+#
+# DO NOT REMOVE OR MODIFY EXISTING ENTRIES
+#
+# This file contain the fixed numeric id of every supported time zone id.
+# Every zone id in this file must be supported by java.util.TimeZone and the
+# Joda time library.  This is because Presto uses both java.util.TimeZone and
+# the Joda time## for during execution.
+#
+# suppress inspection "UnusedProperty" for whole file
```
avirtuos: This file makes me super nervous. Can't we get this from Java runtime somehow? what happens if we country changes its offset?

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/data/TimeZoneKeyTest.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.SortedSet;
+
+import static com.amazonaws.athena.connector.lambda.data.TimeZoneKey.MAX_TIME_ZONE_KEY;
+import static com.amazonaws.athena.connector.lambda.data.TimeZoneKey.UTC_KEY;
+import static java.util.Locale.ENGLISH;
+import static org.junit.Assert.*;
+
+/**
+ * Class for testing timezone to be used in encoding datetime value and timezone value to a single long
+ * a copy from Presto TimeZoneKeyTest
```
avirtuos: please ensure we are putting proper attribution to presto for any code that we reuse.

Issue #137, opened 2020-04-13
[BUG] type inferrence error with connector DocDB


soojinj:
**Describe the bug**
when there is a list within a struct there is an error with conversion (tbd on detail)

**To Reproduce**
1. create a document with 
```
        List<Document> documents = new ArrayList<>();
        Document result = new Document();
        documents.add(result);

        Document listStruct = new Document();
        listStruct.put("substruct", "someSubStruct");
        List<Document> list = new ArrayList<>();
        list.add(listStruct);
        Document structWithList = new Document();
        structWithList.put("Lists", list);
        result.put("ComplexStruct", structWithList);
```

2. call getTableRequest
3. call readRecordRequest with inferred schema from #2

**Screenshots / Exceptions / Errors**
```
java.lang.RuntimeException: Error while processing field ComplexStruct

	at com.amazonaws.athena.connectors.docdb.DocDBRecordHandler.lambda$readWithConstraint$0(DocDBRecordHandler.java:158)
	at com.amazonaws.athena.connector.lambda.data.S3BlockSpiller.writeRows(S3BlockSpiller.java:174)
	at com.amazonaws.athena.connectors.docdb.DocDBRecordHandler.readWithConstraint(DocDBRecordHandler.java:136)
	at com.amazonaws.athena.connector.lambda.handlers.RecordHandler.doReadRecords(RecordHandler.java:191)
	at com.amazonaws.athena.connectors.docdb.DocDBRecordHandlerTest.simpleTest(DocDBRecordHandlerTest.java:279)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
[...]

PR #138, opened 2020-04-14
Fixing List type in a Document for docdb connector

soojinj:
*Issue #, if available:*
https://github.com/awslabs/aws-athena-query-federation/issues/137

*Testing:*
Unit test


avirtuos:
Nice catch Soojin! Thanks for taking this up so quickly.

Issue #140, opened 2020-04-15
[BUG] describe table query will sometimes generate a disjointed schema view.


shurvitz:
**Describe the bug**
the SchemaBuilder underlying container for the fields is declared a HashMap, but should be a LinkedHashMap instead. The difference is that a LinkedHashMap maintains the order of the elements in the map.

To illustrate the problem, the following schema list was generated by running a 'describe table' query on the elasticsearch connector (currently in development). Even though, all fields were added to the SchemaBuilder object in order, the resulting list displays the field '**address.street**' out of position (should come after '**address.town**'):

**address.zip**	varchar
**address.town**	varchar
phone.mobile	varchar
phone.landline	varchar
members	array<varchar>
**address.street** varchar
account	varchar
email	array<varchar> 

**To Reproduce**
1. Go to the Athena console.
2. Create a new query: `describe 'lambda:elasticsearch'.movies.customer;`
3. Press Run query.
4. See Results

**Expected behavior**
Expected schema:

**address.zip**	varchar
**address.town**	varchar
**address.street**	varchar
phone.mobile	varchar
phone.landline	varchar
members	array<varchar>
account	varchar
email	array<varchar>

avirtuos:
Thanks for filing this @shurvitz . Per our chat, let me know when you raise a PR for the fix.

avirtuos:
Thanks @shurvitz , fix has been merged and will go out in the next release.

PR #145, opened 2020-04-16
Fix BUG (Issue #140) with SchemaBuilder

shurvitz:
*Issue #, if available:* #140 

*Description of changes:*
Changed fields from HashMap to LinkedHashMap. The latter will maintain the order of the elements inserted and will present a consistent schema view.
Consequently, had to change two tests in the jdbc connector to conform to new SchemaBuider.


avirtuos:
Thanks @shurvitz. I've merged your changes into master.

PR #146, opened 2020-04-17
passing in required columns to dynamodb query/scan

soojinj:
*Description of changes:*
passing in required columns to dynamodb query/scan

### DIFF
--- a/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBRecordHandler.java
+++ b/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBRecordHandler.java
@@ -50,6 +50,7 @@ import com.google.common.cache.LoadingCache;
 import org.apache.arrow.util.VisibleForTesting;
 import org.apache.arrow.vector.types.Types;
 import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.arrow.vector.types.pojo.Schema;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBRecordHandler.java
```diff
@@ -219,6 +221,12 @@ private AmazonWebServiceRequest buildReadRequest(Split split, String tableName)
             }
         }
 
+        // Only read columns that are needed in the query
+        String projectionExpression = schema.getFields()
+                .stream()
+                .map(field -> field.getName())
```
atennak1, line 227: do we need to alias columns in case they are in the filter expression?

soojinj: schema already has the dynamodb column names (after columnNameMapping is applied) so no need to do more

atennak1: I meant the '#' prepending for filter expressions

soojinj: that is taken care of in the filterExpression and already contains the right format

atennak1: I vaguely remember having to do it for the projection expression as well but if you tested this with `select a from table where <constraint on a>` then I'm good.

soojinj: yep one of the tests ive done is 

select i.item_id, i.cost, o.ship_address.city from itemtable i, ordertable o where i.item_id = o.item_id and o.returned = false order by i.item_name;

atennak1: can you try `select o.returned, i.item_id, i.cost, o.ship_address.city from itemtable i, ordertable o where i.item_id = o.item_id and o.returned = false order by i.item_name;`?

the item_id constraint isn't pushed down

atennak1: actually 'returned' is in both the projection and filter so we should be good.

PR #164, opened 2020-05-11
Spillbucketauthz

luc-liu:
Added spill bucket authorization check (prackage-private for ease of unit test) in MetadataHandler, added unit test for it in CompositeHandlerTest

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
@@ -48,7 +48,7 @@ public class SpillLocationVerifier
     public SpillLocationVerifier(AmazonS3 amazons3)
     {
         this.amazons3 = amazons3;
-        this.bucket = "";
+        this.bucket = null;
         this.state = BucketState.UNCHECKED;
     }
 
@@ -63,7 +63,7 @@ public class SpillLocationVerifier
             return;
         }
 
-        if (!bucket.equals(spillBucket)) {
+        if (bucket == null || !bucket.equals(spillBucket)) {
             logger.info("Spill bucket has been changed from {} to {}", bucket, spillBucket);
             bucket = spillBucket;
             state = BucketState.UNCHECKED;
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
@@ -21,7 +21,6 @@ package com.amazonaws.athena.connector.lambda.domain.spill;
  */
 
 import com.amazonaws.services.s3.AmazonS3;
-import com.amazonaws.services.s3.AmazonS3ClientBuilder;
 import com.google.common.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -39,14 +38,16 @@ public class SpillLocationVerifier
     private enum BucketState
     {UNCHECKED, VALID, INVALID}
 
+    private final AmazonS3 s3;
     private String bucket;
     private BucketState state;
 
     /**
      * @param bucket The name of the spill bucket.
      */
-    public SpillLocationVerifier(String bucket)
+    public SpillLocationVerifier(AmazonS3 s3, String bucket)
     {
+        this.s3 = s3;
         this.bucket = bucket;
         this.state = BucketState.UNCHECKED;
     }
@@ -58,8 +59,12 @@ public class SpillLocationVerifier
      */
     public void checkBucketAuthZ(String spillBucket)
     {
+        if (spillBucket == null) {
+            throw new RuntimeException("Spill bucket cannot be null!");
+        }
+
         if (!bucket.equals(spillBucket)) {
-            logger.debug("Spill bucket has been changed from " + bucket + " to " + spillBucket);
+            logger.info("Spill bucket has been changed from {} to {}", bucket, spillBucket);
             bucket = spillBucket;
             state = BucketState.UNCHECKED;
         }
@@ -78,7 +83,6 @@ public class SpillLocationVerifier
     @VisibleForTesting
     void updateBucketState()
     {
-        AmazonS3 s3 = getS3();
         Set<String> buckets = s3.listBuckets().stream().map(b -> b.getName()).collect(Collectors.toSet());
 
         if (!buckets.contains(bucket)) {
@@ -88,7 +92,7 @@ public class SpillLocationVerifier
             state = BucketState.VALID;
         }
 
-        logger.debug("The state of bucket " + bucket + " has been updated to " + state + " from " + BucketState.UNCHECKED);
+        logger.info("The state of bucket {} has been updated to {} from {}", bucket, state, BucketState.UNCHECKED);
     }
 
     /**
@@ -107,14 +111,5 @@ public class SpillLocationVerifier
         else {
             // no op
         }
-    }
-
-    /**
-     * Helper function to get the AmazonS3 instance of the lambda
-     */
-    @VisibleForTesting
-    AmazonS3 getS3()
-    {
-        return AmazonS3ClientBuilder.standard().build();
     }
 }
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
@@ -38,17 +38,17 @@ public class SpillLocationVerifier
     private enum BucketState
     {UNCHECKED, VALID, INVALID}
 
-    private final AmazonS3 s3;
+    private final AmazonS3 amazons3;
     private String bucket;
     private BucketState state;
 
     /**
-     * @param bucket The name of the spill bucket.
+     * @param amazons3 The S3 object for the account.
      */
-    public SpillLocationVerifier(AmazonS3 s3, String bucket)
+    public SpillLocationVerifier(AmazonS3 amazons3)
     {
-        this.s3 = s3;
-        this.bucket = bucket;
+        this.amazons3 = amazons3;
+        this.bucket = "";
         this.state = BucketState.UNCHECKED;
     }
 
@@ -59,8 +59,8 @@ public class SpillLocationVerifier
      */
     public void checkBucketAuthZ(String spillBucket)
     {
-        if (spillBucket == null) {
-            throw new RuntimeException("Spill bucket cannot be null!");
+        if (spillBucket == null || spillBucket.equals("")) {
+            return;
         }
 
         if (!bucket.equals(spillBucket)) {
@@ -83,16 +83,21 @@ public class SpillLocationVerifier
     @VisibleForTesting
     void updateBucketState()
     {
-        Set<String> buckets = s3.listBuckets().stream().map(b -> b.getName()).collect(Collectors.toSet());
+        try {
+            Set<String> buckets = amazons3.listBuckets().stream().map(b -> b.getName()).collect(Collectors.toSet());
 
-        if (!buckets.contains(bucket)) {
-            state = BucketState.INVALID;
-        }
-        else {
-            state = BucketState.VALID;
-        }
+            if (!buckets.contains(bucket)) {
+                state = BucketState.INVALID;
+            }
+            else {
+                state = BucketState.VALID;
+            }
 
-        logger.info("The state of bucket {} has been updated to {} from {}", bucket, state, BucketState.UNCHECKED);
+            logger.info("The state of bucket {} has been updated to {} from {}", bucket, state, BucketState.UNCHECKED);
+        }
+        catch (Exception ex) {
+            throw new RuntimeException("Error while checking bucket ownership for " + bucket, ex);
+        }
     }
 
     /**
@@ -102,14 +107,13 @@ public class SpillLocationVerifier
     @VisibleForTesting
     void passOrFail()
     {
-        if (state == BucketState.UNCHECKED) {
-            throw new RuntimeException("Bucket state should have been checked already.");
-        }
-        else if (state == BucketState.INVALID) {
-            throw new RuntimeException("You do NOT own the spill bucket with the name: " + bucket);
-        }
-        else {
-            // no op
+        switch (state) {
+            case UNCHECKED:
+                throw new RuntimeException("Bucket state should have been checked already.");
+            case INVALID:
+                throw new RuntimeException("You do NOT own the spill bucket with the name: " + bucket);
+            default:
+                return;
         }
     }
 }
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
@@ -1,0 +1,120 @@
+package com.amazonaws.athena.connector.lambda.domain.spill;
+
+/*-
+ * #%L
+ * Amazon Athena Query Federation SDK
+ * %%
+ * Copyright (C) 2019 - 2020 Amazon Web Services
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
+import com.google.common.annotations.VisibleForTesting;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * This class is used to track the bucket and its state, and check its validity
+ */
+public class SpillLocationVerifier
+{
+    private static final Logger logger = LoggerFactory.getLogger(SpillLocationVerifier.class);
+
+    private enum BucketState
+    {UNCHECKED, VALID, INVALID}
+
+    private String bucket;
+    private BucketState state;
+
+    /**
+     * @param bucket The name of the spill bucket.
+     */
+    public SpillLocationVerifier(String bucket)
+    {
+        this.bucket = bucket;
+        this.state = BucketState.UNCHECKED;
+    }
+
+    /**
+     * Public function used to check if the account that calls the lambda function owns the spill bucket
+     *
+     * @param spillBucket The name of the spill bucket.
+     */
+    public void checkBucketAuthZ(String spillBucket)
+    {
+        if (!bucket.equals(spillBucket)) {
+            logger.debug("Spill bucket has been changed from " + bucket + " to " + spillBucket);
+            bucket = spillBucket;
+            state = BucketState.UNCHECKED;
+        }
+
+        if (state == BucketState.UNCHECKED) {
+            updateBucketState();
+        }
+
+        passOrFail();
+    }
+
+    /**
+     * Helper function to check bucket ownership if it hasn't been checked before
+     *
+     */
+    @VisibleForTesting
+    void updateBucketState()
+    {
+        AmazonS3 s3 = getS3();
+        Set<String> buckets = s3.listBuckets().stream().map(b -> b.getName()).collect(Collectors.toSet());
+
+        if (!buckets.contains(bucket)) {
+            state = BucketState.INVALID;
+        }
+        else {
+            state = BucketState.VALID;
+        }
+
+        logger.debug("The state of bucket " + bucket + " has been updated to " + state + " from " + BucketState.UNCHECKED);
+    }
+
+    /**
+     * Helper function to throw exception if lambda function doesn't own the spill bucket
+     *
+     */
+    @VisibleForTesting
+    void passOrFail()
+    {
+        if (state == BucketState.UNCHECKED) {
+            throw new RuntimeException("Bucket state should have been checked already.");
+        }
+        else if (state == BucketState.INVALID) {
+            throw new RuntimeException("You do NOT own the spill bucket with the name: " + bucket);
+        }
+        else {
+            // no op
+        }
+    }
+
+    /**
+     * Helper function to get the AmazonS3 instance of the lambda
+     */
+    @VisibleForTesting
+    AmazonS3 getS3()
+    {
+        return AmazonS3ClientBuilder.standard().build();
+    }
+}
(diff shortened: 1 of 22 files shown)


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -426,6 +431,7 @@ public abstract GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplit
      */
     public PingResponse doPing(PingRequest request)
     {
+        checkBucketAuthZ(AmazonS3ClientBuilder.standard().build(), spillBucket);
```
avirtuos, line 434: Ping is meant to be light-weight so that it aids in both discovery and cold-start. I don't think we should apply this from PING. What are your thoughts on doing this from getSplits which is the 1st place that the spill bucket is potentially used?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -446,6 +452,21 @@ public void onPing(PingRequest request)
...
+    /**
+     * Helper function used to check if the account that calls the lambda function owns the spill bucket
+     *
+     * @param s3 The S3 object for the account.
+     * @param bucket The name of the spill bucket.
+     */
+    @VisibleForTesting
+    void checkBucketAuthZ(final AmazonS3 s3, final String bucket)
+    {
+        Set<String> buckets = s3.listBuckets().stream().map(b -> b.getName()).collect(Collectors.toSet());
```
avirtuos, line 464: IsI dont think we need to do this more than once so we should cache the check and only re-check if the bucket name changes.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
luc-liu: There's a tradeoff. If a customer's request will result in spilling, then checking as early as possible avoids wasting resources before spilling fails. However, an unauthorized bucket will prevent a customer from getting the response even when the response is in-memory

luc-liu: I've done some experiments. Checking bucket ownership takes around 600 ms for an account that owns 25 buckets. This brings the total doPing time up to 620 ms. In terms of latency, checking bucket ownership is not "light-weight".

If this is what you are concerned about, I can move it to getSplits. What are your thoughts on potentially wasting some resources before checking bucket ownership?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
luc-liu: This is easily fixable. I'll make the changes once we settle on where the checking should be.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
avirtuos: getSplits happens pretty early in the query lifecycle so I'm ok with it. We also don't expect this violation to happen very often. When combined with the caching I mentioned below it should be fine.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
avirtuos: Regardless of where we do the check, caching makes sense.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandlerTest.java
```diff
@@ -219,4 +227,71 @@ public void doPing()
         compositeHandler.handleRequest(allocator, req, new ByteArrayOutputStream(), objectMapper);
         verify(mockMetadataHandler, times(1)).doPing(any(PingRequest.class));
     }
+
+    @Test
+    public void checkBucketAuthZ()
```
avirtuos, line 232: I dont think this test makes sense on CompositeHandler, I think it belongs on MetadataHandlerTest

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
luc-liu: I'll move it to getSplits and add caching. Thanks for the comment, Anthony!

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandlerTest.java
luc-liu: There's no MetadataHandlerTest. Other concrete extensions of MetadataHander (such as DynamoDBMetadataHandler) are not in the SDK package, making unit tests for a private method in base class difficult.

avirtuos: ExampleMetadataHandlerTest in the sdk module

luc-liu: Unfortunately, ExampleMetadataHandlerTest is in package "package com.amazonaws.athena.connector.lambda.examples", while MetadataHandler is in "package com.amazonaws.athena.connector.lambda.handlers". The only test that is in the same package as Metadatahandler is CompositeHandlerTest.

avirtuos: I dont think the package matters, you'll setup a test case in ExampleMetadataHandlerTest which attempts to use a bad bucket. calling getSplits should then result in the error you expect.

luc-liu: Please see the revisions: simple caching is added to bucket authz checking, checking is moved to before getSplits, unit tests are moved to ExampleMetadataHandlerTest.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -115,6 +120,10 @@
     private final String spillBucket;
     private final String spillPrefix;
     private final String sourceType;
+    private static BucketState bucketState;
```
avirtuos, line 123: Why static? If this is because of something I said, that is my mistake. I think it is ok for this to be instance specific.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -446,6 +458,37 @@ public void onPing(PingRequest request)
...
+        }
+
+        AmazonS3 s3 = getS3();
+        Set<String> buckets = s3.listBuckets().stream().map(b -> b.getName()).collect(Collectors.toSet());
+        if (!buckets.contains(spillBucket)) {
+            bucketState = BucketState.INVALID;
+            throw new RuntimeException("You do NOT own the spill bucket with the name: " + spillBucket);
+        }
+
+        bucketState = BucketState.VALID;
```
avirtuos, line 489: Not all customers will want to use spill, we should only enforce this if they actually set a spill bucket. I also don't see where we are re-checking if the bucket changed. Instead of storing just the state, I think you should store both the state and the bucket that you checked. If the bucket changes you need to re-check it.

Additionally, I think we should extract this logic into a separate class called SpillLocationVerifier.

atennak1: pretty sure env var change causes instance re-creation so should be no need to re-check

avirtuos: That is 100% true for Lambda but this code wont always run in Lambda :)

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
luc-liu: OK, I'll make the changes

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
luc-liu: @avirtuos Can you give me a scenario as to how the spill bucket would be changed? 

Moving to a new class is a good idea, this would make unit tests cleaner and more accessible

avirtuos: customers can update the spill bucket any time they want. I'm not sure if this answers your question though so please elaborate if you need more context.

luc-liu: As far as I know, when an instance of any MetadataHandler child class is initialized, there's no way to update the spillbucket variable in the base class. So I'm not sure how a customer can change the spillbucket. Or, do you mean a customer can somehow change the SPILL_BUCKET_ENV after initialization? So in that case, any change in SPILL_BUCKET_ENV should be checked before performing getSplits?

avirtuos: Ah, i see what you mean. Fair point. My request still remains, the cache should be per-spill location. If the spill location changes, the cache of the validity check should expire. The core reason behind this request is that a future maintainer of the configuration (e.g. how we set the spill bucket) has no way to know that other parts of the code (e.g. your check) depend on the value of spill bucket never changing. Its a simple change to have the cache track the spill bucket value so my suggestion is for us to be defensive about it.

luc-liu: OK, thanks for the clarification, I'll push the new changes soon.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * Helper function to get the AmazonS3 instance of the lambda
+     */
+    @VisibleForTesting
+    AmazonS3 getS3()
+    {
+        return AmazonS3ClientBuilder.standard().build();
+    }
```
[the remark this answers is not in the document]
luc-liu: Is it necessary for SpillLocationVerifier to hold an instance of S3? Since it records the check status, most likely checking doesn't happen frequently, so S3 won't be needed that often. Do you worry that we might use this outside of Lambda, where the environment SpillLocationVerifier runs might not be able to build an S3 instance?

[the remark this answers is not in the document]
avirtuos: I think it's fine to hold it given the role of this component vs creating a new one each time. Creation of these clients is not cheap.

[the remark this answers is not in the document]
luc-liu: ah, didn't know that. I'll make the changes.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+    {UNCHECKED, VALID, INVALID}
+
+    private final AmazonS3 s3;
+    private String bucket;
+    private BucketState state;
+
+    /**
+     * @param bucket The name of the spill bucket.
+     */
+    public SpillLocationVerifier(AmazonS3 s3, String bucket)
```
avirtuos, line 48: Why are you requiring a bucket in the constructor? Sorry if I missed this in the last revision but it seems needless you should only use the bucket provided to checkBucketAuthZ(...)

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Public function used to check if the account that calls the lambda function owns the spill bucket
+     *
+     * @param spillBucket The name of the spill bucket.
+     */
+    public void checkBucketAuthZ(String spillBucket)
+    {
+        if (spillBucket == null) {
+            throw new RuntimeException("Spill bucket cannot be null!");
```
avirtuos, line 63: This shouldn't be a runtime exception, the bucket may be empty if the customer is not using spill.  null or empty bucket should cause you to return.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Helper function to check bucket ownership if it hasn't been checked before
+     *
+     */
+    @VisibleForTesting
+    void updateBucketState()
+    {
+        Set<String> buckets = s3.listBuckets().stream().map(b -> b.getName()).collect(Collectors.toSet());
```
avirtuos, line 86: What happens if this throws? Maybe we can surround with a try-catch and throw like the below:

throw new RuntimeException("Bucket ownership check failed for bucket XXXX.", causingException)

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+    {
+        if (state == BucketState.UNCHECKED) {
+            throw new RuntimeException("Bucket state should have been checked already.");
+        }
+        else if (state == BucketState.INVALID) {
+            throw new RuntimeException("You do NOT own the spill bucket with the name: " + bucket);
+        }
+        else {
+            // no op
+        }
```
avirtuos, line 113: No need for the else

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Helper function to throw exception if lambda function doesn't own the spill bucket
+     *
+     */
+    @VisibleForTesting
+    void passOrFail()
+    {
+        if (state == BucketState.UNCHECKED) {
```
avirtuos, line 105: how about using switch(state){
case UNCHECKED:
....
case INVALID:
...
}

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
luc-liu: The SpillLocationVerifier is used to verify the validity of the spill bucket, so if there's no spill bucket, there's no need to initialize a SpillLocationVerifier. Instead of assigning an empty string to "bucket" in SpillLocationVerifier, the user of this class should know what the bucket is. If the user doesn't know the bucket at initialization, they should consciously pass an empty string to the constructor of SpillLocationVerifier. IMO.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
luc-liu: I don't understand the purpose of letting the costumer setting this to null or empty when they don't need spill. I thought we implement bucket auth checking to guard against user error. In this case, the customer could pass in null or empty bucket accidentally without the intend of bypassing spill. My point is, why do we give the customer a free pass when they think their query will not spill (when it might).

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
luc-liu: public void checkBucketAuthZ(String spillBucket) is called inside a try block in MetadataHandler::handleRequest, the catch (Exception ex) should catch any exceptions

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
luc-liu: there's only one state that is of concern here - "BucketState.INVALID", "BucketState.UNCHECKED" shouldn't happen since validity checking always happens before, so using switch might be overkill

avirtuos: i think being defensive is better here and checking both states seems like a good way to save someone lots of debugging int he future :)

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
avirtuos: I'm saying you can throw a more useful error message because it will otherwise show up as an S3 list failure and the customer will be confused.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
avirtuos: Not all customers want to use spill, but when they do we want to ensure they are spilling to a location they own.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
avirtuos: You dont need the spill bucket in the constructor to do this. For example, when you create an S3 client you dont need to tell it at construction time what S3 key you plan on reading instead you do that when you call getObject. The same applies here, since the bucket you are being asked to check can change there is little point to taking it as a constructor argument.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
luc-liu: I think I'm starting to understand your point (as I was writing a rebuttal...:-)). Bucket ownership is a high risk issue, a customer's data privacy is paramount, we'd rather be defensive, so we check ownership when the customer put in a spill bucket. But if they leave the spill bucket field empty (empty String) or remove the spill bucket field in Lambda (null pointer), they choose not to use a spill bucket. If their query requires spilling, the worst that can happen is that Lambda will throw a bucket unauthz exception from the S3Spiller and there is no data security issue.

I hope we are on the same page?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
luc-liu: Ah, that makes sense

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
luc-liu: This makes sense. I'll remove the bucket from the constructor.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
avirtuos: not sure I 100% follow but if they don't set a spill bucket then we wont spill so there can't be a violation of the semantics this class is trying to enforce. This class is enforcing that you only write to spill buckets you own. it isn't trying to enforce that you have a spill bucket set. :)

luc-liu: I agree, SpillLocationVerifier should handle if the customer owns the bucket and nothing more. But as far as I know, SDK still spills even when the customer doesn't provide a spill bucket. The logic in the SDK is: if (size of data > in-memory capacity) spill happens, and the customer doesn't have control over the config for in-memory capacity. So I just want to point out that if the customer provides no spill bucket, spilling might still happen, and at that point, the SDK may throw an S3 exception.

luc-liu: I did some tests. In current configuration, if SpillLocationVerifier doesn't throw exception when spillbucket is null or empty, Athena console will show the following error: 

Your query has the following error(s):

GENERIC_INTERNAL_ERROR: BucketName cannot be empty/null

The SDK logs in CloudWatch went through without any warnings, and there was no spill. So even if the customer enter null or empty spillbucket, Athena still throws an error somewhere.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
```diff
@@ -0,0 +1,119 @@
...
+    private String bucket;
+    private BucketState state;
+
+    /**
+     * @param amazons3 The S3 object for the account.
+     */
+    public SpillLocationVerifier(AmazonS3 amazons3)
+    {
+        this.amazons3 = amazons3;
+        this.bucket = "";
```
avirtuos, line 51: why setting to empty string vs null?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/SpillLocationVerifier.java
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * @param spillBucket The name of the spill bucket.
+     */
+    public void checkBucketAuthZ(String spillBucket)
+    {
+        if (spillBucket == null || spillBucket.equals("")) {
+            return;
+        }
+
+        if (!bucket.equals(spillBucket)) {
```
avirtuos, line 66: per comments above maybe this should be changed to:
`java

if(bucket == null || !bucket.equals(spillBucket)`

avirtuos:
Congrats @luc-liu on your first commit! Thanks for addressing all the feedback.

alecswan:
@luc-liu, what was the reason for making this change which now requires the lambda function to have s3:ListAllMyBuckets permissions? Why is it not sufficient for the lambda function to have adequate CRUD permissions under an **s3 prefix/key** of a bucket?

avirtuos:
This change is used to ensure you own the bucket you are writing to. It helps avoid 'bucket name' sniping type security issues.

Issue #166, opened 2020-05-12
[BUG] Null 'List' within a Struct fails with all connectors


avirtuos:
**Describe the bug**
It is not possible to read or write a null List field value when the List is part of a Struct. 

`2020-04-16 18:00:30 XXXXXXX WARN  CompositeHandler:104 - handleRequest: Completed with an exception.
java.lang.RuntimeException: Error while processing field XXXXX
	at com.amazonaws.athena.connectors.docdb.DocDBRecordHandler.lambda$readWithConstraint$0(DocDBRecordHandler.java:158) ~[task/:?]
	at com.amazonaws.athena.connector.lambda.data.S3BlockSpiller.writeRows(S3BlockSpiller.java:174) ~[task/:?]
	at com.amazonaws.athena.connectors.docdb.DocDBRecordHandler.readWithConstraint(DocDBRecordHandler.java:136) ~[task/:?]
	at com.amazonaws.athena.connector.lambda.handlers.RecordHandler.doReadRecords(RecordHandler.java:191) ~[task/:?]
	at com.amazonaws.athena.connector.lambda.handlers.RecordHandler.doHandleRequest(RecordHandler.java:157) ~[task/:?]
	at com.amazonaws.athena.connector.lambda.handlers.CompositeHandler.handleRequest(CompositeHandler.java:135) ~[task/:?]
	at com.amazonaws.athena.connector.lambda.handlers.CompositeHandler.handleRequest(CompositeHandler.java:100) [task/:?]
	at lambdainternal.EventHandlerLoader$2.call(EventHandlerLoader.java:906) [LambdaSandboxJava-1.0.jar:?]
	at lambdainternal.AWSLambda.startRuntime(AWSLambda.java:341) [LambdaSandboxJava-1.0.jar:?]
	at lambdainternal.AWSLambda.<clinit>(AWSLambda.java:63) [LambdaSandboxJava-1.0.jar:?]
	at java.lang.Class.forName0(Native Method) ~[?:1.8.0_201]
	at java.lang.Class.forName(Class.java:348) [?:1.8.0_201]
	at lambdainternal.LambdaRTEntry.main(LambdaRTEntry.java:119) [LambdaJavaRTEntry-1.0.jar:?]
Caused by: java.lang.NullPointerException
	at com.amazonaws.athena.connector.lambda.data.BlockUtils.writeStruct(BlockUtils.java:584) ~[task/:?]
[...]

PR #167, opened 2020-05-12
[#166] Fixes NPE when attempting to read/write NULL value for a List field within a Struct

avirtuos:
[#166] Fixes NPE when attempting to read/write NULL value for a List field within a Struct

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/FieldResolver.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/FieldResolver.java
@@ -48,11 +48,11 @@ public interface FieldResolver
         public Object getFieldValue(Field field, Object value)
         {
             Types.MinorType minorType = Types.getMinorTypeForArrowType(field.getType());
-            if (minorType == Types.MinorType.LIST) {
-                return ((List) value).iterator();
-            }
-            else if (value instanceof Map) {
+            if (value instanceof Map) {
                 return ((Map<String, Object>) value).get(field.getName());
+            }
+            else if (minorType == Types.MinorType.LIST) {
+                return ((List) value).iterator();
             }
             throw new RuntimeException("Expected LIST type but found " + minorType);
         }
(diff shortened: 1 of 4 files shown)


athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/data/BlockTest.java
```diff
@@ -865,6 +866,117 @@ public void ListOfStructsTest()
...
+        ArrowRecordBatch batch = actualSerDe.deserialize(blockOut.toByteArray());
+
+        /**
+         * Generate and write the block
+         */
+        Block actualBlock = actualAllocator.createBlock(actualSchema);
+        actualBlock.loadRecordBatch(batch);
+        batch.close();
+
+        for (
```
soojinj, line 942: nit put in one line

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/FieldResolver.java
```diff
@@ -48,12 +48,12 @@
         public Object getFieldValue(Field field, Object value)
         {
             Types.MinorType minorType = Types.getMinorTypeForArrowType(field.getType());
-            if (minorType == Types.MinorType.LIST) {
```
soojinj, line 51: just curious why change the order?

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/data/BlockTest.java
soojinj: the for loop

avirtuos: yikes, auto-formatter 0wned me. Fixed now

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/FieldResolver.java
avirtuos: basically its impossible to get an iterator of this type from a Map so its never going to work. It was supposed to be reversed.

PR #190, opened 2020-06-12
Propagate more useful caller identity metadata to connectors

### LINKED ISSUE #26: FederatedIdentity is not always populated
Certain call patterns result in the principle arn portion of the FederatedIdentity not being set when Athena calls a connector or being set to "unknown". This is a known issue and one that we hope to resolve in the weeks after the preview opens to the public.

In general we are still gathering feedback on the best way to federate identity to connectors. Please feel free to share your thoughts or suggestions on this issue. The actual fix for ensuring this field is populated will be made within Athena, not within this repo.

sethtdenney:
Closes #26 

This changes the SDK to allow for more useful identity metadata to be propagated to connectors from Athena.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,51 @@
+name: "Code scanning - action"
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 12 * * 6'
+
+jobs:
+  CodeQL-Build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        # We must fetch at least the immediate parents so that if this is
+        # a pull request then we can checkout the head.
+        fetch-depth: 2
+
+    # If this run was triggered by a pull request event, then checkout
+    # the head of the pull request instead of the merge commit.
+    - run: git checkout HEAD^2
+      if: ${{ github.event_name == 'pull_request' }}
+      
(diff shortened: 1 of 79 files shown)


avirtuos:
did you approve your own pull request? i'll connect offline

avirtuos:
Per offline chat, I'm not sure this is backwards compatible at the wire format level. We may need to bump the protocol # to allow client/server to auto-negotiate capabilities. Please run a test of old connector with new Presto and see what happens. sync with @atennak1 once you think you know if it is/isn't backawards compatible. Akila may already know the answer to this too.

atennak1:
Think we should be ok because the old fields are retained in the SerDe

https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/FederatedIdentitySerDe.java#L61-L62

avirtuos:
But what 'eats' the new fields so they don't mess up the next object getting parsed? Is that being provided by ObjectMapper? or what?

atennak1:
@avirtuos https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/FederatedIdentitySerDe.java#L87

burhan94:
Test worked, confirmed that the changes are backwards compatible by doing the following:

1. set up cloudwatch connector through SAR
2. built presto locally and attached it to the lambda
3. ran a select * on all_log_streams
4. verified that results are same as when running on Athena console.

PR #180, opened 2020-05-26
Elasticsearch Connector

shurvitz:
*Description of changes:*
This is the single-split implementation of the Elasticsearch Connector.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #77, opened 2019-12-10
Athena ElasticSearch connector


YakobovLior:
Hello,

Are there are plans/requests for ElasticSearch connector for Athena? (not just AWS ElasticSearch service).

Thanks,
Lior

atennak1:
Hi @YakobovLior .  We would like to support this but haven't gotten around to it yet.  We can leave this open for others to +1 their interest to help us prioritize.  We're also happy to review a PR if anyone writes an ElasticSearch connector themselves!

avirtuos:
We've received a few requests for elastic search. I've not been tracking the AWS service separately from the native OS one. May do so once we start getting deeper and understand if the connector needs to be differentiated or just tested against both. We can use this issue to track demand. I'd recommend any other folks that are interested add their interest to this issue via +1 and comment.

avirtuos:
@sethtdenney when I'm back for vacation we should chat about this and see where this is on the priority list. We got asked about this a lot at reInvent.

avirtuos:
This work has started and we hope to have something to share / test in the next month.

chrishiste:
+1

I probably won't but If I write a connector I will share the code.

pvieito:
+1 here!

avirtuos:
Good news, our initial work on this connector is now in PR. There is still work to be done but happy to share some of the process.

pvieito:
Great! Unfortunately it seems to not be working with Elasticsearch 6.X:

Query:
```
describe federation_elasticsearch.exaple_domain.example_index;
```

Error:
```
java.lang.RuntimeException: com.amazonaws.services.lambda.invoke.LambdaFunctionException: Error processing request to map index (example_index): Elasticsearch exception [type=illegal_argument_exception, reason=request [/example_index/_mapping] contains unrecognized parameter: [ignore_throttled]]
```

YakobovLior:
Hey @atennak1, 
Thank you for the update, I will give a try with ElasticSearch 7.5.2 in a few days and update.

Lior

PR #194, opened 2020-06-18
Elasticsearch parallel scan support

shurvitz:
*Issue #, if available:*
#180 #77 
*Description of changes:*
1. Added support for parallel scans.

1. Added a client cache for client reusability.

1. Upgraded the testing platform.

### DIFF
--- a/athena-elasticsearch/README.md
+++ b/athena-elasticsearch/README.md
@@ -66,11 +66,17 @@ endpoint.
     `=` are used by this connector as separators for the domain-endpoint pairs. Therefor, they 
     should **NOT** be used anywhere inside the stored secret.
 
-5. **spill_bucket** - When the data returned by your Lambda function exceeds Lambda’s limits, 
+4. **query_timeout_cluster** - timeout period (in seconds) for Cluster-Health queries used in the
+generation of parallel scans.
+
+5. **query_timeout_search** - timeout period (in seconds) for Search queries used in the retrieval
+of documents from an index.
+
+6. **spill_bucket** - When the data returned by your Lambda function exceeds Lambda’s limits,
 this is the bucket that the data will be written to for Athena to read the excess from (e.g. 
 my_bucket).
 
-6. **spill_prefix** - (Optional) Defaults to sub-folder in your bucket called 
+7. **spill_prefix** - (Optional) Defaults to sub-folder in your bucket called
 'athena-federation-spill'. Used in conjunction with spill_bucket, this is the path within the 
 above bucket where large responses spill. You should configure an S3 lifecycle on this 
 location to delete old spills after X days/hours.
--- a/athena-elasticsearch/README.md
+++ b/athena-elasticsearch/README.md
@@ -175,11 +175,14 @@ will allow users with permission the ability to deploy instances of the connecto
 
 ## Performance
 
-The Athena Elasticsearch Connector does not currently support parallel scans but will attempt 
-to push down predicates as part of its document search queries (parallel scans will be supported 
-in the next development phase).
+The Athena Elasticsearch Connector supports shard-based parallel scans. Using cluster health information
+retrieved from the Elasticsearch instance, the connector generates multiple requests (for a document
+search query) that are split per shard and run concurrently.
 
-The following example demonstrates this connector's ability to utilize predicate push-down.
+**NOTE: Only primary shards that are active at the time of the query will be considered for parallel scans**.
+
+Additionally, the connector will push down predicates as part of its document search queries. The following
+example demonstrates this connector's ability to utilize predicate push-down.
 
 **Query:**
 ```sql
--- a/athena-elasticsearch/README.md
+++ b/athena-elasticsearch/README.md
@@ -185,8 +185,6 @@ The Athena Elasticsearch Connector supports shard-based parallel scans. Using cl
 retrieved from the Elasticsearch instance, the connector generates multiple requests (for a document
 search query) that are split per shard and run concurrently.
 
-**NOTE: Only primary shards that are active at the time of the query will be considered for parallel scans**.
-
 Additionally, the connector will push down predicates as part of its document search queries. The following
 example demonstrates this connector's ability to utilize predicate push-down.
 
--- a/athena-elasticsearch/athena-elasticsearch.yaml
+++ b/athena-elasticsearch/athena-elasticsearch.yaml
@@ -51,6 +51,12 @@ Parameters:
     Description: "List of one or more domain-names and their respective endpoints (including secret credentials) in the format: domain1=endpoint1,domain2=endpoint2,... (e.g. movies=https://${secret-credentials}:www.mymovies.com)."
     Default: ""
     Type: String
+  QueryTimeoutCluster:
+    Description: "timeout period (in seconds) for Cluster-Health queries used in the generation of parallel scans."
+    Type: Number
+  QueryTimeoutSearch:
+    Description: "timeout period (in seconds) for Search queries used in the retrieval of documents from an index."
+    Type: Number
 Resources:
   ConnectorConfig:
     Type: 'AWS::Serverless::Function'
@@ -62,6 +68,8 @@ Resources:
           spill_prefix: !Ref SpillPrefix
           auto_discover_endpoint: !Ref AutoDiscoverEndpoint
           domain_mapping: !Ref DomainMapping
+          query_timeout_cluster: !Ref QueryTimeoutCluster
+          query_timeout_search: !Ref QueryTimeoutSearch
       FunctionName: !Sub "${AthenaCatalogName}"
       Handler: "com.amazonaws.connectors.athena.elasticsearch.ElasticsearchCompositeHandler"
       CodeUri: "./target/athena-elasticsearch-1.0.jar"
--- a/athena-elasticsearch/athena-elasticsearch.yaml
+++ b/athena-elasticsearch/athena-elasticsearch.yaml
@@ -52,10 +52,12 @@ Parameters:
     Default: ""
     Type: String
   QueryTimeoutCluster:
-    Description: "timeout period (in seconds) for Cluster-Health queries used in the generation of parallel scans."
+    Description: "timeout period (in seconds) for Cluster-Health queries used in the generation of parallel scans (default is 10 seconds)."
+    Default: 10
     Type: Number
   QueryTimeoutSearch:
-    Description: "timeout period (in seconds) for Search queries used in the retrieval of documents from an index."
+    Description: "timeout period (in seconds) for Search queries used in the retrieval of documents from an index (default is 12 minutes)."
+    Default: 720
     Type: Number
 Resources:
   ConnectorConfig:
--- a/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
+++ b/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
@@ -118,10 +118,13 @@ public class AwsRestHighLevelClient
         ClusterHealthResponse response = cluster().health(request, RequestOptions.DEFAULT);
 
         if (response.isTimedOut()) {
-            throw new RuntimeException("Request timed out.");
+            throw new RuntimeException("Request timed out for index (" + index + ").");
         }
         else if (response.getActiveShards() == 0) {
-            throw new RuntimeException("There are no active shards.");
+            throw new RuntimeException("There are no active shards for index (" + index + ").");
+        }
+        else if (index == null || !response.getIndices().containsKey(index)) {
+            throw new RuntimeException("Request has an invalid index (" + (index == null ? "null" : index) + ").");
         }
 
         return response.getIndices().get(index).getShards().keySet();
--- a/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
+++ b/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
@@ -28,6 +28,8 @@ import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.CredentialsProvider;
 import org.apache.http.impl.client.BasicCredentialsProvider;
+import org.elasticsearch.action.admin.cluster.health.ClusterHealthRequest;
+import org.elasticsearch.action.admin.cluster.health.ClusterHealthResponse;
 import org.elasticsearch.action.admin.indices.alias.get.GetAliasesRequest;
 import org.elasticsearch.action.search.SearchRequest;
 import org.elasticsearch.action.search.SearchResponse;
@@ -38,6 +40,7 @@ import org.elasticsearch.client.RestClientBuilder;
 import org.elasticsearch.client.RestHighLevelClient;
 import org.elasticsearch.client.indices.GetMappingsRequest;
 import org.elasticsearch.client.indices.GetMappingsResponse;
+import org.elasticsearch.cluster.health.ClusterShardHealth;
 import org.elasticsearch.search.SearchHit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -55,6 +58,11 @@ public class AwsRestHighLevelClient
         extends RestHighLevelClient
 {
     private static final Logger logger = LoggerFactory.getLogger(AwsRestHighLevelClient.class);
+
+    /**
+     * Timeout period for the cluster health request (30 seconds).
+     */
+    private static final String CLUSTER_HEALTH_TIMEOUT = "30s";
 
     /**
      * Constructs a new client (using a builder) injected with credentials.
@@ -92,6 +100,33 @@ public class AwsRestHighLevelClient
         GetMappingsResponse mappingsResponse = indices().getMapping(mappingsRequest, RequestOptions.DEFAULT);
 
         return (LinkedHashMap<String, Object>) mappingsResponse.mappings().get(index).sourceAsMap();
+    }
+
+    /**
+     * Retrieves cluster-health information for shards associated with the specified index. The request will time out
+     * if response is not received within 30 seconds.
+     * @param index is used to restrict the request to a specified index.
+     * @return a Map of cluster Ids with their associated health information.
+     * @throws IOException if an error occurs while sending the request to the Elasticsearch instance, the request
+     * times out, or no active-primary shards are present.
+     */
+    public Map<Integer, ClusterShardHealth> getShardHealthInfo(String index)
+            throws IOException
+    {
+        ClusterHealthRequest request = new ClusterHealthRequest(index).timeout(CLUSTER_HEALTH_TIMEOUT);
+        // Set request to shard-level details
+        request.level(ClusterHealthRequest.Level.SHARDS);
+
+        ClusterHealthResponse response = cluster().health(request, RequestOptions.DEFAULT);
+
+        if (response.isTimedOut()) {
+            throw new IOException("Request timed out.");
+        }
+        else if (response.getActivePrimaryShards() == 0) {
+            throw new IOException("There are no active primary shards.");
+        }
+
+        return response.getIndices().get(index).getShards();
     }
 
     /**
--- a/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
+++ b/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
@@ -118,19 +118,6 @@ public class AwsRestHighLevelClient
     }
 
     /**
-     * Shuts down the client.
-     */
-    public void shutdown()
-    {
-        try {
-            close();
-        }
-        catch (IOException error) {
-            logger.error("Unable to shutdown client:", error);
-        }
-    }
-
-    /**
      * A builder for the AwsRestHighLevelClient class.
      */
     public static class Builder
--- a/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
+++ b/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
@@ -40,7 +40,7 @@ import org.elasticsearch.client.RestClientBuilder;
 import org.elasticsearch.client.RestHighLevelClient;
 import org.elasticsearch.client.indices.GetMappingsRequest;
 import org.elasticsearch.client.indices.GetMappingsResponse;
-import org.elasticsearch.cluster.health.ClusterShardHealth;
+import org.elasticsearch.common.unit.TimeValue;
 import org.elasticsearch.search.SearchHit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -50,6 +50,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 
 /**
  * This class is used to create a new REST client injected with either AWS credentials or username/password credentials.
@@ -58,11 +59,6 @@ public class AwsRestHighLevelClient
         extends RestHighLevelClient
 {
     private static final Logger logger = LoggerFactory.getLogger(AwsRestHighLevelClient.class);
-
-    /**
-     * Timeout period for the cluster health request (30 seconds).
-     */
-    private static final String CLUSTER_HEALTH_TIMEOUT = "30s";
 
     /**
      * Constructs a new client (using a builder) injected with credentials.
@@ -104,16 +100,18 @@ public class AwsRestHighLevelClient
 
     /**
      * Retrieves cluster-health information for shards associated with the specified index. The request will time out
-     * if response is not received within 30 seconds.
+     * if no results are returned after a period of time indicated by timeout.
      * @param index is used to restrict the request to a specified index.
-     * @return a Map of cluster Ids with their associated health information.
+     * @param timeout is the command timeout period in seconds.
+     * @return a set of shard ids for the specified index.
      * @throws IOException if an error occurs while sending the request to the Elasticsearch instance.
      * @throws RuntimeException if the request times out, or no active-primary shards are present.
      */
-    public Map<Integer, ClusterShardHealth> getShardHealthInfo(String index)
+    public Set<Integer> getShardIds(String index, long timeout)
             throws RuntimeException, IOException
     {
-        ClusterHealthRequest request = new ClusterHealthRequest(index).timeout(CLUSTER_HEALTH_TIMEOUT);
+        ClusterHealthRequest request = new ClusterHealthRequest(index)
+                .timeout(new TimeValue(timeout, TimeUnit.SECONDS));
         // Set request to shard-level details
         request.level(ClusterHealthRequest.Level.SHARDS);
 
@@ -122,11 +120,11 @@ public class AwsRestHighLevelClient
         if (response.isTimedOut()) {
             throw new RuntimeException("Request timed out.");
         }
-        else if (response.getActivePrimaryShards() == 0) {
-            throw new RuntimeException("There are no active primary shards.");
+        else if (response.getActiveShards() == 0) {
+            throw new RuntimeException("There are no active shards.");
         }
 
-        return response.getIndices().get(index).getShards();
+        return response.getIndices().get(index).getShards().keySet();
     }
 
     /**
--- a/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
+++ b/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
@@ -40,6 +40,7 @@ import org.elasticsearch.client.RestClientBuilder;
 import org.elasticsearch.client.RestHighLevelClient;
 import org.elasticsearch.client.indices.GetMappingsRequest;
 import org.elasticsearch.client.indices.GetMappingsResponse;
+import org.elasticsearch.cluster.health.ClusterHealthStatus;
 import org.elasticsearch.common.unit.TimeValue;
 import org.elasticsearch.search.SearchHit;
 import org.slf4j.Logger;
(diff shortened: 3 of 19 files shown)


athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
```diff
@@ -94,6 +102,33 @@ public AwsRestHighLevelClient(RestClientBuilder builder)
...
+        // Set request to shard-level details
+        request.level(ClusterHealthRequest.Level.SHARDS);
+
+        ClusterHealthResponse response = cluster().health(request, RequestOptions.DEFAULT);
+
+        if (response.isTimedOut()) {
+            throw new IOException("Request timed out.");
+        }
+        else if (response.getActivePrimaryShards() == 0) {
+            throw new IOException("There are no active primary shards.");
```
avirtuos, line 126: I dont think this is an IOException, Runtime is probably more appropriate.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
```diff
@@ -94,6 +102,33 @@ public AwsRestHighLevelClient(RestClientBuilder builder)
...
+            throws IOException
+    {
+        ClusterHealthRequest request = new ClusterHealthRequest(index).timeout(CLUSTER_HEALTH_TIMEOUT);
+        // Set request to shard-level details
+        request.level(ClusterHealthRequest.Level.SHARDS);
+
+        ClusterHealthResponse response = cluster().health(request, RequestOptions.DEFAULT);
+
+        if (response.isTimedOut()) {
+            throw new IOException("Request timed out.");
```
avirtuos, line 123: I don't think IoException is appropriate here, possibly Runtime or some other checked exception if you really want it to be checked.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
```diff
@@ -117,19 +152,6 @@ public SearchResponse getDocuments(SearchRequest request)
...
-     */
-    public void shutdown()
-    {
-        try {
-            close();
-        }
-        catch (IOException error) {
-            logger.error("Unable to shutdown client:", error);
-        }
-    }
```
avirtuos, line 131: Why are you removing this? Something needs to be able to shut these down eventually even if you are reusing the client.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClientFactory.java
```diff
@@ -33,38 +35,71 @@
...
+    public synchronized AwsRestHighLevelClient getClient(String endpoint)
+    {
+        AwsRestHighLevelClient client = clientCache.get(endpoint);
+
+        if (client == null) {
+            client = createClient(endpoint);
+            clientCache.put(endpoint, client);
+        }
+
+        return client;
```
avirtuos, line 87: you should run at least a basic 'test' on the client before returning it to ensure it is still valid. For example, some cheap/fast request. You might also need to have a way to remove a client from the cache incase it is causing requests to fail.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClientFactory.java
```diff
@@ -33,38 +35,71 @@
...
      * Constructs a new client factory (using a builder) that will create clients injected with credentials.
      * @param useAwsCredentials if true the factory create clients injected with AWC credentials.
-     *                              If false, it will create clients injected with username/password credentials.
+     *                          If false, it will create clients injected with username/password credentials.
      */
-    protected AwsRestHighLevelClientFactory(boolean useAwsCredentials)
+    public AwsRestHighLevelClientFactory(boolean useAwsCredentials)
     {
         this.useAwsCredentials = useAwsCredentials;
+        this.clientCache = new HashMap<>();
```
avirtuos, line 68: We shouldn't' allow clients to be unbounded size.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClientFactory.java
```diff
@@ -33,38 +35,71 @@
...
+    }
+
+    /**
+     * Attempts to get a REST client from the client cache synchronously using the endpoint as a key. If the client
+     * does not exist in the cache, createClient() is called to create a new client. The newly created client is
+     * then inserted into the cache and stored for later use.
+     * @param endpoint is the Elasticsearch instance endpoint.
+     * @return an Elasticsearch REST client.
+     */
+    public synchronized AwsRestHighLevelClient getClient(String endpoint)
```
avirtuos, line 78: rename to getOrCreate(...) which is a better description of what it does.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -83,6 +87,17 @@
     // A Map of the domain-names and their respective endpoints.
     private Map<String, String> domainMap;
 
+    /**
+     * Key used to store shard information in the Split's properties map (later used by the Record Handler).
+     */
+    private static final String SHARD_KEY = "shard";
```
avirtuos, line 93: probably needs to be protected so that the record handler can reference it?

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -83,6 +87,17 @@
...
+    /**
+     * Value used in combination with the shard ID to store shard information in the Split's properties map (later
+     * used by the Record Handler). The completed value is sent as a request preference to retrieve a specific shard
+     * from the Elasticsearch instance (e.g. "_shards:5" - retrieve shard number 5).
+     */
+    private static final String SHARD_VALUE = "_shards:";
```
avirtuos, line 99: probably needs to be protected so that the record handler can reference it?

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -179,9 +194,6 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
             catch (IOException error) {
                 throw new RuntimeException("Error retrieving indices: " + error.getMessage(), error);
             }
```
avirtuos, line 196: when reusing the client I think you need to consider 'destroying' the client if you encounter an exception from the client otherwise you might not know if the client is 'bad'. The hbase connector has some handling for this, we basically wrapp the client to catch and auto-retry with a new client: https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/connection/HBaseConnection.java

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -179,9 +194,6 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
...
-            finally {
-                client.shutdown();
```
avirtuos, line 183: who/what is closing clients if this is removed everywhere?

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -265,43 +274,59 @@ public void getPartitions(BlockWriter blockWriter, GetTableLayoutRequest request
...
+     * cluster/shard health information.
      */
     @Override
     public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest request)
             throws RuntimeException
     {
         logger.debug("doGetSplits: enter - " + request);
 
+        // Create set of splits
+        Set<Split> splits = new LinkedHashSet<>();
```
avirtuos, line 295: is order important?

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -265,43 +274,59 @@ public void getPartitions(BlockWriter blockWriter, GetTableLayoutRequest request
...
+                    ClusterShardHealth shardHealth = entry.getValue();
+                    // Process only primary active shards.
+                    if (shardHealth.isPrimaryActive()) {
+                        // Every split must have a unique location if we wish to spill to avoid failures
+                        SpillLocation spillLocation = makeSpillLocation(request);
+                        // Create a new split (added to the splits set) that includes the domain and endpoint, and
+                        // shard information (to be used later by the Record Handler).
+                        splits.add(new Split(spillLocation, makeEncryptionKey(), ImmutableMap
+                                .of(domain, endpoint, SHARD_KEY, SHARD_VALUE + shardId.toString())));
+                    }
```
avirtuos, line 317: What does it mean if this is false? should we fail the query? will we potentially miss data if we proceed without it?

athena-elasticsearch/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
avirtuos, line 1: What does this file do? can we add some comments?

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
shurvitz: Changed to RuntimeException.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
shurvitz: Changed to RuntimeException.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
shurvitz: shutdown() was a wrapper around the REST client's close() method. The wrapper was needed because Mockito V1.x could not mock final methods. Since I've upgraded to Mockito V3.x, this method is no longer needed.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClientFactory.java
shurvitz: Client health and age tests are now performed by the new CacheableAwsRestHighLevelClient before returning clients from the cache.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClientFactory.java
shurvitz: CacheableAwsRestHighLevelClient has a max size of 10 clients.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClientFactory.java
shurvitz: changed to getOrCreate(...)

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
shurvitz: Changed to protected and reused in Record Handler.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
shurvitz: Not needed in Record Handler. value is extracted using ElasticsearchMetadataHandler.SHARD_KEY as a key on the split's properties map and passed as is to the request's preferences argument.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
shurvitz: Client health is checked in CacheableAwsRestHighLevelClient before returning it from the cache. If the client fails the test, it is discarded from the cache.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
shurvitz: Clients are closed by the client cache upon eviction from the cache.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
shurvitz: Making the order deterministic allows me to ensure that the expect **shard** and **endpoint** values in the properties map are compared against the actual map values in the correct split. Otherwise, I could be comparing the expected information against the wrong split and failing the test.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
shurvitz: I use primary shards as a differentiator from replicated shards do avoid possible duplication. Shard health can be unpredictably in flux. A shard can be currently active, or in some unavailable state. Rather then wait for all shards to be active or fail the request, I think we should strive for eventual consistency and work with what we have. The assumption here is that some of the data is better than none. I do fail the order if none of the shards are active.

athena-elasticsearch/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
shurvitz: This is a configuration file used by Mockito to enable support for final methods. I don't think I can add a comment inside, so I'll add a README file.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
avirtuos: I dont think we can allow eventual consistency here. We need some way to know if every share has at least 1 active replica in the elastic-search cluster or we should fail the query. Is that possible?

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
avirtuos: I dont think this is a good reason for taking a dependency on ordering. Why can't the UT be based on a set and compare that way? there is no where in the SDK or even in Athena where order is preserved for splits.

avirtuos:
Still two unresolved comments from v1. I let them open for you. They are regarding the shard health and the ordering of splits for UT validation. I'll now look at your code changes in v2.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchRecordHandler.java
```diff
@@ -128,27 +135,36 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
                 do {
                     // Process the search request injecting the search-source, and setting the from position
                     // used for pagination of results.
                     SearchResponse searchResponse = client
                             .getDocuments(searchRequest.source(searchSource.from(currPosition)));
+
+                    // Throw on query timeout.
+                    if (searchResponse.isTimedOut()) {
+                        throw new IOException("Request for " + shard + " timed out.");
+                    }
```
avirtuos, line 166: im not sure this should be an IOException, also when does this happen?

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
shurvitz: Changed to HashSet.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
shurvitz: It seems that the shard-health query returns only one set of shards for the index without duplicates. Removed check for primary shards.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchRecordHandler.java
shurvitz: Not sure we need a timeout check here - removed.

shurvitz: Doesn't look like a timeout check is needed here - removed.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
```diff
@@ -56,6 +59,11 @@
 {
     private static final Logger logger = LoggerFactory.getLogger(AwsRestHighLevelClient.class);
 
+    /**
+     * Timeout period for the cluster health request (30 seconds).
+     */
+    private static final String CLUSTER_HEALTH_TIMEOUT = "30s";
```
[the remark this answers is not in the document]
shurvitz: Added two new env variables to configure the timeout periods of the cluster-health and search queries.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
```diff
@@ -94,6 +102,33 @@ public AwsRestHighLevelClient(RestClientBuilder builder)
...
+     * @throws RuntimeException if the request times out, or no active-primary shards are present.
+     */
+    public Map<Integer, ClusterShardHealth> getShardHealthInfo(String index)
+            throws RuntimeException, IOException
+    {
+        ClusterHealthRequest request = new ClusterHealthRequest(index).timeout(CLUSTER_HEALTH_TIMEOUT);
+        // Set request to shard-level details
+        request.level(ClusterHealthRequest.Level.SHARDS);
+
+        ClusterHealthResponse response = cluster().health(request, RequestOptions.DEFAULT);
```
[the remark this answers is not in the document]
shurvitz: This communicates directly with the cluster. The cluster will send back shard-level information for a specific index.

athena-elasticsearch/README.md
```diff
@@ -66,11 +66,17 @@ endpoint.
     `=` are used by this connector as separators for the domain-endpoint pairs. Therefor, they 
     should **NOT** be used anywhere inside the stored secret.
 
-5. **spill_bucket** - When the data returned by your Lambda function exceeds Lambda’s limits, 
+4. **query_timeout_cluster** - timeout period (in seconds) for Cluster-Health queries used in the
+generation of parallel scans.
+
+5. **query_timeout_search** - timeout period (in seconds) for Search queries used in the retrieval
+of documents from an index.
```
avirtuos, line 73: Are these optional? Are these optional? if not, they should be and we should offer sane default values. Also, can you explain the search timeout a bit more? what specific activity does this control? Is this the maximum duration of a search? e.g. any search taking longer than this will fail due to timeout? or is this about connection health and this even a 10 second timeout can be OK fo queries taking 100 minutes to complete.

athena-elasticsearch/athena-elasticsearch.yaml
```diff
@@ -51,6 +51,12 @@ Parameters:
     Description: "List of one or more domain-names and their respective endpoints (including secret credentials) in the format: domain1=endpoint1,domain2=endpoint2,... (e.g. movies=https://${secret-credentials}:www.mymovies.com)."
     Default: ""
     Type: String
+  QueryTimeoutCluster:
+    Description: "timeout period (in seconds) for Cluster-Health queries used in the generation of parallel scans."
+    Type: Number
+  QueryTimeoutSearch:
+    Description: "timeout period (in seconds) for Search queries used in the retrieval of documents from an index."
+    Type: Number
```
avirtuos, line 59: Please include default values.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
```diff
@@ -94,6 +98,35 @@ public AwsRestHighLevelClient(RestClientBuilder builder)
...
+
+        if (response.isTimedOut()) {
+            throw new RuntimeException("Request timed out.");
+        }
+        else if (response.getActiveShards() == 0) {
+            throw new RuntimeException("There are no active shards.");
```
avirtuos, line 124: include the index in the error. 

Also, what if a cluster has 5 shards and 1 of the shards is offline and no backup for that shard is available? what detects that situation?

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
```diff
@@ -94,6 +98,35 @@ public AwsRestHighLevelClient(RestClientBuilder builder)
...
+        }
+
+        return response.getIndices().get(index).getShards().keySet();
```
avirtuos, line 127: lots of potential for nulls here, isn't there?

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -83,6 +86,21 @@
     // A Map of the domain-names and their respective endpoints.
     private Map<String, String> domainMap;
 
+    // Env. variable that holds the query timeout period for the Cluster-Health queries.
+    private static final String QUERY_TIMEOUT_CLUSTER = "query_timeout_cluster";
+    private Long queryTimeout;
```
avirtuos, line 91: final?

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -117,6 +136,7 @@ protected ElasticsearchMetadataHandler(AWSGlue awsGlue,
         this.domainMap = this.domainMapProvider.getDomainMap(null);
         this.clientFactory = clientFactory;
         this.glueTypeMapper = new ElasticsearchGlueTypeMapper();
+        this.queryTimeout = new Long(30);
```
avirtuos, line 139: allow this to be set via constructor. Also, why is this a Long (boxed type) vs. long (primitive type)?

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchRecordHandler.java
```diff
@@ -64,6 +66,10 @@
     // is external to Amazon (false), and the domain_mapping environment variable should be used instead.
     private static final String AUTO_DISCOVER_ENDPOINT = "auto_discover_endpoint";
 
+    // Env. variable that holds the query timeout period for the Search queries.
+    private static final String QUERY_TIMEOUT_SEARCH = "query_timeout_search";
+    private Long queryTimeout;
```
avirtuos, line 71: final and should be long not Long. This site explains a bit about java 'Boxing' (e.g. Long vs long), incase that concept is new to you since Java is relatively new to you. https://www.baeldung.com/java-primitives-vs-objects

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchRecordHandler.java
```diff
@@ -78,6 +84,7 @@ public ElasticsearchRecordHandler()
         this.typeUtils = new ElasticsearchTypeUtils();
         this.clientFactory = new AwsRestHighLevelClientFactory(getEnv(AUTO_DISCOVER_ENDPOINT)
                 .equalsIgnoreCase("true"));
+        this.queryTimeout = new Long(getEnv(QUERY_TIMEOUT_SEARCH));
```
avirtuos, line 87: switch to Long.parseLong(getEnv(...));

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchRecordHandler.java
```diff
@@ -88,6 +95,7 @@ protected ElasticsearchRecordHandler(AmazonS3 amazonS3, AWSSecretsManager secret
 
         this.typeUtils = new ElasticsearchTypeUtils();
         this.clientFactory = clientFactory;
+        this.queryTimeout = new Long(30);
```
avirtuos, line 98: allow constructor to inject this.

athena-elasticsearch/README.md
shurvitz: Normally, if there's a communication/health issue, an exception will be thrown. However, if the query takes a significantly long time due to limited resources and/or high churn rate bringing the cluster to its knees, the timeout period will allow the connector to recover rather then be stuck waiting.
For the cluster-health request - 10 seconds should be enough.
For the Search request - 30 minutes sounds right (would be the same as the default timeout for Presto queries).

avirtuos: Given than lambda limits runtime to 15 minutes currently, we should probably set our timeout to be  less than 15 minutes (maybe 12 by default per-shard)

athena-elasticsearch/athena-elasticsearch.yaml
shurvitz: Set defaults (in seconds):
  QueryTimeoutCluster:
    Default: 10
  QueryTimeoutSearch:
    Default: 1800

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
shurvitz: Changed error messages.

I'm not sure how we could detect that a specific shard in not available (documentation on that is a bit sketchy). Irrespective, the query for that shard will either throw or time-out. It's better that way, because a shard may only be unavailable temporarily (a short period of time) and may recover in time.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
shurvitz: changed to final long.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchMetadataHandler.java
shurvitz: changed to final long and passed via constructor.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchRecordHandler.java
shurvitz: Changed to final long.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/AwsRestHighLevelClient.java
shurvitz: I'm assuming what you're referring to is the .get(index) which access's a member of a Map. The index is always specified in the request. The behavior that I've seen If the index does not exist in the cluster, is that the request will time-out (throwing a RuntimeException) and won't get to this point. But, just to be on the safe side, i'll include a check that the index is not null and exists in the Map.

shurvitz:
```java
        else if (index == null || !response.getIndices().containsKey(index)) {
            throw new RuntimeException("Request has an invalid index (" + (index == null ? "null" : index) + ").");
        }
```

PR #199, opened 2020-07-01
fixed issue with span function for null values

abhishekpradeepmishra:
*Issue #, if available:*

*Description of changes: When adding multiple ranges to constraints map, the ranges are merged to get correct high and low values. The merge operation has been failing for null values. Final range after merge has both high and low values as null*


abhishekpradeepmishra:
Will send it from fork master branch

PR #200, opened 2020-07-01
Fixed issue with span function for null values

abhishekpradeepmishra:
*Issue #, if available:*

*Description of changes: Multiple range values in constraint map are merged to get high and low values. Currently it produces both high and low values as null. Added code to do a null check inside Range.span function*


abhishekpradeepmishra:
Existing test cases breaking, Needs more testing

PR #217, opened 2020-07-16
Remove reflection code that doesn't work in DDB type inference

atennak1:
*Issue #, if available:* N/A

*Description of changes:* If user has an empty List we were trying to infer the element type using reflection, but due to type erasure that would never work, so removed it and added fallback to string/varchar.

### DIFF
--- a/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTypeUtils.java
+++ b/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTypeUtils.java
@@ -71,7 +71,7 @@ public final class DDBTypeUtils
      */
     public static Field inferArrowField(String key, Object value)
     {
-        logger.debug("inferArrowField invoked for key {} with value {} of class {}", key, value,
+        logger.debug("inferArrowField invoked for key {} of class {}", key,
                 value != null ? value.getClass() : null);
         if (value == null) {
             return null;
--- a/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTypeUtils.java
+++ b/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTypeUtils.java
@@ -71,6 +71,8 @@ public final class DDBTypeUtils
      */
     public static Field inferArrowField(String key, Object value)
     {
+        logger.debug("inferArrowField invoked for key {} with value {} of class {}", key, value,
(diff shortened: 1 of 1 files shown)


athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTypeUtils.java
```diff
@@ -71,6 +71,8 @@ private DDBTypeUtils() {}
      */
     public static Field inferArrowField(String key, Object value)
     {
+        logger.debug("inferArrowField invoked for key {} with value {} of class {}", key, value,
```
avirtuos, line 74: is it safe to log values? That might be a security issue

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTypeUtils.java
```diff
@@ -90,14 +92,8 @@ else if (value instanceof BigDecimal) {
...
-                try {
-                    Object subVal = ((Collection) value).getClass()
-                            .getTypeParameters()[0].getGenericDeclaration().newInstance();
-                    child = inferArrowField(key + ".child", subVal);
-                }
-                catch (IllegalAccessException | InstantiationException ex) {
-                    throw new RuntimeException(ex);
-                }
+                logger.warn("Automatic schema inference encountered empty List or Set {}. Unable to determine element types. Falling back to VARCHAR representation", key);
+                child = inferArrowField("", "");
```
avirtuos, line 96: we have almost this exact same bug in DocDB. We should think about how to centralize some of this logic. It might be challenging since all the types are different but an abstract with key method to implement per-connector (e.g. 'isEmptyList') might work. Its out of scope for your fix but wanted to call it out.

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTypeUtils.java
atennak1: removed value

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTypeUtils.java
atennak1: good call out, created https://github.com/awslabs/aws-athena-query-federation/issues/218

PR #220, opened 2020-07-21
Adding Vertica connector

rohit-masur:
Adding Vertica Connector


avirtuos:
Note for other reviewers. There is an open question on where the Vertical driver used by this connector will be hosted. It was in maven central at some point but the university that hosted it no longer provides the mirror. We have confirmation from Vertica sales/legal that it is ok to redistribute and host per krragan@

athena-vertica/athena-vertica.yaml
```diff
@@ -0,0 +1,102 @@
...
+    Default: 3008
+    Type: Number
+  DisableSpillEncryption:
+    Description: "WARNING: If set to 'true' encryption for spilled data is disabled."
+    Default: "false"
+    Type: String
+  VerticaConnectionString:
+    Description: 'The Vertica connection details to use by default if not catalog specific connection is defined and optionally using SecretsManager (e.g. ${secret_name}).'
+    Type: String
+    Default: "jdbc:vertica://<HOST_NAME>:<PORT_NAME>/<DB_NAME>?user=<USER>&password=<PASSWORD>"
```
avirtuos, line 55: Take a look at our docdb connector, we allow customers to provide part or all of the connection string using SecretsManager. This avoids then having sensitive data in the connection string. Below is an excerpt from the docdb readme which shows how it works. Most of the code you need is already provided by MetadataHandler...it has a function will will find and replace any 'secrets' in a string with the actual value from SecretsManager.

> To support these two SQL statements we'd need to add two environment variables to our Lambda function:
> 
> docdb_instance_1 - The value should be the DocumentDB connection details in the format of:mongodb://:@:/?ssl=true&ssl_ca_certs=rds-combined-ca-bundle.pem&replicaSet=rs0
> docdb_instance_2 - The value should be the DocumentDB connection details in the format of: mongodb://:@:/?ssl=true&ssl_ca_certs=rds-combined-ca-bundle.pem&replicaSet=rs0

> You can also optionally use SecretsManager for part or all of the value for the preceeding connection details.  For example, if I set a Lambda environment variable for docdb_instance_1 to be "mongodb://${docdb_instance_1_creds}@myhostname.com:123/?ssl=true&ssl_ca_certs=rds-combined-ca-bundle.pem&replicaSet=rs0" the Athena Federation SDK will automatically attempt to retrieve a secret from AWS SecretsManager named "docdb_instance_1_creds" and inject that value in place of "${docdb_instance_1_creds}". Basically anything between ${...} is attempted as a secret in SecretsManager. If no such secret exists, the text isn't replaced.

athena-vertica/athena-vertica.yaml
```diff
@@ -0,0 +1,102 @@
...
+  AWSSecretsManagerSecretName:
+    Description: "Secret Name of AWS Access Keys in AWS Secrets Manager"
+    Type: String
+    Default: "MyAwsCreds"
```
avirtuos, line 59: see above, you wont need this if you build on the example from the docdb connector.

athena-vertica/athena-vertica.yaml
```diff
@@ -0,0 +1,102 @@
...
+      Description: "A guided example for writing and deploying your own federated Amazon Athena connector for a custom source."
+      Runtime: java8
+      Timeout: !Ref LambdaTimeout
+      MemorySize: !Ref LambdaMemory
+      Policies:
+        - Statement:
+            - Action:
+                - athena:GetQueryExecution
+                - s3:ListAllMyBuckets
+                - lambda:InvokeFunction
```
avirtuos, line 92: shouldnt need this anymore

I also don't see access to read S3 objects that vertica writes, where is that handled?

athena-vertica/packaged.yaml
```diff
@@ -0,0 +1,111 @@
+Transform: AWS::Serverless-2016-10-31
```
avirtuos, line 1: This file should _not_ be committed. It gets generated when you publish.

athena-vertica/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+        <dependency>
+            <groupId>org.antlr</groupId>
+            <artifactId>ST4</artifactId>
+            <version>4.3.1</version>
+        </dependency>
+        <dependency>
+            <groupId>com.googlecode.json-simple</groupId>
+            <artifactId>json-simple</artifactId>
+            <version>1.1.1</version>
+        </dependency>
```
avirtuos, line 80: why json dependency? we already have jackson JSON parser in the dependency enclosure.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param connStr Vertica connection details
+     * @return A JDBC connection if the connection succeeded, else the function will throw.
+     */
+
+    public synchronized Connection getOrCreateConn(String connStr) throws SQLException
+    {
+        try { Class.forName("com.vertica.jdbc.Driver").newInstance();} catch
+        (ClassNotFoundException e){System.out.println("not found");} catch (IllegalAccessException | InstantiationException e) {
+            e.printStackTrace();
```
avirtuos, line 56: replace with logging, or in reality we shold probably just throw and allow this to bubble up.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+        Connection result = clientCache.get(connStr);
+
+        if (result == null || !connectionTest(result)) {
+            logger.info("conn: enter");
```
avirtuos, line 61: would be good to have a better log message, like: "Unable to reuse an existing connection, creating a new one."

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
```diff
@@ -0,0 +1,100 @@
...
+            result = DriverManager.getConnection(connStr);
+            clientCache.put(connStr, result);
```
avirtuos, line 63: is it worth testing the connection before we add it to the cache?

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Runs a 'quick' test on the connection and then returns it if it passes.
+     */
+
+    private boolean connectionTest(Connection conn)
+    {
+        try {
+
+            conn.isValid(5);
```
avirtuos, line 77: What does the 5 control? we should perhaps make this configurable and also move it to a private final member on the class.

Also, what does this method do under the hood? How does it check if the connection is valid?

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
```diff
@@ -0,0 +1,100 @@
...
+            logger.info("connectionTest: Testing connection completed - success.");
```
avirtuos, line 78: I dont think we need to log on success.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
```diff
@@ -0,0 +1,100 @@
...
+            return true;
+        }
+        catch (RuntimeException | SQLException ex) {
+            logger.warn("getOrCreateConn: Exception while testing existing connection.", ex);
+        }
+        logger.info("connectionTest: Testing connection completed - fail.");
```
avirtuos, line 84: if it failes we will already have logged above from the try/catch. I think we can remove this log line.

athena-vertica/athena-vertica.yaml
```diff
@@ -0,0 +1,102 @@
...
+    Type: String
+    Default: "jdbc:vertica://<HOST_NAME>:<PORT_NAME>/<DB_NAME>?user=<USER>&password=<PASSWORD>"
+  AWSSecretsManagerSecretName:
+    Description: "Secret Name of AWS Access Keys in AWS Secrets Manager"
+    Type: String
+    Default: "MyAwsCreds"
+  AWSRegion:
+    Description: "Region Name"
+    Type: String
+    Default: "US-WEST-2"
```
avirtuos, line 63: we dont usually include anything about the region. Why is this needed vs auto-detecting it?

athena-vertica/README.md
```diff
@@ -0,0 +1,4 @@
+## Vertica Athena Connector
+A custom connector that connects to Vertica database
```
avirtuos, line 2: I know this is an early feedback CR, just noting that we need to come back and add the readme.

athena-vertica/LICENSE.txt
```diff
@@ -0,0 +1 @@
+my license
```
avirtuos, line 1: please update the license to use the same license as the other connectors. It just avoids confusion when someone comes across the older commits in your branch. Best practice is to include the real license as early as possible. I know this is an early CR though, so no worries.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.util.*;
+import java.util.stream.Collectors;
+/**
+ * Query builder for database table split.
+ */
+public class VerticaJdbcSplitQueryBuilder {
+    private static final Logger LOGGER = LoggerFactory.getLogger(VerticaJdbcSplitQueryBuilder.class);
+    private static final String EXPORT_BUCKET = "export_bucket";
```
avirtuos, line 49: rename variable to EXPORT_BUCKET_KEY so maintainers know this is the key, not the actual bucket itself.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+    public String buildSql(
+            final String schema,
+            final String table,
+            final Schema tableSchema,
+            final Constraints constraints,
+            final String queryID)
+            throws SQLException {
+
+        //get the bucket where export results wll be uploaded
+        String s3ExportBucket = System.getenv(EXPORT_BUCKET);
```
avirtuos, line 81: take this as a constructor argument or paramenter for this function. We should limit where we use System.getenv(...) since in the future these connectors will run in more than just Lambda. That will make it easier to see all the places we need to repalce System.getenv(...). Said anotherway, System.getenv(...) should only be used from MetadataHandler and/or RecordHandler.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+        StringBuilder sql = new StringBuilder();
+
+        String exportSqL = "EXPORT TO PARQUET(" +
+                "directory = 's3://"+ s3ExportBucket + "/"+queryID.replace("-","")+ "'" +
+                ", Compression='snappy'\n" +
+                ", fileSizeMB=1000) \n" +
+                "AS \n";
```
avirtuos, line 88: String template will be in a future version? I see you have the dependency in the POM but was surprised to see this string concat.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+
+        List<String> clauses = toConjuncts(tableSchema.getFields(), constraints, accumulator);
+
+        if (!clauses.isEmpty()) {
+            sql.append(" WHERE ")
+                    .append(Joiner.on(" AND ").join(clauses));
+        }
+
+        //wrap the export around the generated sql
+        LOGGER.info("Generated SQL without the constraints : {}", sql.toString());
```
avirtuos, line 117: this says /without constraints but the constraints were added a few lines abov.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+
+        //Using StringTemplates to fill in the values of constraints
+        ST sqlTemplate = new ST(sql.toString());
```
avirtuos, line 120: now i'm really confused :) using string template but still doing a good amount of String concatenation above. Why? Let me know if you need some examples of how to use StringTemplate to build SQL. We have a few in-progress connectors that use it and I can share links with you internally.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+
+    protected String getFromClauseWithSplit(String schema, String table)
+    {
+        StringBuilder tableName = new StringBuilder();
+        if (!Strings.isNullOrEmpty(schema)) {
+            tableName.append(quote(schema)).append('.');
+        }
+        tableName.append(quote(table));
+
+        return String.format(" FROM %s ", tableName);
```
avirtuos, line 166: This is something we should be using StringTemplate for.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -0,0 +1,385 @@
...
+                    tableSchemaBuilder.addIntField(definition.getString("COLUMN_NAME"));
+                }
+                //If VARCHAR
+                if (definition.getString("TYPE_NAME").equalsIgnoreCase("Varchar")) {
+                    tableSchemaBuilder.addStringField(definition.getString("COLUMN_NAME"));
+                }
+                //If Numeric
+                if (definition.getString("TYPE_NAME").equalsIgnoreCase("NUMERIC")) {
+                    tableSchemaBuilder.addDecimalField(definition.getString("COLUMN_NAME"), 10, 2);
+                }
```
avirtuos, line 214: you've already got a todo here, I'd add that this should move into a new class, perhaps a VerticaSchemaUtils class.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -0,0 +1,385 @@
...
+            }
+
+        }
+        catch (SQLException e) {
+           logger.info("SQL Exception in doGetTable : "+ e.getMessage());
+        }
```
avirtuos, line 220: swallowing exceptions is bad practice. In this case you should rethrow the exception...perhaps as 

throw new RuntimeException("SQL Exception in doGetTable." + e.getMessage, e);

This way Athena will see the failure and show it to the customer. As the code currently stands, Athena may not know anyhting went wrong and may throw a weird (not helpful) error to the customer.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -0,0 +1,385 @@
...
+     *
+     * @param partitionSchemaBuilder The SchemaBuilder you can use to add additional columns and metadata to the
+     * partitions response.
+     * @param request The GetTableLayoutResquest that triggered this call.
+     */
+    @Override
+    public void enhancePartitionSchema(SchemaBuilder partitionSchemaBuilder, GetTableLayoutRequest request) {
+
+        logger.info("{}: Catalog {}, table {}", request.getQueryId(), request.getTableName().getSchemaName(), request.getTableName());
+        partitionSchemaBuilder.addField("preparedStmt", new ArrowType.Utf8());
```
avirtuos, line 239: nicely done.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -0,0 +1,385 @@
...
+        Constraints contr  = request.getConstraints();
+
+        try {
+            Connection connection = getOrCreateConn(request);
+
+            //Generate the SQL from user provided query
+            String colNames = schemaName.getFields().stream().map(Field::getName).collect(Collectors.joining(","));
+            sqlStmt.append(colNames);
+
+            String preparedStmt = verticaJdbcSplitQueryBuilder.buildSql(tableName.getSchemaName(), tableName.getTableName(), schemaName, contr, request.getQueryId());
```
avirtuos, line 268: This line is a bit long, can you format it as follows:

```java
String preparedStmt = verticaJdbcSplitQueryBuilder.buildSql(tableName.getSchemaName(), 
                                                            tableName.getTableName(), 
                                                            schemaName, 
                                                            contr, 
                                                            request.getQueryId());
```

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -0,0 +1,385 @@
...
+
+            // write the prepared statement to the partition column created in enhancePartitionSchema
+            blockWriter.writeRows((Block block, int rowNum) -> {
+                boolean matched;
+                matched = block.setValue("preparedStmt", rowNum, preparedStmt);
+                return matched ? 1 : 0;
+            });
```
avirtuos, line 275: you can simplify this to:

```java
blockWriter.writeRows((Block block, int rowNum) -> 
                block.setValue("preparedStmt", rowNum, preparedStmt) ? 1 : 0;
            );
```

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -0,0 +1,385 @@
...
+        } catch (SQLException e) {
+           logger.info("Exception in getPartitions: "+ e.getMessage());
+        }
```
avirtuos, line 278: similar comment to avoid, you should catch and rethrow as runtime otherwise Athena won't know this failed and can't give a good error message.

throw new RuntimeException("Exception in getPartitions: "+ e.getMessage(), e);

SQLException is a bit special in that it is a checked exception and you can't throw that directly in the SDK...you probably got some compilation errors about that because the other code in the SDK isn't expecting SQLException. In those cases, you can re-throw as RuntimeException.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -0,0 +1,385 @@
...
+     * @return A GetSplitsResponse which primarily contains:
+     * 1. A Set<Split> which represent read operations Amazon Athena must perform by calling your read function.
+     * 2. (Optional) A continuation token which allows you to paginate the generation of splits for large queries.
+     */
+    @Override
+    public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest request)
+    {
+        //ToDo: implement use of a continuation token to use in case of larger queries
+
+        logger.info("doGetSplits: " + request);
```
avirtuos, line 300: the request is already being logged by the SDK so this is redundant.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -0,0 +1,385 @@
...
+            temp.add("access_key", awsAccessId);
+            temp.add("secret_key", awsSecretKey);
+            String credS = temp.render();
+
+            PreparedStatement credPS = connection.prepareStatement(credS);
+            PreparedStatement exportSQL = connection.prepareStatement(sqlStatement);
+
+            // Set the AWS Region on Vertica
+            String defaultRegion = System.getenv(DEFAULT_REGION).toLowerCase();
+            String awsRegion =  "SELECT SET_CONFIG_PARAMETER('AWSRegion','"+defaultRegion+"') AS REGION_NAME";
```
avirtuos, line 348: We should move all of this to its own function. Also, is this all being done to provide Vertical with a set of credentials it can use to write to S3? If so, can we discuss this offline. as I have some questions about if customers will be comfortable with doing it this way as these credentials don't seem to expire. Notable, does Vertical support using STS credentials (key, id, token) which can be made to expire periodically and that the Lambda function already can get without the need for  Secretsmanager.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -0,0 +1,385 @@
...
+
+            //execute the query to set Region
+            ResultSet rSet = setAwsRegion.executeQuery();
+
+            //execute the query to export the data to S3
+            exportSQL.execute();
+
+        } catch (SQLException e) {
+            logger.info("Exception in doGetSplits executing the Prepared Statement " + e.getMessage());
+        }
```
avirtuos, line 363: you need to rethrow as RuntimeException.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -0,0 +1,385 @@
...
+
+        //generating the file name
+        String exportBucket = System.getenv("export_bucket");
+        String s3File = "s3://" + exportBucket + "/" + request.getQueryId().replace("-","") + ".parquet";
+
+        //creating the split
+        Split split = Split.newBuilder(makeSpillLocation(request), makeEncryptionKey())
```
avirtuos, line 370: looks like we are only exporting to a single file? Is this our first iteration and we are planning to do the multi-file export for better parallelism later?

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
```diff
@@ -0,0 +1,318 @@
...
+            builder.withExtractor(next.getName(), extractor);
+        }
+        GeneratedRowWriter rowWriter = builder.build();
+
+        /*
+         Using S3 Select to read the S3 Parquet files generated in the split
+         */
+
+        //List all the parquet files generated in the split
+        ObjectListing objectListing = amazonS3.listObjects(new ListObjectsRequest().withBucketName(exportBucket).withPrefix(id));
```
avirtuos, line 133: When we chatted this should be done inside getSplits so that we generate more than 1 split. The current code generates only 1 split and so will be limited to 15 minute read timeout. Let chat offline, i may not have done a good job explain this.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
```diff
@@ -0,0 +1,318 @@
...
+        for (S3ObjectSummary objectSummary : objectListing.getObjectSummaries()) {
+            if(objectSummary.getSize() > 0){
```
avirtuos, line 135: if objects are <1 what does it mean?

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
```diff
@@ -0,0 +1,318 @@
...
+
+                    BufferedReader streamReader = new BufferedReader(new InputStreamReader(resultInputStream, "UTF-8"));
+                    JSONArray jsonArray = new JSONArray();
+                    String inputStr;
+
+                    //map to contain the column name and value pair.
+                    HashMap<String, Object> map = new HashMap<>();
+
+                    while ((inputStr = streamReader.readLine()) != null) {
+                        JSONObject data = (JSONObject) jsonParser.parse(inputStr);
```
avirtuos, line 152: I thought we were getting parquet files from vertica?

athena-vertica/athena-vertica.yaml
rohit-masur: Used the above approach to resolve user credentials in connection string.

athena-vertica/packaged.yaml
rohit-masur: ok

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
rohit-masur: Replaced with RutTimeException

athena-vertica/LICENSE.txt
rohit-masur: updated.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
rohit-masur: done.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
rohit-masur: Done. passing the bucketname as a argument

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
rohit-masur: added the other datatypes and moved the functionality to a new class as suggested

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
rohit-masur: thank you. replaced with proper exception handling

athena-vertica/athena-vertica.yaml
rohit-masur: Removed the Lambda Invoke permission and added S3 permissions

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
rohit-masur: done.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
rohit-masur: removed.

athena-vertica/README.md
rohit-masur: thanks, will update this as we go ahead.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
rohit-masur: removed the logging.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
rohit-masur: done

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
rohit-masur: done.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
rohit-masur: Thank you. Replaced with proper exception handling.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
rohit-masur: removed.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
rohit-masur: Yes. This was to get creds to write to S3. I am working on using STS to get the user authenticated in Vertica. In the meantime, I have moved this to a different function.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
rohit-masur: done.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
rohit-masur: This was a left over code I forgot to delete. I do not use this file anywhere in the program flow. Removed this.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
rohit-masur: I have updated the code in my latest commit and now listing all the objects that vertica exported and building a split for each object. the record reader will then read the file in each split. Let's communicate internally to review this and make any changes necessary

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
rohit-masur: I was trying to say that if the file is not empty (file size > 0), then read it. I was handling a case where Vertica exports an empty file.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
rohit-masur: The AWS S3 SDK provides a [InputSerialization](https://docs.aws.amazon.com/AmazonS3/latest/API/API_InputSerialization.html) class for Parquet format, but an [OutputSerialization](https://docs.aws.amazon.com/AmazonS3/latest/API/API_OutputSerialization.html) only for CSV and JSON.  I am trying to use S3 Select to read the parquet file but have to serialize the result as a JSON to read it. Let's chat offline if we need to make changes to this.

athena-vertica/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>aws-athena-query-federation</artifactId>
+        <groupId>com.amazonaws</groupId>
+        <version>1.0</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>athena-vertica</artifactId>
```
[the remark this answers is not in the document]
rohit-masur: Thank you. Added the Vertica module to the parent POM.

avirtuos:
I'll take a 2nd pass at this tomorrow. Just updating here so you know that I didn't forget.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @Note Connection String format is expected to be like:
+ *  jdbc:vertica://<HOSTNAME></>>:5433/VerticaDB?user&password>
+ */
+
+public class VerticaConnectionFactory
+{
+    private static final Logger logger = LoggerFactory.getLogger(VerticaConnectionFactory.class);
+    private final Map<String, Connection> clientCache = new HashMap<>();
+    private final int CONNECTION_TIMEOUT_TIME_IN_SECONDS = 5;
```
avirtuos, line 43: I dont think we should hard code this, it should be configurable.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
```diff
@@ -0,0 +1,339 @@
...
+            final String schema,
+            final String table,
+            final Schema tableSchema,
+            final Constraints constraints,
+            final String queryID)
+
+    {
+        ST exportSqlST = new ST("EXPORT TO PARQUET(" +
+                "directory = 's3://<s3ExportBucket>/<queryID>'," +
+                " Compression='snappy', fileSizeMB=1000) AS ");
```
avirtuos, line 81: can we move this to a stg file in the class path so we don't have the long strings in the code? You can put a file called 'Vertica.stg' in the your app and then just use STGroupFile(...) to load that file from the class path.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
```diff
@@ -0,0 +1,320 @@
...
+        public HashMap<String, Object> getNameValue() {
+            return this.nameValue;
+        }
+    }
+
+
+    /*
+    Method to create the Parquet read request
+     */
+    private static SelectObjectContentRequest generateBaseParquetRequest(String bucket, String key)
```
avirtuos, line 301: Is this using S3 select? I dont think we should use S3 select, especially since you are doing a Select *... are we doing this just to avoid having parquet dependencies in our code? What format does S3 select use for results?

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaSchemaUtils.java
```diff
@@ -0,0 +1,109 @@
...
+        {
+            logger.info("Building the schema for table {} ",name);
+            SchemaBuilder tableSchemaBuilder = SchemaBuilder.newBuilder();
+            DatabaseMetaData dbMetadata = connection.getMetaData();
+            ResultSet definition = dbMetadata.getColumns(null, name.getSchemaName(), name.getTableName(), null);
+            while(definition.next())
+            {
+                //todo: review this again and add the Vertica data types missed
+                //If Bit
+                if (definition.getString("TYPE_NAME").equalsIgnoreCase("BIT"))
```
avirtuos, line 50: can we turn this into if/else or use continue; to skip the rest of the checks once we find one that matches? otherwise this is going to check a bunch of things needlessly.

athena-vertica/athena-vertica.yaml
```diff
@@ -0,0 +1,118 @@
...
+      Description: "A guided example for writing and deploying your own federated Amazon Athena connector for a custom source."
+      Runtime: java8
+      Timeout: !Ref LambdaTimeout
+      MemorySize: !Ref LambdaMemory
+      Policies:
+        - Statement:
+            - Action:
+                - athena:GetQueryExecution
+                - s3:ListAllMyBuckets
+                - secretsmanager:GetSecretValue
```
avirtuos, line 97: We can't request * for GetSecretValue. Please look at the docdb connetor or hbase connector for how to structure the yaml for secrets manager.

athena-vertica/athena-vertica.yaml
```diff
@@ -0,0 +1,118 @@
...
+                - ec2:CreateNetworkInterfaces
+                - sts:GetSessionToken
```
avirtuos, line 99: Why do we need this? Please look at the docdb connector (copy pasted below) as I think the perms you need here are too broad:

```
      Policies:
        - Statement:
            - Action:
                - secretsmanager:GetSecretValue
              Effect: Allow
              Resource: !Sub 'arn:aws:secretsmanager:*:*:secret:${SecretNameOrPrefix}'
          Version: '2012-10-17'
        - Statement:
            - Action:
                - glue:GetTableVersions
                - glue:GetPartitions
                - glue:GetTables
                - glue:GetTableVersion
                - glue:GetDatabases
                - glue:GetTable
                - glue:GetPartition
                - glue:GetDatabase
                - athena:GetQueryExecution
                - s3:ListAllMyBuckets
              Effect: Allow
              Resource: '*'
          Version: '2012-10-17'
        #S3CrudPolicy allows our connector to spill large responses to S3. You can optionally replace this pre-made policy
        #with one that is more restrictive and can only 'put' but not read,delete, or overwrite files.
        - S3CrudPolicy:
            BucketName: !Ref SpillBucket
        #VPCAccessPolicy allows our connector to run in a VPC so that it can access your data source.
        - VPCAccessPolicy: {}
```

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -0,0 +1,386 @@
...
+     * and executes the queries
+     */
+    private void executeQueriesOnVertica(Connection connection, String sqlStatement)
+    {
+        try
+        {
+            //todo: find a way to work with STS credentials
+            //get the AWS Keys from Secrets Manager
+            String awsAccessId = resolveSecrets("${access_key}");
+            String awsSecretKey = resolveSecrets("${secret_key}");
```
avirtuos, line 333: Does this mean the customer needs to have a Secret in Secretsmanager calls access_key? Perhaps this is just a placeholder for the STS idea you had?

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
avirtuos: I didn't realize we were planning to use S3 Select. Let's connect offline about this. I'm pretty sure this will be slower than if we did it natively ourselves since we will always being doing a select *.

avirtuos:
Any updates here?

athena-vertica/pom.xml
rohit-masur: removed

athena-vertica/athena-vertica.yaml
rohit-masur: removed

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaSchemaUtils.java
rohit-masur: converted to case statement

athena-vertica/athena-vertica.yaml
rohit-masur: changed to more granular permissions

athena-vertica/athena-vertica.yaml
rohit-masur: changed to more granular permissions

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
rohit-masur: Removed this. The Instance Profiles of the EC2 instance hosting Vertica can be given permissions to the export S3 bucket

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
rohit-masur: Yes. Using S3 Select to avoid the parquet/avro/hadoop dependencies. Including these causes the app size > 50MB and also causes dependency conflicts. Still researching a way to read parquet files natively.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
rohit-masur: Can you internally share some examples/demo of using STGroupFile? Thank you.

rohit-masur:
Pushed updated code which includes more granular permissions to the serverless application and some other code changes. Added some Unit Tests (still work in progess)

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
avirtuos: take a look at the Timestream connector - https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-timestream/src/main/java/com/amazonaws/connectors/timestream/query/QueryFactory.java

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
avirtuos: Did we make any progress on this? If S3 select is our only option then I suppose we can simply note it in the docs. Is the customer you are working with using the connector with S3 Select too?

avirtuos (changes requested):

athena-vertica/athena-vertica.yaml
```diff
@@ -75,37 +72,50 @@ Resources:
...
               Ref: SpillBucket
-          #S3CrudPolicy allows our connector to spill large responses to S3. You can optionally replace this pre-made policy
-          #with one that is more restrictive and can only 'put' but not read,delete, or overwrite files.
-        - S3CrudPolicy:
+        - S3WritePolicy:
+            BucketName:
+              Ref: SpillBucket
+        - S3ReadPolicy:
+            BucketName:
+              Ref: ExportBucket
```
avirtuos, line 113: i think we should call this VerticaExportBucket

athena-vertica/athena-vertica.yaml
```diff
@@ -75,37 +72,50 @@ Resources:
...
+        - S3WritePolicy:
             BucketName:
               Ref: ExportBucket
-        #VPCAccessPolicy allows our connector to run in a VPC so that it can access your data source.
+      #VPCAccessPolicy allows our connector to run in a VPC so that it can access your data source.
+      VpcConfig:
+        SecurityGroupIds:
+          - !Ref LambdaSecurityGroup
```
avirtuos, line 120: Why is this needed? can we add some comments?

avirtuos (changes requested):

avirtuos:
Any update? I have a few comments posted and there also seems to be failing build issues

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
rohit-masur: used StringTemplates

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaJdbcSplitQueryBuilder.java
rohit-masur: Updated to use STGroupFile(..)

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -145,8 +142,10 @@ public ListSchemasResponse doListSchemaNames(BlockAllocator allocator, ListSchem
             Connection client = getConnection(request);
             DatabaseMetaData dbMetadata = client.getMetaData();
             ResultSet rs  = dbMetadata.getTables(null, null, null, TABLE_TYPES);
+
             while (rs.next())
             {
+                logger.info(rs.getString(TABLE_SCHEMA));
```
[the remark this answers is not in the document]
rohit-masur: removed

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaMetadataHandler.java
```diff
@@ -254,18 +255,24 @@ public void getPartitions(BlockWriter blockWriter, GetTableLayoutRequest request
         //get the bucket where export results wll be uploaded
         String s3ExportBucket = System.getenv(EXPORT_BUCKET_KEY);
         //build the SQL query
-
+        Random r = new Random();
+        int randomInt = r.nextInt(100) + 1;
+        String queryID = request.getQueryId().replace("-","").concat(String.valueOf(randomInt));
```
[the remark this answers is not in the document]
rohit-masur: Added code to use UUID.randomUUID().toString().
Needed this in case user issues multiple federated queries in a single query. Current design uses the Athena QueryId  as names for the S3 files generated by export. In case of multiple exports, the current design will cause conflicts.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
```diff
@@ -135,90 +138,79 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
         SelectObjectContentRequest request = generateBaseParquetRequest(exportBucket, s3ObjectKey);
-        try (SelectObjectContentResult result = amazonS3.selectObjectContent(request))
-        {
+        try (SelectObjectContentResult result = amazonS3.selectObjectContent(request)) {
             InputStream resultInputStream = result.getPayload().getRecordsInputStream();
-            JSONParser jsonParser = new JSONParser();
-
             BufferedReader streamReader = new BufferedReader(new InputStreamReader(resultInputStream, StandardCharsets.UTF_8));
             String inputStr;
+            logger.info(streamReader.lines().toString());
```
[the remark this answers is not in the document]
rohit-masur: removed

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
```diff
@@ -135,90 +138,79 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
         Types.MinorType fieldType = mapOfNamesAndTypes.get(fieldName);
+        logger.info(String.valueOf(fieldType));
         switch (fieldType)
         {
             case BIT:
                 return (BitExtractor) (Object context, NullableBitHolder dst) ->
                 {
                     Object value = ((RowContext) context).getNameValue().get(fieldName);
-                    dst.isSet = 1;
+                    dst.isSet = 0;
```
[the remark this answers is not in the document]
rohit-masur: I had removed them it during some of my tests, forgot to revert. Added back in.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
```diff
@@ -135,90 +138,79 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
             case TINYINT:
                 return (TinyIntExtractor) (Object context, NullableTinyIntHolder dst) ->
                 {
                     Object value = ((RowContext) context).getNameValue().get(fieldName);
-                    dst.isSet = 1;
+                    dst.isSet = 0;
                     dst.value = Byte.parseByte(value.toString());
                 };
             case SMALLINT:
                 return (SmallIntExtractor) (Object context, NullableSmallIntHolder dst) ->
```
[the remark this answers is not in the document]
rohit-masur: I had removed them it during some of my tests, forgot to revert. Added back in.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
```diff
@@ -135,90 +138,79 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
                 {
                     Object value = ((RowContext) context).getNameValue().get(fieldName);
                     dst.value = Short.parseShort(value.toString());
-                    dst.isSet = 1;
+                    dst.isSet = 0;
```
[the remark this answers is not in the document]
rohit-masur: I had removed them it during some of my tests, forgot to revert. Added back in.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
rohit-masur: Currently, S3 select is the only option. I have mentioned this in the Readme file. Yes, customer is also using S3 Select.

athena-vertica/athena-vertica.yaml
rohit-masur: SecurityGroup that should be applied to the Connector Lambda function to enable it to connect to the VPC of Vertica database

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
rohit-masur: Removed this. Unnecessary to test every time.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
rohit-masur: removed

athena-vertica/athena-vertica.yaml
```diff
@@ -52,21 +52,19 @@ Parameters:
   SecurityGroupIds:
     Description: 'One or more SecurityGroup IDs corresponding to the SecurityGroup that should be applied to the Lambda function. (e.g. sg1,sg2,sg3)'
     Type: 'List<AWS::EC2::SecurityGroup::Id>'
+    Default: 'sg-0c36c83ed0b5b509d'
```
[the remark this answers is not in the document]
rohit-masur: I think this was a older commit. Removed the defaults

athena-vertica/athena-vertica.yaml
```diff
@@ -20,7 +20,7 @@ Parameters:
     Description: 'The name you will give to this catalog in Athena. It will also be used as the function name. This name must satisfy the pattern ^[a-z0-9-_]{1,64}$'
     Type: String
     AllowedPattern: ^[a-z0-9-_]{1,64}$
-    Default: "v2"
+    Default: "v328"
```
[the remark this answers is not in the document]
rohit-masur: I think this was a older commit. Removed the defaults

athena-vertica/athena-vertica.yaml
```diff
@@ -52,21 +52,19 @@ Parameters:
   SecurityGroupIds:
     Description: 'One or more SecurityGroup IDs corresponding to the SecurityGroup that should be applied to the Lambda function. (e.g. sg1,sg2,sg3)'
     Type: 'List<AWS::EC2::SecurityGroup::Id>'
+    Default: 'sg-0c36c83ed0b5b509d'
   SubnetIds:
     Description: 'One or more Subnet IDs corresponding to the Subnet that the Lambda function can use to access you data source. (e.g. subnet1,subnet2)'
     Type: 'List<AWS::EC2::Subnet::Id>'
+    Default: 'subnet-72e59e39'
```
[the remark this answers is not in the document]
rohit-masur: I think this was a older commit. Removed the defaults

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
avirtuos: Why is it unnessesary? I was suggesting testing it more often, not less

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaRecordHandler.java
avirtuos: We have a new option that was just released in lake formation which is more appropriate than S3 Select. It even returns Apache Arrow format already. We can look at moving to that post GA. Its currently in Preview only.  I've raised an issue, #325 , so we dont forget this.

athena-vertica/athena-vertica.yaml
avirtuos: Ah, gotcha.

avirtuos:
@rohit-masur Looks like you accidentally attached the vertica client jar to this PR. We do not accept jars into this repo unless there is a compelling reason to do so. I'm also not certain you (or by extension we) would have the appropriate licensing to redistribute that JAR. please confirm and/or remove the JAR.

rohit-masur:
Removed the JARs

rohit-masur:
janakag@ and I have raised a ticket internally to host the drivers in out Maven repo. Will update here once its done.

rohit-masur:
Vertica has hosted the drivers in the Maven Central repo. I have changed the POM to refer to the Vertica drivers hosted on Maven.

athena-vertica/sar_bucket_policy.json
```diff
@@ -0,0 +1,13 @@
+{
```
avirtuos, line 1: I'm not sure this file should be comitted, isn't this a generated file?

Similarly, should athena-vertica/av.jpg  be comitted or was that also a mistake?

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
avirtuos: any update here?

avirtuos (changes requested):
minor changes, 2 files may need to be removed.

athena-vertica/sar_bucket_policy.json
rohit-masur: Will remove the sar_bucket_policy.json file. 
av.jpg refers to the connector architecture diagram in the ReadMe file. I think we should keep this as this is the first connector where we Export the data to S3 instead of getting it back through the JDBC bridge.

athena-vertica/src/main/java/com/amazonaws/connectors/athena/vertica/VerticaConnectionFactory.java
rohit-masur: sorry for the misunderstanding, I think we should test the connection's validity every time we make a call to get the JDBC connection from the methods in MetadataHandler. Added the code back and moved the timeout variable as a private final member of the class

avirtuos (approved):

avirtuos:
@rohit-masur per my chat with Janak (athena-pm) I am merging this connector. Congrats! and thanks for the contribution. In parallel, Janak asked that you work on the integration test suite and instructions for how we can stand up a Vertica instance in our integration test accounts.

PR #226, opened 2020-07-31
Replace bogus license

ajorg-aws:
It's not clear if this subdir should actually be under MIT-0 but there are license headers in some of the files that specify Apache-2.0. If they should be MIT-0, let's do that instead.


avirtuos:
Thanks for catching this, no clue how we missed it for so long.

Issue #231, opened 2020-08-05
[BUG] Multi-array STRUCT, STRUCT<ARRAY<>, ARRAY<>> w/Glue metadata : with connector DynamoDB


shurvitz:
**Describe the bug**
When defining a STRUCT containing multiple ARRAY attributes, a subsequent select will return only the first ARRAY in the STRUCT.

**To Reproduce**
1. Define a table `movies` in DynamoDB with a corresponding Glue table that has the following column type: `STRUCT<actors:ARRAY<STRING>,genre:ARRAY<STRING>>`.
2. Send query: `select * from movies;`
3. Results will show only values for **actors**.  

 **Expected behavior**
Both **actors** and **genre** should appear

**Additional context**
Glue parser misinterprets the end token of the first ARRAY as the end token of the STRUCT resulting with the following Field definition: `namelist: Struct<actors: List<actors: Utf8>>`.
https://github.com/awslabs/aws-athena-query-federation/blob/408a6f6723e07319155aa27a67f4bde9aa83cd3a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/GlueFieldLexer.java#L111

shurvitz:
This issue was found with the DynamoDB connector, however, it is a generic issue and affects any connector using supplemental Glue metadata definitions with the above criteria.

shurvitz:
Fixed by #232

PR #232, opened 2020-08-05
[#231] Glue lex support for multi-array struct STRUCT<ARRAY<>, ARRAY<>>.

shurvitz:
**Issue #, if available:** #231

**Description of changes:** Added support for supplemental Glue metadata when parsing a field defined as a STRUCT containing multiple arrays attributes (e.g. `STRUCT<actors:ARRAY<STRING>,genre:ARRAY<STRING>>`).

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/GlueFieldLexer.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/GlueFieldLexer.java
@@ -112,6 +112,14 @@ public class GlueFieldLexer
             Field child = lex(parser.next(), parser, mapper);
             fieldBuilder.addField(child);
             if (Types.getMinorTypeForArrowType(child.getType()) == Types.MinorType.LIST) {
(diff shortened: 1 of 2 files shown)


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/GlueFieldLexer.java
```diff
@@ -111,6 +111,9 @@ else if (startToken.getValue().toLowerCase().equals(LIST)) {
         while (parser.hasNext() && parser.currentToken().getMarker() != GlueTypeParser.FIELD_END) {
             Field child = lex(parser.next(), parser, mapper);
             fieldBuilder.addField(child);
+            if (Types.getMinorTypeForArrowType(child.getType()) == Types.MinorType.LIST) {
+                parser.next();
```
atennak1, line 115: i think a comment will help to explain why we need to consume an extra token

shurvitz: Added clarification comment in code.

Issue #240, opened 2020-08-25
[FEATURE] Expose KMS Key ID param at SAR level and add additional permissions to support S3 encrypted via AWS-KMS keys.


ancamillo:
**Is your feature request related to a problem? If yes, please describe.**
I would like to programmatically instantiate the AthenaDynamoDBConnector via a Cloudformation template and configure the connector to use an S3 bucket using AWS-KMS encryption. At the moment the connector does not expose at SAR level the kms_key_id parameter but it must be created manually after the deploy on the lambda created by the SAR. In addition, the current policy associated to the lambda requires the addition of the kms:GenerateDataKey and kms:GenerateRandom permissions to work.

**Describe the solution you'd like**
Expose the kms_key_id param at SAR level as optional parameter and add missing permission to lambda role so that the connector can be instantiated automatically via cf without the need of additional manual actions after the deploy.

**Describe alternatives you've considered**
Manually add kms_key_id env variable and required permissions to the lambda after deploy.

**Additional context**
None

PR #241, opened 2020-08-25
Expose KMS Key ID param at SAR level and add additional permissions to support S3 encrypted via AWS-KMS keys for AthenaDynamoDBConnector.

ancamillo:
*Issue #, if available: https://github.com/awslabs/aws-athena-query-federation/issues/240

*Description of changes: Adds the KMSKeyId optional parameter to the AthenaDynamoDBConnector Serverless app which is conditionally passed as an env var to the connector lambda only if not empty as kms_key_id. Adds kms:GenerateDataKey and kms:GenerateRandom to the lambda policy to support S3 AWS-KMS encryption.


athena-dynamodb/athena-dynamodb.yaml
```diff
@@ -70,6 +78,8 @@ Resources:
                 - glue:GetDatabase
                 - athena:GetQueryExecution
                 - s3:ListAllMyBuckets
+                - kms:GenerateDataKey
+                - kms:GenerateRandom
```
avirtuos, line 82: This is too broad. Can we add a conditional policy that grants this only on the specified key?

ancamillo: Ok good point, addressed it. I had to explicitly define a role and policies instead of using the Policy property of connector function which does not support conditional definition (if KMSKeyId is not defined). It is a bit longer, mainly because I had to expand the SAM S3CrudPolicy but I believe it is the only clean way to do it.

athena-dynamodb/athena-dynamodb.yaml
```diff
@@ -45,35 +52,94 @@ Resources:
...
+            Effect: Allow
+            Resource:
+              - Fn::Sub:
+                - arn:${AWS::Partition}:s3:::${bucketName}
+                - bucketName:
+                    Ref: SpillBucket
+              - Fn::Sub:
+                - arn:${AWS::Partition}:s3:::${bucketName}/*
+                - bucketName:
+                    Ref: SpillBucket
```
avirtuos, line 124: Why did we change this?

athena-dynamodb/athena-dynamodb.yaml
```diff
@@ -45,35 +52,94 @@ Resources:
...
+        - "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
+      AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Principal:
+              Service:
+                - lambda.amazonaws.com
+            Action:
+              - "sts:AssumeRole"
```
avirtuos, line 78: Why did we change this?

athena-dynamodb/athena-dynamodb.yaml
ancamillo: This is the S3CrudPolicy but the S3CrudPolicy is an AWS SAM Policy Templates (https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-policy-template-list.html#s3-crud-policy) and can't be referenced in a AWS::IAM::Policy so I basically had to copy paste its content into the policy. I had to use an AWS::IAM::Policy because it is the only way I see to conditionally create it or not based on if KMSKeyId is passed or not. So basically from line 105 to line 124 is equivalent to the previous code:
        - S3CrudPolicy:
            BucketName: !Ref SpillBucket

athena-dynamodb/athena-dynamodb.yaml
ancamillo: I had to explicitly create a role for the lambda instead of using the existing Policies property in the Lambda because using Policies there is no way to define the Resource part of:

          - Action:
              - kms:GenerateDataKey
            Effect: Allow
            Resource: !Sub "arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${KMSKeyId}"

for both the cases where KMSKeyId is passed and when it is not, because when KMSKeyId is not passed the above statement fails and to avoid that instead, when KMSKeyId is not passed I just do not create the FunctionKMSPolicy at all using the condition "HasKMSKeyId". So to recap I had to modify the role\policy creation from the previous way that implicitly created them via just listing the permissions under the lambda Policies key to explicitly create them so that I can avoid to create the FunctionKMSPolicy if KMSKeyId is not passed and I would not be able to define the Resource filter for the kms:GenerateDataKey permission. Hope it makes sense but happy to explain it further if required.

avirtuos: Ack.

athena-dynamodb/athena-dynamodb.yaml
avirtuos: Ack.

athena-dynamodb/athena-dynamodb.yaml
avirtuos: looks good now

avirtuos:
@ancamillo Can you propose how to do this for all the connectors (or at least 1 more)? So the team can then apply it to all the other connectors after we merge your PR? At present, I'm still a bit fuzzy on why some of the other changes were made (e.g. to the DynamoDB parts of the policy). One more example might help. Cloudwatch logs might be a good one to help illustrate the pattern.

ancamillo:
> @ancamillo Can you propose how to do this for all the connectors (or at least 1 more)? So the team can then apply it to all the other connectors after we merge your PR? At present, I'm still a bit fuzzy on why some of the other changes were made (e.g. to the DynamoDB parts of the policy). One more example might help. Cloudwatch logs might be a good one to help illustrate the pattern.

Just added the CloudWatch logs example too following the same procedure:
- add KMSKeyId parameter;
- define condition HasKMSKeyId that is true if kms_key_id is passed otherwise false;
- create a kms_key_id env var for the lambda connector only if the condition is true;
- define the role of the lambda outside the lambda instead of inside it with same permissions;
- create a KMSPolicy policy only if HasKMSKeyId is true which enables the lambda to use the key\encrypted S3 bucket.

avirtuos (approved):
looks good to me. Only thing left before we merge this is release testing.

avirtuos:
@burhan94 or @shurvitz can one of you deploy and do a quick test of this? Should only take a few minutes and then we can merge it.

shurvitz:
@burhan94 @avirtuos, will take a look.

shurvitz:
@ancamillo @avirtuos , Looks like KMSKeyId is **not** optionalParameters: [KMSKeyId] must have values

shurvitz (commented):
`KMSKeyId:` definition is missing:
```
    Default: ""
```
needed to make optional.

shurvitz (changes requested):
Please make the KMS key ID optional.

ancamillo:
> Please make the KMS key ID optional.

Done. I was deploying this from the console and it was working without any value for  KMS Key ID because I realize the console was passing anyway an empty string. With the Default: "" as per your suggestion the deploy should work even without any value for KMS Key Id.

shurvitz (approved):
Looks good, thanks.

shurvitz:
@avirtuos ready for you to merge.

burhan94 (approved):

PR #251, opened 2020-09-23
Catching all Athena exceptions in QueryStatusChecker

atennak1:
*Issue #, if available:* N/A

*Description of changes:* See some query failures due to `Rate exceeded (Service: AmazonAthena; Status Code: 400; Error Code: ThrottlingException)`, so catching all Exceptions instead of just RuntimeException, since I can't think of a reason why we would want to throw any Athena exceptions all the way up.


burhan94:
Theres a build failure due to unused import

shurvitz (approved):
Approved.

PR #280, opened 2020-10-23
Add example for the Jdbc connection string

sbourgin:
*Description of changes:*
Update readme file of athena-jdbc and add an example for the Jdbc connection string

### DIFF
--- a/athena-jdbc/README.md
+++ b/athena-jdbc/README.md
@@ -54,6 +54,8 @@ Multiplexer provides a way to connect to multiple database instances of any type
 
 ```
 ${catalog}_connection_string    Database instance connection string. One of two types specified above. Required.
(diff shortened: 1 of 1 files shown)


athena-jdbc/README.md
```diff
@@ -54,6 +54,8 @@ Multiplexer provides a way to connect to multiple database instances of any type
 
 ```
 ${catalog}_connection_string    Database instance connection string. One of two types specified above. Required.
+                                If the catalog is myredshiftcatalog then the environment variable name should be myredshiftcatalog_connection_string
```
atennak1, line 57: adjust to "If the catalog *as registered with Athena* is myredshiftcatalog..."

atennak1 (changes requested):

atennak1 (approved):

athena-jdbc/README.md
sbourgin: Done

PR #284, opened 2020-11-03
env variable log level for athena-federation-sdk

ahuwaa:
*Description of changes:*
This is a suggestion on providing more control over logging. In its current state, the athena-federation-sdk uses log level `INFO` to log full requests e.g.
https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java#L232
https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/RecordHandler.java#L154
This includes potentially sensitive information if stored within Split metadata or records written back to Athena after `READ_RECORDS`. It can also be incredibly verbose if there is a lot of data being processed. While this defaults to `INFO`, an env variable can be specified during runtime to limit logging to `warn` or up to not expose the sensitive info e.g. `ATHENA_FEDERATION_SDK_LOG_LEVEL=warn`

For custom connectors, the log4j2 conf can be modified appropriately but for existing connectors deployed from SAM, this provides a bit more control.

*Testing:*
* Compared results from `AWS_REGION=us-west-2 mvn test` and `AWS_REGION=us-west-2 ATHENA_FEDERATION_SDK_LOG_LEVEL=ERROR mvn test` within `athena-aws-cmdb`. 

### DIFF
--- a/athena-aws-cmdb/src/main/resources/log4j2.xml
+++ b/athena-aws-cmdb/src/main/resources/log4j2.xml
@@ -28,7 +28,7 @@
         </Lambda>
     </Appenders>
     <Loggers>
-        <Logger name="com.amazonaws.athena.connector.lambda" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-info}" />
+        <Logger name="com.amazonaws.athena.connector.lambda" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-warn}" />
         <Root level="info">
             <AppenderRef ref="Lambda" />
         </Root>
--- a/athena-aws-cmdb/src/main/resources/log4j2.xml
+++ b/athena-aws-cmdb/src/main/resources/log4j2.xml
@@ -28,6 +28,7 @@
         </Lambda>
     </Appenders>
     <Loggers>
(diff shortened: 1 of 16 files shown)


atennak1:
This is great! Can you update the description with what testing you did?

avirtuos (approved):
looks good.

avirtuos:
@burhan94 can you test and merge this? I think we want this rolled out with a default of 'WARN' log level so you might need to tweak the PR before merging. I also think you want to expedite this to happen before the upcoming GA.

ahuwaa:
Testing info added

ahuwaa:
Note also this env control only applies to the `athena-federation-sdk` package. Assuming this code is also what is used for SAM deployments, I was thinking if this approach moves forward then it may also make sense to add another env variable for the custom connector log level itself.

burhan94:
sent updated commit with default level as warn. Confirmed in testing that only WARN log statements show up from the federation SDK.

atennak1 (approved):

PR #285, opened 2020-11-05
Integration-Test Suite

shurvitz:
**Description of changes:** Integration-Test Suite using CDK to create CloudFormation stack with all resources. The Suite manages the infra (creates/deletes the CF stack) and allows each connector to customize their integration-tests.


athena-federation-sdk-tools/pom.xml
```diff
@@ -70,6 +70,24 @@
             <version>${log4j2Version}</version>
             <scope>runtime</scope>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-cloudformation -->
```
avirtuos, line 73: This need to be test dependencies so they don't go into the jar that is fed to Lambda

shurvitz: changed CloudFormation dependecy scope to provided `<scope>provided</scope>` in sdk-tools pom. Dependency will need to be provided in the specific connectors' pom with a `<scope>test</scope>`

athena-federation-sdk/pom.xml
```diff
@@ -22,7 +22,6 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <slf4jVersion>1.7.30</slf4jVersion>
         <log4j2Version>2.13.3</log4j2Version>
-        <aws-sdk.version>1.11.769</aws-sdk.version>
```
burhan94, line 25: why was this removed?

shurvitz: The variable was duplicated from the main app pom causing a conflict situation where the Federation SDK could have a different AWS SDK version than the Connectors.

burhan94 (approved):

PR #316, opened 2020-12-07
Fixes LambdaFunctionException: Unable to create domain map - Elasticsearch connector

shurvitz:
**Issue #, if available:**
When using Amazon Elasticsearch service as a data-source provider, if the account being used has more than 5 Domain Names defined in Amazon Elasticsearch service, the following error will be observed in Athena for `ListDatabases`/`show databases` requests:
```
java.lang.RuntimeException: com.amazonaws.services.lambda.invoke.LambdaFunctionException: Unable to create domain map: Please provide a maximum of 5 Elasticsearch domain names to describe. (Service: AWSElasticsearch; Status Code: 400;...
```
**Description of changes:**
Since the `DescribeElasticsearchDomains` Elasticsearch API can only process 5 domain names in a single call, the lambda function was modified to generate multiple calls to that API when the list of domain names exceeds 5.

### DIFF
--- a/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchDomainMapProvider.java
+++ b/athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchDomainMapProvider.java
@@ -87,7 +87,7 @@ class ElasticsearchDomainMapProvider
     private Map<String, String> getDomainMapFromAmazonElasticsearch()
             throws RuntimeException
     {
-        final AWSElasticsearch awsEsClient = AWSElasticsearchClientBuilder.defaultClient();
+        final AWSElasticsearch awsEsClient = getElasticsearchClient();
         final Map<String, String> domainMap = new HashMap<>();
 
         try {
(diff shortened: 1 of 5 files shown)


burhan94:
Can you add some unit tests to this?

shurvitz:
There's already a UT that checks the logic when `autoDiscoverEndpoint=false`. I don't see a way to include a test for `autoDiscoverEndpoint=true` due to multiple dependencies. One being the dependency on the flag itself, and the second is the dependency on an the Elasticsearch client that is declared locally.

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchDomainMapProvider.java
```diff
@@ -127,6 +127,15 @@ protected ElasticsearchDomainMapProvider(boolean autoDiscoverEndpoint)
...
     }
 
+    /**
+     * Gets an Amazon Elasticsearch client.
+     * @return Amazon Elasticsearch client.
+     */
+    protected AWSElasticsearch getElasticsearchClient()
+    {
+        return AWSElasticsearchClientBuilder.defaultClient();
+    }
```
avirtuos, line 137: Can we inject a factory instead? This way the test can inject a mock instead of using a spy / partial mock?

avirtuos (changes requested):

athena-elasticsearch/src/main/java/com/amazonaws/connectors/athena/elasticsearch/ElasticsearchDomainMapProvider.java
shurvitz: Done.

avirtuos (approved):

PR #320, opened 2020-12-16
Allow PostGreSqlCompositeHandler to be used as Lambda handler class

Meosit:
Removed unused constructor parameters to allow to use this class as AWS Lambda handler directly.


avirtuos (approved):
Thanks for the contribution, seems this was overlooked when the connector was originally written.

PR #328, opened 2020-12-23
[Bug-Fix]: Duplicate splits generated when splits are continued

verdantforce:
*Description of changes:*
This MR addresses duplicate splits bug in the current implementation for athena-jdbc connectors.  I believe that the issue here is that when the number of splits is too large, the continuation token encoded should be next **unprocessed** partition.  Otherwise, the each partition at the partition boundary (MAX_SPLIT_PER_REQUEST) gets included twice in the current implementation.  One time when the MAX_SPLIT_PER_REQUEST is reached, and another time when the next request resumes from the partition that was just processed.


avirtuos (approved):

avirtuos:
Thanks for the fix, sorry we missed this initially.

verdantforce:
No problem.  I just want to take this opportunity to say it is amazing what this sdk unlocks for our company's custom use cases.  Keep up the good work.  👍

avirtuos:
Always happy to hear about the great things our customers are doing with Athena. Happy Holidays!

PR #331, opened 2021-01-05
Integ-Test module refactor for VPC config + Redshift integ-tests 

shurvitz:
**Description of changes:**

Refactored `athena-federation-integ-test` module to support VPC configuration, and added integration tests for the Redshift connector.


athena-dynamodb/README.md
```diff
@@ -51,14 +51,19 @@ Review the "Policies" section of the athena-dynamodb.yaml file for full details
...
 
 To use this connector in your queries, navigate to AWS Serverless Application Repository and deploy a pre-built version of this connector. Alternatively, you can build and deploy this connector from
 source follow the below steps or use the more detailed tutorial in the athena-example module:
 
-1. From the athena-federation-sdk dir, run `mvn clean install` if you haven't already.
-2. From the athena-dynamodb dir, run `mvn clean install`.
-3. From the athena-dynamodb dir, run  `../tools/publish.sh S3_BUCKET_NAME athena-dynamodb` to publish the connector to your private AWS Serverless Application Repository. The S3_BUCKET in the command
+1. From the **athena-federation-sdk** dir, run `mvn clean install` if you haven't already.
+2. From the **athena-federation-integ-test** dir, run `mvn clean install` if you haven't already.
+3. From the **athena-dynamodb** dir, run `mvn clean install`.
```
avirtuos, line 65: Will this run integration tests? hopefully not since mvn clean install should be able to succeed in a sandbox (e.g. no remote resources)

athena-federation-integ-test/README.md
```diff
@@ -88,7 +88,35 @@ Provide implementation for the following 4 abstract methods in the test class:
      * access to multiple connector-specific AWS services (e.g. DynamoDB, Elasticsearch etc...)
      * @return A policy document object.
      */
-    protected abstract PolicyDocument getConnectorAccessPolicy();
+    protected abstract Optional<PolicyDocument> getConnectorAccessPolicy();
```
avirtuos, line 91: public API changes should not be taken lightly. Why does this need to change?

athena-federation-integ-test/README.md
```diff
@@ -88,7 +88,35 @@ Provide implementation for the following 4 abstract methods in the test class:
...
+
+To use a VPC configuration, the following method must be overridden in your integration test class to `return true`
+(default is false):
+
+```java
+    /**
+     * Must be overridden in the extending class when the connector supports a VPC configuration.
+     * @return false (default)
+     */
+    protected boolean connectorSupportsVpcConfig()
```
avirtuos, line 109: What is an example of a connector that doesn't support VPC config?

athena-federation-integ-test/README.md
```diff
@@ -88,7 +88,35 @@ Provide implementation for the following 4 abstract methods in the test class:
...
+```
+Access to the VPC configuration (VPC Id, Security Group Id, Subnet Ids, and Availability Zones) can be obtained via the
+Stack object:
+
+```java
+    @Override
+    protected void setUpStackData(final Stack stack)
+    {
+        ConnectorVpcAttributes vpcAttributes = ((ConnectorWithVpcStack) stack).getConnectorVpcAttributes();
```
avirtuos, line 118: do you really need the boolean or can you just make getCOnnectorVpcAttributes() return an Optional?

athena-federation-integ-test/README.md
```diff
@@ -145,9 +173,13 @@ locally from the terminal.
...
 the first time, and each time the connector's code changes:
 
-1. Build the SDK and connector (SDK only needs to build once): `mvn clean install`
-2. Export the IAM credentials for the AWS account used for testing purposes.
-3. Package the connector (from the connector's directory):
+1. From the **athena-federation-sdk** dir, run `mvn clean install` if you haven't done so already.
+2. From the **athena-federation-integ-test** dir, run `mvn clean install` if you haven't done so already.
+3. From your connector's dir, run `mvn clean install`.
+4. Export the IAM credentials for the AWS account used for testing purposes. The account's Athena console should have
+   `FederationIntegrationTests` Workgroup defined and set to `Athena engine version 2` (for more information on the
```
avirtuos, line 180: Is the WorkGroup name configurable? If not, it should be.

athena-federation-integ-test/pom.xml
```diff
@@ -82,6 +82,12 @@
             <artifactId>aws-java-sdk-cloudformation</artifactId>
             <version>${aws-sdk.version}</version>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-ec2 -->
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-java-sdk-ec2</artifactId>
+            <version>${aws-sdk.version}</version>
+        </dependency>
```
avirtuos, line 90: Is this to support the new VPC info? and can you remind me if the appropriately setup the integ-test pom and dependencie such that they don't end up in the Lambda jar?

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/CloudFormationClient.java
```diff
@@ -77,8 +83,7 @@ private void processCreateStackRequest(CreateStackRequest createStackRequest)
             throws RuntimeException
     {
         // Create CloudFormation stack.
-        AmazonCloudFormation cloudFormationClient = AmazonCloudFormationClientBuilder.defaultClient();
-        CreateStackResult result = cloudFormationClient.createStack(createStackRequest);
+        CreateStackResult result = client.createStack(createStackRequest);
```
avirtuos, line 86: why did we go from a descriptive name "cloudFormationClient' to a generic name 'client'?

athena-dynamodb/README.md
shurvitz: No, only UT. To run integration tests you need to explicitly run: `mvn failsafe:integration-test`.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/CloudFormationTemplateProvider.java
```diff
@@ -156,6 +80,12 @@ private Map getEnvironmentVars()
      */
     protected abstract void setSpecificResource(final Stack stack);
 
+    /**
+     * Must be overridden to indicate whether the connector being tested supports a VPC configuration.
+     * @return true/false
+     */
+    protected abstract boolean isSupportedVpcConfig();
+
```
avirtuos, line 88: i'm not thrilled with boolean APIs like this, can we instead have a getVPCConfig() that returns Optional? and we can use it for both? this way it avoid confusion where someone implements one correctly but not the other

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/IntegrationTestBase.java
```diff
@@ -51,11 +52,10 @@
     private static final String ATHENA_QUERY_RUNNING_STATE = "RUNNING";
     private static final String ATHENA_QUERY_FAILED_STATE = "FAILED";
     private static final String ATHENA_QUERY_CANCELLED_STATE = "CANCELLED";
-    private static final String ATHENA_FEDERATION_WORK_GROUP = "AmazonAthenaPreviewFunctionality";
+    private static final String ATHENA_FEDERATION_WORK_GROUP = "FederationIntegrationTests";
```
avirtuos, line 55: can we make this configurable with a good default like this one?

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/providers/ConnectorVpcAttributesProvider.java
```diff
@@ -0,0 +1,133 @@
...
+     * @return Security Group Id
+     * @throws RuntimeException No Security Group Ids were found.
+     */
+    private String getSecurityGroupId(final String vpcId)
+            throws RuntimeException
+    {
+        DescribeSecurityGroupsResult sgResult = ec2Client.describeSecurityGroups();
+        for (SecurityGroup securityGroup : sgResult.getSecurityGroups()) {
+            if (securityGroup.getVpcId().equals(vpcId)) {
+                return securityGroup.getGroupId();
```
avirtuos, line 99: how do you know this is the right security group?

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/providers/ConnectorVpcAttributesProvider.java
```diff
@@ -0,0 +1,133 @@
...
+     * Gets the default VPC Id used for configuring the Lambda function (e.g. vpc-xxxx).
+     * @return VPC Id
+     * @throws RuntimeException No VPCs were found.
+     */
+    private String getDefaultVpcId()
+            throws RuntimeException
+    {
+        DescribeVpcsResult vpcsResult = ec2Client.describeVpcs();
+        for (Vpc vpc : vpcsResult.getVpcs()) {
+            if (vpc.getIsDefault()) {
```
avirtuos, line 80: why the default VPC?

avirtuos (changes requested):
Per my chat with @shurvitz he is going to rework this a bit. I dont think we can auto-detect the VPC, SGs, and subnets like this in the vast majority of cases. Instead, this likely need to be explicit configs to the test suite otherwise customers will have a tough time running the tests anywhere but an isolated account. An isolated account does have its advantages but its unrealistic to think such strict expectations of 1 VPC and 1 SG, etc... will hold in practice. I also think there are many other configs that need to be supported but aren't yet known.

athena-federation-integ-test/README.md
shurvitz: When I first worked on the integ-test fwk, I based it on the DynamoDb connector which has a dynamodb-specific policy. Later, when working on the Redshift integ-tests, I realized that the connector-specific policy is optional (e.g. Redshift doesn't need one). Rather than returning null and then checking for nulls in the framework, I figured that using Optional<> would be preferable here.

athena-federation-integ-test/README.md
shurvitz: DynamoDB

athena-federation-integ-test/README.md
shurvitz: Currently the fwk uses the boolean in the stack factory to determine what kind of stack to generate (with/without VPC). This will need to be reworked when switching to a config file. The values from the config file, thus, can drive the creation of the stack rather than a boolean return value.

athena-federation-integ-test/README.md
shurvitz: Can be added as a parameter to the config file with a default value.

athena-federation-integ-test/pom.xml
shurvitz: this will be removed when I switch the VPC logic in favor of the config file. Moot point, but these dependencies are not in the Lambda jar since we are including the integ-test module as a test dependency.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/CloudFormationClient.java
shurvitz: I'll change the name.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/CloudFormationTemplateProvider.java
shurvitz: This will be removed in favor of a config file.

shurvitz: I will include a public API to get the VPC config that returns an Optional.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/IntegrationTestBase.java
shurvitz: will include as parameter in config file with the value above as default.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/providers/ConnectorVpcAttributesProvider.java
shurvitz: will change logic in favor of config file.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/providers/ConnectorVpcAttributesProvider.java
shurvitz: will change logic in favor of config file.

athena-dynamodb/README.md
avirtuos: so why do these instructions say to build the integ test package? should we caveat this to say (only if you also plan to run mvn failsafe:integration-test ?

athena-example/README.md
```diff
@@ -279,3 +279,8 @@ describe `lambda:<function_name>`.schema1.table1;
 *note that the <function_name> corresponds to the name of your Lambda function.*
 
 Don't forget to try out [Amazon Athena ML](https://github.com/awslabs/aws-athena-query-federation/wiki/AthenaML-Tutorial) to apply SageMaker model inference right from your Amazon Athena queries.
+
+## Integration Tests
+
+Although the topic of integration tests is not covered in this example, you can learn more about how to create and
+run integration tests for your connector in the [Integration-Test Module](https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-integ-test/README.md).
```
avirtuos, line 286: Is it possible to add an example here? i think this connector reads from S3 so maybe we can write an example for it?

avirtuos: its ok to do this as a separate PR but i think it is important.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/CloudFormationClient.java
```diff
@@ -140,9 +145,9 @@ protected void deleteStack()
         logger.info("------------------------------------------------------");
 
         try {
-            AmazonCloudFormation cloudFormationClient = AmazonCloudFormationClientBuilder.defaultClient();
             DeleteStackRequest request = new DeleteStackRequest().withStackName(stackName);
             cloudFormationClient.deleteStack(request);
+            cloudFormationClient.shutdown();
```
avirtuos, line 150: shouldnt this be done in a finally block?

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/ConnectorStackAttributesProvider.java
```diff
@@ -0,0 +1,109 @@
...
+     * Sets defaults for environment variables (spill_bucket, spill_prefix, disable_spill_encryption) if not provided
+     * by connector.
+     * @return A Map containing the environment variables key-value pairs.
+     * @throws RuntimeException The spill_bucket is neither specified in the environment vars nor the test config file.
+     */
+    private void setUpEnvironmentVars()
+            throws RuntimeException
+    {
+        // Check for missing spill_bucket
+        if (!environmentVariables.containsKey(LAMBDA_SPILL_BUCKET_TAG)) {
```
avirtuos, line 78: remind me why we are using environment variables for anything instead of only using the config file? seems confusing to have a mix.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/ConnectorStackAttributesProvider.java
```diff
@@ -0,0 +1,109 @@
...
+        if (!environmentVariables.containsKey(LAMBDA_SPILL_PREFIX_TAG)) {
+            // Add missing spill_prefix environment variable
+            environmentVariables.put(LAMBDA_SPILL_PREFIX_TAG, "athena-spill");
+        }
+
+        // Check for missing disable_spill_encryption environment variable
+        if (!environmentVariables.containsKey(LAMBDA_DISABLE_SPILL_ENCRYPTION_TAG)) {
+            // Add missing disable_spill_encryption environment variable
+            environmentVariables.put(LAMBDA_DISABLE_SPILL_ENCRYPTION_TAG, "false");
+        }
```
avirtuos, line 98: would it be better to throw here than have defaults? we provide defaults in SAR only for convenience in the UI... i'm not sure thats appropriate here since we have a robust config file infra you've added.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/data/ConnectorPackagingAttributes.java
```diff
@@ -0,0 +1,64 @@
...
+package com.amazonaws.athena.connector.integ.data;
+
+/**
+ * Contains the the connector's packaging attributes.
+ */
+public class ConnectorPackagingAttributes
+{
+    private final String s3Bucket;
+    private final String s3Key;
+    private final String lambdaFunctionHandler;
```
avirtuos, line 29: comments to describe what these are used for? a javadoc that is basically the name of the class isn't super useful for future maintainers.

athena-jdbc/README.md
```diff
@@ -178,13 +178,19 @@ A partition is represented by two partition columns of type varchar. We leverage
 
 **Note:** In case of Redshift partition_schema and partition_name will always be '*'. It does not support external partitions. Performance with huge datasets is slow.
 
+### Running Integration Tests
+
+See [Running Integration Tests](https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-integ-test/README.md#running-integration-tests).
+
```
avirtuos, line 184: in addition to the link, can you maybe put the commands here? i thought its just the build of the integ test and then the run of the tests.

athena-jdbc/README.md
```diff
@@ -178,13 +178,19 @@ A partition is represented by two partition columns of type varchar. We leverage
...
+To use this connector in your queries, navigate to AWS Serverless Application Repository and deploy a pre-built version of this connector. Alternatively, you can build and deploy this connector from
+source follow the below steps or use the more detailed tutorial in the athena-example module:
 
-1. From the athena-federation-sdk dir, run `mvn clean install` if you haven't already.
-2. From the athena-jdbc dir, run `mvn clean install`.
-3. From the athena-jdbc dir, run  `../tools/publish.sh S3_BUCKET_NAME athena-jdbc` to publish the connector to your private AWS Serverless Application Repository. The S3_BUCKET in the command is where a copy of the connector's code will be stored for Serverless Application Repository to retrieve it. This will allow users with permission to do so, the ability to deploy instances of the connector via 1-Click form. Then navigate to [Serverless Application Repository](https://aws.amazon.com/serverless/serverlessrepo)
+1. From the **athena-federation-sdk** dir, run `mvn clean install` if you haven't already.
+2. From the **athena-federation-integ-test** dir, run `mvn clean install` if you haven't already.
```
avirtuos, line 191: will mvn clean install fail if i skip this?

athena-jdbc/README.md
avirtuos: Also, what settings do i set that are connector specific in that test config?

athena-jdbc/src/test/java/com/amazonaws/connectors/athena/jdbc/integ/RedshiftIntegTest.java
```diff
@@ -0,0 +1,281 @@
...
+public class RedshiftIntegTest extends IntegrationTestBase
+{
+    private static final Logger logger = LoggerFactory.getLogger(RedshiftIntegTest.class);
+
+    private static final String REDSHIFT_DB_NAME = "public";
+    private static final String REDSHIFT_DB_PORT = "5439";
+    private static final String REDSHIFT_DB_USERNAME = "integusername";
+    private static final String REDSHIFT_DB_PASSWORD = "IntegPassword1";
+    private static final String REDSHIFT_TABLE_MOVIES = "movies";
+    private static final String REDSHIFT_TABLE_BDAY = "bday";
```
avirtuos, line 64: these seem like good candidates for the config file... username and password would ideally come from a SecretsManager secret. not be in this code or in a config file.

athena-jdbc/src/test/java/com/amazonaws/connectors/athena/jdbc/integ/RedshiftIntegTest.java
```diff
@@ -0,0 +1,281 @@
...
+    private final String lambdaFunctionName;
+    private final String clusterName;
+    private final String clusterEndpoint;
+    private final Map<String, String> environmentVars;
+
+    public RedshiftIntegTest()
+    {
+        lambdaFunctionName = getLambdaFunctionName();
+        clusterName = "redshift-" + UUID.randomUUID();
+        clusterEndpoint = clusterName + ".c9afbdb0synx.us-east-1.redshift.amazonaws.com";
```
avirtuos, line 77: why are we hard coding any portion of the hostname/endpoint? are we saying no one should use this outside of us-east-1?

athena-jdbc/test-config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "athena_work_group" : "FederationIntegrationTests",
+  "spill_bucket" : "",
+  "vpc_id" : "",
+  "security_group_id" : "",
+  "subnet_ids" : [],
+  "availability_zones" : []
```
avirtuos, line 7: should add the username, password and an option SecretsManager secret ARN here instead of in code.

avirtuos (changes requested):
Good iteration, getting really close. I posted some comments.

athena-dynamodb/README.md
shurvitz: Failure to compile the module first will result in a compilation error (added note).

athena-dynamodb/test-config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "athena_work_group" : "FederationIntegrationTests",
+  "spill_bucket" : ""
```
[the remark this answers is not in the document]
shurvitz: Modified and moved to etc sub-folder.

athena-example/README.md
shurvitz: Let's target this for a future enhancement.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/CloudFormationClient.java
shurvitz: Moved to finally block.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/ConnectorStackAttributesProvider.java
shurvitz: moved to config file.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/ConnectorStackAttributesProvider.java
shurvitz: moved to config file.

athena-federation-integ-test/src/main/java/com/amazonaws/athena/connector/integ/data/ConnectorPackagingAttributes.java
shurvitz: Added detailed description.

athena-jdbc/README.md
shurvitz: Added additional info.

athena-jdbc/README.md
shurvitz: If you run `mvn clean install` from the root directory of the repo, this will compile okay. But, if you're building individual modules, the integ-test module needs to be built first. Failure to do so will result in compilation errors. Added note to that effect.

athena-jdbc/src/test/java/com/amazonaws/connectors/athena/jdbc/integ/RedshiftIntegTest.java
shurvitz: In this instance, we are creating a strictly temporary DB instance/cluster whose persistence is only for the duration of the test, so credentials obfuscation via SecretsManager would be overkill. But, I do agree that we may have other instances in the future (e.g. Vertica connector maybe) where the DB instance might be persistent. So, let's target SecretsManager for a separate PR.

Moved params to config file.

athena-jdbc/src/test/java/com/amazonaws/connectors/athena/jdbc/integ/RedshiftIntegTest.java
shurvitz: Redshift doesn't give you the ability to customize the endpoint. Furthermore, they don't provide a good API for extracting the endpoint after-the-fact (if they did, we would still have to break the CloudFormation process into 2 separate stacks since the Lambda also needs the connection_string containing the endpoint). I did, however, move this to the config file so it would be easier to change by the user if need be.

athena-jdbc/test-config.json
shurvitz: Added params to config file (will tackle SecretsManager in a separate PR).

avirtuos (changes requested):
Where are the SecretsManager changes we discussed? I'm not comfortable with that being a follow up PR as this version requires passwords to existing in plain text in config files. The SecretsManager changes should take less than 1hr, possibly only a few minutes given we already have plenty of examples of using SecretsManager in this repository. 

Once you make that change, I think this can be merged.

athena-jdbc/README.md
```diff
@@ -178,13 +178,23 @@ A partition is represented by two partition columns of type varchar. We leverage
...
+1. From the **athena-federation-sdk** dir, run `mvn clean install` if you haven't already.
+2. From the **athena-federation-integ-test** dir, run `mvn clean install` if you haven't already
+   (**Note: failure to follow this step will result in compilation errors**).
+3. From the **athena-jdbc** dir, run `mvn clean install`.
+4. From the **athena-jdbc** dir, run  `../tools/publish.sh S3_BUCKET_NAME athena-jdbc` to publish the connector to your private AWS Serverless Application Repository. The S3_BUCKET in the command is where a copy of the connector's code will be stored for Serverless Application Repository to retrieve it. This will allow users with permission to do so, the ability to deploy instances of the connector via 1-Click form. Then navigate to [Serverless Application Repository](https://aws.amazon.com/serverless/serverlessrepo)
```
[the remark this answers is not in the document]
shurvitz: Due to differences in specific implementations, there are going to be different test classes for each DB type. So Redshift is RedshiftIntegTest, Postgres will probably be PostgresIntegTest, etc... I think that is the cleaner approach.

athena-dynamodb/README.md
```diff
@@ -51,14 +51,26 @@ Review the "Policies" section of the athena-dynamodb.yaml file for full details
 4. CloudWatch Logs - This is a somewhat implicit permission when deploying a Lambda function but it needs access to cloudwatch logs for storing logs.
 1. Athena GetQueryExecution - The connector uses this access to fast-fail when the upstream Athena query has terminated.
 
+### Running Integration Tests
+
+Integration tests can be executed by specifying the following command: `mvn failsafe:integration-test` from the
```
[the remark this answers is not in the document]
shurvitz: Fixed section.

athena-dynamodb/README.md
```diff
@@ -51,14 +51,23 @@ Review the "Policies" section of the athena-dynamodb.yaml file for full details
 4. CloudWatch Logs - This is a somewhat implicit permission when deploying a Lambda function but it needs access to cloudwatch logs for storing logs.
 1. Athena GetQueryExecution - The connector uses this access to fast-fail when the upstream Athena query has terminated.
 
+### Running Integration Tests
+
+Integration tests can be executed by specifying the following command: `mvn failsafe:integration-test`
+(see [Running Integration Tests](https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-integ-test/README.md#running-integration-tests)).
+In order for the tests to run successfully, however, some test configuration attributes must be provided in the `etc/test-config.json`
+JSON file (for additional information see the README in the [Integration-Test Module](https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-integ-test/README.md)).
+
```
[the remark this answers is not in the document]
shurvitz: Fixed section.

athena-federation-integ-test/README.md
```diff
@@ -20,12 +20,12 @@ connector.
...
 This section explains the steps necessary to create integration tests using the
-Integration-Test module. For an actual code example, see the DynamoDB connector
+Integration-Test framework. For an actual code example, see the DynamoDB connector
 [DynamoDbIntegTest integration-test class](https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DynamoDbIntegTest.java).
 
 ### Dependencies
 
-Add the Integration-Test module as a test dependency in the specific connector's `pom.xml` file (replace `2020.46.1`
+Add the Integration-Test framework as a test dependency in the specific connector's `pom.xml` file (replace `2020.46.1`
 with current version):
```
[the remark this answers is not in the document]
shurvitz: Fixed.

athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DynamoDbIntegTest.java
```diff
@@ -43,8 +44,10 @@
  */
 public class DynamoDbIntegTest extends IntegrationTestBase {
     private static final Logger logger = LoggerFactory.getLogger(DynamoDbIntegTest.class);
-    private static final String DATABASE_NAME = "default";
 
+    Map<String, String> userSettings = getUserSettings().orElseThrow(() ->
+            new RuntimeException("user_settings attribute must be provided in test-config.json."));
+    private final String dynamodbDbName = userSettings.get("dynamodb_db_name");
```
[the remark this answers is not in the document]
shurvitz: moved to constructor.

PR #351, opened 2021-02-11
Add ssl support for Redis Connector

burhan94:
Currently the redis connector does not support encrypted redis clusters. This change will allow users to add a rediss:// prefix to their hostname to allow ssl access.


burhan94:
Closing since customer also requires multi-cluster support. Need to look into other support options, and will include SSL support with multi-cluster support.

TejjeT:
@burhan94 any ETA on this ? Just want to make sure I am not duplicating the effort on writing something similar.

burhan94:
Hey, I ended up closing the PR since I'm not sure if Jedis is the right tool for this. I'm weighing my options for using Jedis vs Lettuce for our redis client, so adding SSL support this way could make it difficult if we decide to switch to Lettuce in terms of backwards compatibility. Unfortunately we had to re-prioritize, so it might be some time before we can get back to this since we're already spread pretty thin.

mannharleen:
Why has the issues been closed?

Issue #356, opened 2021-02-18
[BUG] Query that returns zero rows produces an exception with connector Athena-Vertica 


asimjalis:
**Describe the bug**
Athena-Vertica connector throws exception if query returns zero rows.

**To Reproduce**
1. Go to Athena console.
2. Run query on Vertica using catalog connected with Athena-Vertica connector that returns 0 rows.
3. Athena console displays error message.

**Expected behavior**
No error should be thrown.

**Screenshots / Exceptions / Errors**

Here is a query that I ran.

```
SELECT * FROM "vertica_feb_1"."store".store_dimension WHERE store_state='XX'
```

Here is the error that I got.

```
Your query has the following error(s):

GENERIC_USER_ERROR: Encountered an exception[java.lang.RuntimeException] from your LambdaFunction[arn:aws:lambda:us-west-2:861469877924:function:vertica_feb_1] executed in context[retrieving meta-data] with message[Export to S3 from Vertica was not successful]

This query ran against the "store" database, unless qualified by the query. Please post the error message on our forum or contact customer support with Query Id: 27fa69fa-dd74-4add-ba52-6fddb98c31b4.
```

**Connector Details (please complete the following information):**
 - Version: 2021.35.10 
 - Name: vertica
 - Athena Query IDs: N/A

**Additional context**
The error only occurs if the query references table columns. If the query predicate is `1=0` then no error occurs. For example, this query does not produce an error.

```
SELECT * FROM "vertica_feb_1"."store".store_dimension WHERE 1=0
```

avirtuos:
@rohit-masur This looks like a real bug to me. Can you update the connector to generate '0' splits in this situation? I think that is the appropriate fix.

rohit-masur:
@avirtuos : Yes, will take this up

PR #359, opened 2021-02-21
[Issue-356]-fix-zero-records-error

rohit-masur:
Issue #356 

Description:
Athena-Vertica connector throws exception if query returns zero rows.

In case of zero records for the issued query, Vertica will not export any data to S3. Added the logic to create an "empty" split to handle this situation. 

In addition, added the logic to check if the logged-in user has access and privileges to the table being queried.


burhan94:
Hi Rohit, 

Did you get a chance to test this locally?

rohit-masur:
Hey @burhan94, yes I tested this locally on a Vertica cluster.  Attached some screenshots for your reference. 
May be @asimjalis can test this cluster too to confirm.

Screen Shot 2021-03-05 at 8 37 40 AM
Screen Shot 2021-03-05 at 8 37 17 AM

avirtuos (approved):

Issue #373, opened 2021-03-23
[BUG] Query with range in sort keys and identifying partition key with connector dynamoDbConnector


laurihak:
**Describe the bug**
I have played around to query against index and it seems to be working fine. Range and another where clause to key expression attribute does not work.

Dynamodb table:
partition key:  id
sort key: datetime

So i Query the table as : 
SELECT *
FROM <aws.lambda.dynamo.connector.table>
WHERE partitionkey = 'id' AND sortKey BETWEEN 'dateFrom' AND 'dateTo'

Having the error: 

When deleting the the primary key from query it works fine.

WHERE sortKey BETWEEN 'dateFrom' AND 'dateTo'

**To Reproduce**
1. Have table with primary and sortKey
2. Query the table with primary key = 'yourPrimaryKey' and sortKey > 'from' AND  sortKey < 'to'.
Optional syntax: sortKey BETWEEN 'from' AND 'to'
3. Send the query from Athena

**Expected behavior**
Expected to query range with sort key.

I think the problem might be related to that dynamodb key-condition-expression should use BETWEEN if the original query includes the same key twice.

**Screenshots / Exceptions / Errors**
Your query has the following error(s):

GENERIC_USER_ERROR: Encountered an exception[com.amazonaws.services.dynamodbv2.model.AmazonDynamoDBException] from your LambdaFunction[arn:aws:lambda:eu-west-1:#######:function:dynamolambda] executed in context[S3SpillLocation{bucket='########', key='athena-spill/########f', directory=true}] with message[KeyConditionExpressions must only contain one condition per key (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ValidationException; Request ID: JL285EOAIDF2V6N92FRA1TRS0NVV4KQNSO5AEMVJF66Q9ASUAAJG; Proxy: null)]

This query ran against the "default" database, unless qualified by the query. Please post the error message on our forum or contact customer support with Query Id: 9027cae2-72d6-4029-bb50-612161c71e7c.

**Connector Details (please complete the following information):**
 - Version: 2021.6.1
 - Name: DynamoDb, AthenaDynamoDBConnector
 - Athena Query IDs 9027cae2-72d6-4029-bb50-612161c71e7c

miettal:
I face same issue.

last year grafana Athena datasouce was released, and I tried to connect dynamodb and grafana via athena.
but this connector is not support for filtering range key by using BETWEEN.
[...]

PR #375, opened 2021-03-24
#373 DynamoDB range of key needs BETWEEN syntax

laurihak:
*Issue #, if available:*
#373

*Description of changes:*
generateSingleColumnFilter -> rangeConjucts 
Changed the way sort key range is written to dynamodb request.
DynamoDB needs BETWEEN clause to not get error 

I also had to to change the test to fit the correct syntax

AWS DOCS:

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LegacyConditionalParameters.KeyConditions.html

Screen Shot 2021-03-24 at 19 30 54


atennak1 (commented):
Thanks for the PR! But I don't think this fixes the case where the start and/or end of the range is exclusive, e.g. `1 <= x < 10`.  Since BETWEEN is always inclusive I think we'll have to use BETWEEN but then somehow remove 1 or 2 bookend records once they're returned from DDB.  That's going to be kind of ugly but I don't see another way.

laurihak:
Yes, i think dynamodb does not support exlusive sorting with range of key. Removing records after the request will be kind of a hassle, but I agree it is the only way, if it really needs to be supported.

avirtuos:
Hey @atennak1 if you have a sec, whats your take on the current state of this. Should we abandon or whats needed to move forward?

atennak1:
@avirtuos @laurihak thought about it again and I think this PR will work as-is.  The connector will return an extra record or two, but Athena should still enforce the predicates on the returned data, so query results should still be correct.  Would be good to validate this with some test queries on a ranged dataset.

henrymai:
@laurihak

Thank you for your contribution.
We are going to merge a safer version of this patch that only handles the fully inclusive range situation here: https://github.com/awslabs/aws-athena-query-federation/pull/787

PR #381, opened 2021-03-29
Elasticsearch Integration Tests

shurvitz:
**Description of changes:**
Added integration tests to the Elasticsearch connector.


athena-elasticsearch/src/test/java/com/amazonaws/connectors/athena/elasticsearch/integ/ElasticsearchIndexUtils.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private final AwsRestHighLevelClient client;
+    private final String index;
+
+    private Integer requestId;
+
+    public ElasticsearchIndexUtils(String domainName, String index) {
+        AwsRestHighLevelClientFactory clientFactory = new AwsRestHighLevelClientFactory(true);
+        ElasticsearchDomainMapProvider domainMapProvider = new ElasticsearchDomainMapProvider(true);
+        Optional<String> endpoint = Optional.ofNullable(domainMapProvider.getDomainMap(null).get(domainName));
```
jobrandw, line 50: Won't passing in null here cause an exception?  

```
     * @throws RuntimeException The domain map cannot be created due to an error with the AWS ES SDK or an invalid
     * domainMapping variable (empty, null, or contain invalid information that cannot be parsed successfully).
     */
    public Map<String, String> getDomainMap(String domainMapping)
            throws RuntimeException
```

shurvitz: No. This will only cause an exception when `autoDiscoverEndpoint` is **false** (see code below). In this case `autoDiscoverEndpoint` will always be **true**  because of the instantiation: `domainMapProvider` = new ElasticsearchDomainMapProvider(**true**);
```java
    protected Map<String, String> getDomainMap(String domainMapping)
            throws RuntimeException
    {
        if (autoDiscoverEndpoint) {
            // Get domain mapping via the AWS ES SDK (1.x).
            return getDomainMapFromAmazonElasticsearch();
        }
        else {
            // Get domain mapping from the domainMapping variable.
            return getDomainMapFromEnvironmentVar(domainMapping);
        }
    }
```

jobrandw (approved):

PR #391, opened 2021-04-12
ListTableMetadata pagination in the Federation SDK.

shurvitz:
**Description of changes:**
This change enhances the support of Athena's ListTableMetadata API for federated catalogs by providing pagination for the request in the Federation SDK. The enhancement provides an interface for connectors to paginate the request directly to the data-source for added efficiency (where applicable).

**Testing:**
1. CP not changed, but Lambda changed - performed several tests: Ran integration tests on updated Lambda (to test QE), Created local stack and tested the GetTables and ListTablesMetadata API observing the nextToken returned and using it in the next request.
2. CP changed, but Lambda not changed - Ran same tests (GetTables and ListTableMetadata APIs) on local stack that's been updated with Lambda that was not. Observed nextToken and was able to get next set of results using it.
3. CP changed and Lmabda changed - Same as above.


avirtuos:
I'm a bit confused by a few things in this PR. Can you grab 15-30 min on my calendar and include @burhan94 before this gets pushed. Thanks

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -240,10 +281,20 @@ protected final void doHandleRequest(BlockAllocator allocator,
...
+                        assertNotNull(response);
+                        objectMapper.writeValue(outputStream, response);
+                    }
+                }
+                else {
+                    try (ListTablesResponse response = doListTables(allocator, (ListTablesRequest) req)) {
+                        logger.info("doHandleRequest: response[{}]", response);
+                        assertNotNull(response);
+                        objectMapper.writeValue(outputStream, response);
+                    }
```
avirtuos, line 297: we shouldnt need to branch here.... its just one thing

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/paginators/Paginator.java
```diff
@@ -0,0 +1,92 @@
...
+    private final int pageSize;
+
+    /**
+     * Constructor for the paginator.
+     * @param items A collection to T items to be paginated.
+     * @param startName A name used as the start of the paginated results.
+     * @param pageSize The maximum number of T items that can be be returned in each paginated result.
+     * @param getNameFromItem A function that will convert a T item to String.
+     */
+    public Paginator(Collection<T> items, String startName, int pageSize, Function<T, String> getNameFromItem)
```
avirtuos, line 48: im not convinced we can or need to provide a generic paginator in the SDK if we allow the connectors to handle pagination in a way that aligns to their internal implementations.

avirtuos (changes requested):
Per our meeting I'd recommend rethinking this approach by adding pagination to the existing listTables(...) facility in a way that is backwards compatible with existing (none updated callers) and also for new callers that are communicating with old connectors (this is where the capabilities stuff comes in.

athena-aws-cmdb/athena-aws-cmdb.yaml
```diff
@@ -36,6 +36,12 @@ Parameters:
     Description: "WARNING: If set to 'true' encryption for spilled data is disabled."
     Default: 'false'
     Type: String
+  ListTablesPageSize:
+    Description: 'Page size used for the pagination of the ListTableMetadata API (min 1 - 50 max).'
+    Type: Number
+    MinValue: 1
+    MaxValue: 50
+    Default: 50
```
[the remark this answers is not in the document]
shurvitz: Removed

athena-aws-cmdb/athena-aws-cmdb.yaml
```diff
@@ -45,6 +51,7 @@ Resources:
           disable_spill_encryption: !Ref DisableSpillEncryption
           spill_bucket: !Ref SpillBucket
           spill_prefix: !Ref SpillPrefix
+          list_tables_page_size: !Ref ListTablesPageSize
```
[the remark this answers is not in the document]
shurvitz: Removed

athena-cloudwatch-metrics/README.md
```diff
@@ -14,6 +14,10 @@ The Athena Cloudwatch Metrics Connector exposes several configuration options vi
...
 3. **kms_key_id** - (Optional) By default any data that is spilled to S3 is encrypted using AES-GCM and a randomly generated key. Setting a KMS Key ID allows your Lambda function to use KMS for key generation for a stronger source of encryption keys. (e.g. a7e63k4b-8loc-40db-a2a1-4d0en2cd8331)
 4. **disable_spill_encryption** - (Optional) Defaults to False so that any data that is spilled to S3 is encrypted using AES-GMC either with a randomly generated key or using KMS to generate keys. Setting this to false will disable spill encryption. You may wish to disable this for improved performance, especially if your spill location in S3 uses S3 Server Side Encryption. (e.g. True or False)
+5. **list_tables_page_size** - Page size used for the pagination of the ListTableMetadata API (1-50, Default: 50).
+   This specifies the maximum number of tables for which metadata will be retrieved by each paginated request.
+   For additional information, see [ListTableMetadata Pagination](https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-sdk/README.md#ListTableMetadata-Pagination)
+   in the SDK's README.
```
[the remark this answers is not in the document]
shurvitz: Removed

athena-federation-sdk/README.md
```diff
@@ -139,6 +139,40 @@ You can find example MetadataHandlers by looking at some of the connectors in th
...
+connectors' YAML template to generate a paginated response. This of-course relies on the individual connectors to
+provide a full list of tables for each paginated request. However, if your specific data-source provides an API for
+processing a paginated request directly at the source, the SDK's logic can be overridden to implement your own
+pagination logic:
+
+```java
+public class MyMetadataHandler extends MetadataHandler
+{
+    @Override
+    public ListTablesResponse doPaginatedListTables(final BlockAllocator allocator, final ListTablesRequest request)
```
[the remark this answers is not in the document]
shurvitz: Removed. Implementation of the pagination will now be done at the connector level.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -96,6 +101,11 @@
         implements RequestStreamHandler
 {
     private static final Logger logger = LoggerFactory.getLogger(MetadataHandler.class);
+
+    // Page size used in the pagination of the LIST_TABLES request for the ListTableMetadata API.
+    private static final String LIST_TABLES_PAGE_SIZE = "list_tables_page_size";
+    private static final int LIST_TABLES_PAGE_DEFAULT_SIZE = 50;
+
```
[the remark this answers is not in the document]
shurvitz: passed in from serde.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -146,12 +160,29 @@ public MetadataHandler(String sourceType)
...
+     * @param sourceType Used to aid in logging diagnostic info when raising a support case.
+     */
+    @VisibleForTesting
+    public MetadataHandler(EncryptionKeyFactory encryptionKeyFactory,
+                           AWSSecretsManager secretsManager,
+                           AmazonAthena athena,
+                           String sourceType,
+                           String spillBucket,
+                           String spillPrefix,
+                           int listTablesPageSize)
```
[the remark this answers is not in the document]
shurvitz: removed - no longer needed

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -197,6 +229,15 @@ protected SpillLocation makeSpillLocation(MetadataRequest request)
...
 
+    /**
+     * Returns the page size used in the pagination of the LIST_TABLES request.
+     * @return An integer representing the page size used in the pagination of the LIST_TABLES request.
+     */
+    protected int getListTablesPageSize()
+    {
+        return listTablesPageSize;
+    }
+
```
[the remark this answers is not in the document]
shurvitz: removed - no longer needed.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
shurvitz: Pagination implementation moved to the connector.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/paginators/Paginator.java
shurvitz: no longer needed.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
```diff
@@ -32,23 +32,50 @@
...
+     * @param nextToken The pagination starting point for the next page (null indicates the first paginated request).
+     * @param pageSize The page size used for pagination (-1 indicates the request should not be paginated).
+     */
+    @JsonCreator
+    public ListTablesRequest(@JsonProperty("identity") FederatedIdentity identity,
+                             @JsonProperty("queryId") String queryId,
+                             @JsonProperty("catalogName") String catalogName,
+                             @JsonProperty("schemaName") String schemaName,
+                             @JsonProperty("nextToken") String nextToken,
+                             @JsonProperty("pageSize") int pageSize)
```
avirtuos, line 73: why do you need two constructors? you should have only one, this is the tricky part of forwards and backwards compatibility

avirtuos: before you submit for a another review, please test forwards and backwards compatibility (e.g. old client calling new lambda, new client calling old lambda, and new client calling new lambda)

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
```diff
@@ -74,6 +119,7 @@ public String toString()
         return "ListTablesRequest{" +
                 "queryId=" + getQueryId() +
                 ", schemaName='" + schemaName + '\'' +
+                ", pageSize=" + pageSize +
```
avirtuos, line 122: next token?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesResponse.java
```diff
@@ -47,11 +48,27 @@
...
+     * Constructs a new ListTablesResponse object.
+     *
+     * @param catalogName The catalog name that tables were listed for.
+     * @param tables The list of table names (they all must be lowercase).
+     * @param nextToken The pagination starting point for the next request (null indicates the end of the pagination).
+     */
+    @JsonCreator
+    public ListTablesResponse(@JsonProperty("catalogName") String catalogName,
+                              @JsonProperty("tables") Collection<TableName> tables,
+                              @JsonProperty("nextToken") String nextToken)
```
avirtuos, line 66: same comment, dual constructor makes me think we didn't do the compatibility work

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDe.java
```diff
@@ -78,7 +85,16 @@ protected MetadataRequest doRequestDeserialize(JsonParser jparser, Deserializati
         {
             String schemaName = getNextStringField(jparser, SCHEMA_NAME_FIELD);
 
-            return new ListTablesRequest(identity, queryId, catalogName, schemaName);
+            String nextToken = null;
+            int pageSize = -1;
```
avirtuos, line 89: may want to add a comment for why you are doing this. Also, we may want to be stricter about this based on serialization version.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesResponseSerDe.java
```diff
@@ -93,7 +96,14 @@ protected FederationResponse doTypedDeserialize(JsonParser jparser, Deserializat
 
             String catalogName = getNextStringField(jparser, CATALOG_NAME_FIELD);
 
-            return new ListTablesResponse(catalogName, tablesList.build());
+            String nextToken = null;
```
avirtuos, line 99: same comment here about checking the serizliation version and being stricter. curious if we should do that

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDeTest.java
```diff
@@ -39,20 +38,58 @@
...
 
-    @Before
-    public void beforeTest()
+    @Test
+    public void serialize()
             throws IOException
     {
-        expected = new ListTablesRequest(federatedIdentity, "test-query-id", "test-catalog", "test-schema");
+        expected = new ListTablesRequest(federatedIdentity, "test-query-id", "test-catalog",
+                "test-schema");
```
avirtuos, line 46: why did this change?

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDeTest.java
```diff
@@ -39,20 +38,58 @@
...
+        logger.info("deserialize: deserialized[{}]", actual);
+
+        assertEquals(expected, actual);
+
+        logger.info("deserialize: exit");
     }
 
     @Test
-    public void serialize()
+    public void serializeWithNextToken()
```
avirtuos, line 85: why is this a separate test?

athena-federation-sdk/src/test/resources/serde/v2/ListTablesRequestPage.json
```diff
@@ -0,0 +1,16 @@
+{
```
avirtuos, line 1: we dont have an object called ListTableRequestPage, so what is this?

athena-federation-sdk/src/test/resources/serde/v2/ListTablesResponsePage.json
```diff
@@ -0,0 +1,12 @@
+{
+  "@type" : "ListTablesResponse",
```
avirtuos, line 2: we dont have an object called ListTableResponsePage, what am i missing?

avirtuos (changes requested):

avirtuos:
good step forward. please have @burhan94 take a peek at the next revision before me. It will save you time blocking on me for initial feedback.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
```diff
@@ -32,23 +32,50 @@
...
     @JsonCreator
     public ListTablesRequest(@JsonProperty("identity") FederatedIdentity identity,
-            @JsonProperty("queryId") String queryId,
-            @JsonProperty("catalogName") String catalogName,
-            @JsonProperty("schemaName") String schemaName)
+                             @JsonProperty("queryId") String queryId,
+                             @JsonProperty("catalogName") String catalogName,
+                             @JsonProperty("schemaName") String schemaName)
+    {
+        this(identity, queryId, catalogName, schemaName, null, -1);
```
burhan94, line 53: Can we change -1 to a variable and make it a more descriptive name? whenever anyone sees this, or any future maintainers, they shoudn't have to hunt down what it means. It can be a simple NULL_SERIALIZED_PAGE_SIZE or something like that with a comment about what it is.

burhan94 (changes requested):

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
shurvitz: Why would we want to print nextToken? 
1. You wanted it to be encrypted by the connectors which means it's sensitive info.
2. It's encrypted so the value is meaningless printed out.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
shurvitz: We are still using `version 2` of the SDK. Removing the original constructor will be the opposite of backwards compatibility. It would mean that many many files will have to change, not just in the SDK, but for every single connector as well as the control plane.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesResponse.java
shurvitz: We are still using `version 2` of the SDK. Removing the original constructor will be the opposite of backwards compatibility. It would mean that many many files will have to change, not just in the SDK, but for every single connector as well as the control plane.

athena-federation-sdk/src/test/resources/serde/v2/ListTablesResponsePage.json
shurvitz: I wanted to have 2 different JSONs one where the nextToken is `null` and one where it actually contains a String value. This way I could test both scenarios (one where we paginate and one where we don't). But, maybe it's a distinction with no difference as far as the JSON parser is concerned.

athena-federation-sdk/src/test/resources/serde/v2/ListTablesRequestPage.json
shurvitz: I wanted to have 2 different JSONs one where the nextToken is `null` and one where it actually contains a String value. This way I could test both scenarios (one where we paginate and one where we don't). But, maybe it's a distinction with no difference as far as the JSON parser is concerned.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDe.java
shurvitz: Added a `TODO` comment to enforce parsing when we upgraded to V3.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesResponseSerDe.java
shurvitz: Added a `TODO` comment to enforce parsing when we upgraded to V3.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
shurvitz: Removed original constructor and tested all scenarios using a local stack.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
shurvitz: Added

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesResponse.java
shurvitz: Removed original constructor and tested all scenarios using a local stack.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDeTest.java
shurvitz: Removed

athena-federation-sdk/src/test/resources/serde/v2/ListTablesRequestPage.json
shurvitz: Removed

athena-federation-sdk/src/test/resources/serde/v2/ListTablesResponsePage.json
shurvitz: Removed

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
shurvitz: This constructor has been removed. Added a static variable to ListTablesRequest that can be used for that purpose.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
avirtuos: Can you attach testing evidence to the PR (e.g. what test scenarios you ran and how you confirmed they were successful).

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
```diff
@@ -31,24 +31,35 @@
 public class ListTablesRequest
         extends MetadataRequest
 {
+    public static final int NULL_PAGE_SIZE = -1;
```
avirtuos, line 34: I dont think this name properly conveys what this value indicates. Also, every public member should have javadoc.
Can we rename this INF_PAGE_SIZE_VALUE or similar? -1 basically means you shouldn't limit the size of the pages.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/examples/ExampleMetadataHandlerTest.java
```diff
@@ -112,7 +113,8 @@ public void doListSchemas()
     public void doListTables()
     {
         logger.info("doListTables - enter");
-        ListTablesRequest req = new ListTablesRequest(IdentityUtil.fakeIdentity(), "queryId", "default", null);
+        ListTablesRequest req = new ListTablesRequest(IdentityUtil.fakeIdentity(),
+                "queryId", "default", null, null, NULL_PAGE_SIZE);
```
avirtuos, line 117: is this a useful test? wouldn't it be better to have a real page size and also pass a new next tokens and test pagination?

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandlerTest.java
```diff
@@ -107,7 +107,7 @@ public void setUp()
 
         when(mockMetadataHandler.doListTables(any(BlockAllocatorImpl.class), any(ListTablesRequest.class)))
                 .thenReturn(new ListTablesResponse("catalog",
-                        Collections.singletonList(new TableName("schema", "table"))));
+                        Collections.singletonList(new TableName("schema", "table")), null));
```
avirtuos, line 110: would it be more useful to have this test be paginated?

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandlerTest.java
```diff
@@ -214,7 +215,8 @@ public void doListTables()
                     return mockResult;
                 });
 
-        ListTablesRequest req = new ListTablesRequest(IdentityUtil.fakeIdentity(), queryId, catalog, schema);
+        ListTablesRequest req = new ListTablesRequest(IdentityUtil.fakeIdentity(),
+                queryId, catalog, schema, null, NULL_PAGE_SIZE);
```
avirtuos, line 219: i think we should make this test paginated

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDeTest.java
```diff
@@ -43,7 +44,8 @@
     public void beforeTest()
             throws IOException
     {
-        expected = new ListTablesRequest(federatedIdentity, "test-query-id", "test-catalog", "test-schema");
+        expected = new ListTablesRequest(federatedIdentity, "test-query-id", "test-catalog",
+                "test-schema", null, NULL_PAGE_SIZE);
```
avirtuos, line 48: can we use real values here?

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesResponseSerDeTest.java
```diff
@@ -47,7 +47,7 @@ public void beforeTest()
     {
         expected = new ListTablesResponse("test-catalog", ImmutableList.of(
                 new TableName("schema1", "table1"),
-                new TableName("schema1", "table2")));
+                new TableName("schema1", "table2")), null);
```
avirtuos, line 50: better to use real values?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
```diff
@@ -290,7 +290,7 @@ protected ListTablesResponse doListTables(BlockAllocator blockAllocator, ListTab
         }
         while (nextToken != null);
 
-        return new ListTablesResponse(request.getCatalogName(), tables);
+        return new ListTablesResponse(request.getCatalogName(), tables, null);
```
avirtuos, line 293: Why didn't you make this paginated? Its part of the SDK and the above code already supports pagination pretty easily. :) What am i missing?

avirtuos (changes requested):
Good iteration, still has a few gaps but getting much closer. Thanks for working on this and being a open to the suggestions.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesResponseSerDeTest.java
shurvitz: A null is a real value that can be passed in CP.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandlerTest.java
shurvitz: This test only checks that doListTables was called, and null is a valid nextToken.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandlerTest.java
shurvitz: GlueMetadataHandler.doListTables cannot be paginated:

shurvitz: The specific connectors' implementation of `doListTables` calls `GlueMetadataHander.doListTables` to obtain the list of tables from Glue, then adds additional tables from the actual data source. When a list is an aggregation from more than one data source, pagination based solely on one of the lists, would undoubtedly create duplication of tables. The separate lists must be combined first then paginated. Therefore, pagination should take place in the actual connectors' `doListTables` implementation after the aggregation yields a full list of tables.

athena-aws-cmdb/src/main/java/com/amazonaws/athena/connectors/aws/cmdb/AwsCmdbMetadataHandler.java
```diff
@@ -105,7 +105,8 @@ public ListSchemasResponse doListSchemaNames(BlockAllocator blockAllocator, List
     @Override
     public ListTablesResponse doListTables(BlockAllocator blockAllocator, ListTablesRequest listTablesRequest)
     {
-        return new ListTablesResponse(listTablesRequest.getCatalogName(), schemas.get(listTablesRequest.getSchemaName()));
+        return new ListTablesResponse(listTablesRequest.getCatalogName(),
+                schemas.get(listTablesRequest.getSchemaName()), null);
```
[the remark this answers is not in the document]
shurvitz: Spoke with Anthony and agreed to keep as one PR.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
shurvitz: The specific connectors' implementation of doListTables calls GlueMetadataHander.doListTables to obtain the list of tables from Glue, then adds additional tables from the actual data source. When a list is an aggregation from more than one data source, pagination based solely on one of the lists, would undoubtedly create duplication of tables. The separate lists must be combined first then paginated. Therefore, pagination should take place in the actual connectors' doListTables implementation after the aggregation yields a full list of tables.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDeTest.java
shurvitz: `null`, and `-1` are real values that will be passed from the CP. `null` indicates that this is the first round of pagination, while `-1` indicates that the list of tables can be infinite in size or not-paginated.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
shurvitz: Added javadoc and renamed to `UNLIMITED_PAGE_SIZE_VALUE`.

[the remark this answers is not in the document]
shurvitz: Added javadoc and renamed to UNLIMITED_PAGE_SIZE_VALUE (per Anthony's request).

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/examples/ExampleMetadataHandlerTest.java
shurvitz: Added pagination to ExampleMetadataHandler.doListTables implementation and added tests to ExampleMetadataHandlerTest class.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandlerTest.java
shurvitz: Added full pagination support for the GlueMetadataHandler.doListTables API and added tests.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
shurvitz: Added full pagination support for the GlueMetadataHandler.doListTables API and added tests.

shurvitz:
I've address all comments and updated documentation where I though it was applicable. Please let me know if I missed anything. Thanks for reviewing.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandlerTest.java
avirtuos: ack.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDeTest.java
avirtuos: keep in mind we don't normally discuss CP here, as Athena isn't the only thing to call these APIs.

athena-aws-cmdb/src/test/java/com/amazonaws/athena/connectors/aws/cmdb/AwsCmdbMetadataHandlerTest.java
```diff
@@ -154,7 +155,8 @@ public void doListSchemaNames()
     @Test
     public void doListTables()
     {
-        ListTablesRequest request = new ListTablesRequest(identity, queryId, catalog, "schema1");
+        ListTablesRequest request = new ListTablesRequest(identity, queryId, catalog, "schema1",
```
avirtuos, line 158: might be ok to have a convenience constructor which uses a NULL next token to make initial requests easier.

athena-example/src/test/java/com/amazonaws/connectors/athena/example/ExampleMetadataHandlerTest.java
```diff
@@ -129,7 +130,8 @@ public void doListTables()
         }
 
         logger.info("doListTables - enter");
-        ListTablesRequest req = new ListTablesRequest(fakeIdentity(), "queryId", "default", "schema1");
+        ListTablesRequest req = new ListTablesRequest(fakeIdentity(), "queryId", "default",
+                "schema1", null, UNLIMITED_PAGE_SIZE_VALUE);
```
avirtuos, line 134: as i said on the previous review, this tests doesn't seem to be testing anything. This should be testing actual pagination of the example handler.

athena-federation-sdk-tools/src/main/java/com/amazonaws/athena/connector/validation/LambdaMetadataProvider.java
```diff
@@ -100,7 +101,7 @@ public static ListTablesResponse listTables(String catalog,
     log.info("Submitting ListTablesRequest with ID " + queryId);
 
     try (ListTablesRequest request =
-                 new ListTablesRequest(identity, queryId, catalog, schema)) {
+                 new ListTablesRequest(identity, queryId, catalog, schema, null, UNLIMITED_PAGE_SIZE_VALUE)) {
```
avirtuos, line 104: we should add a todo here that the validator should ensure the connector supports pagination. We may need to wait on this until you update the connectors.

athena-federation-sdk/README.md
```diff
@@ -102,6 +102,17 @@ public class MyMetadataHandler extends MetadataHandler
...
+      //be used to paginate the response. The nextToken (request.getNextToken()) will be null if this is the first
+      //paginated request, or will contain a value of the starting point for the current request. pageSize (request
+      //.getPageSize()) will be -1 (ListTablesRequest.UNLIMITED_PAGE_SIZE_VALUE) for non-paginated requests (full
+      //list of tables should be returned), or it will be an integer value > 0. The latter indicates the maximum
+      //number of tables to return in the response (ListTablesResponse).
+      //The response (ListTablesResponse) also has a nextToken (String) attribute used for pagination. Its value
+      //should be set with the starting point for the next paginated request. If all tables have been processed, the
+      //value returned for nextToken should be null.
+      //It is noteworthy to mention that the nextToken value should be obfuscated by the connector implementing the
+      //pagination logic via some encryption algorithm.
```
avirtuos, line 115: this seems like too much detail, maybe just change the existing sentence to "Return a paginated list of tables....". This looks like a copy paste from the README so i can understand why you did it, but its just too much info in my opinion.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/examples/ExampleMetadataHandler.java
```diff
@@ -180,26 +184,70 @@ public ListSchemasResponse doListSchemaNames(BlockAllocator allocator, ListSchem
...
+     * source would likely query that source's metadata to create a real list of TableNames for the requested schema name.
+     * The request (ListTablesRequest) includes a nextToken (String) and pageSize (int) attributes that can
+     * be used to paginate the response. The nextToken (request.getNextToken()) will be null if this is the first
+     * paginated request, or will contain a value of the starting point for the current request. pageSize (request
+     * .getPageSize()) will be -1 (ListTablesRequest.UNLIMITED_PAGE_SIZE_VALUE) for non-paginated requests (full
+     * list of tables should be returned), or it will be an integer value > 0. The latter indicates the maximum
+     * number of tables to return in the response (ListTablesResponse).
+     * The response (ListTablesResponse) also has a nextToken (String) attribute used for pagination. Its value
+     * should be set with the starting point for the next paginated request. If all tables have been processed, the
+     * value returned for nextToken should be null.
```
avirtuos, line 198: its not helpful to copy paste the same large blocks of documentation everywhere. Please consider the context in which you are editing things as well as the existing documentation. Pagination is a fairly simple concept, this explaination it simply too large.  Perhaops the only useful minute detail would be to mentioned the UNLIMITED page size constant so implementers can ensure their impl handles such cases where a client wants it all at once.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
```diff
@@ -95,6 +97,10 @@
 {
     private static final Logger logger = LoggerFactory.getLogger(GlueMetadataHandler.class);
 
+    /**
+     * The maximum number of tables returned in a single response (as defined in the Glue API docs).
+     */
+    protected static final int GET_TABLES_REQUEST_MAX_RESULTS = 100;
```
avirtuos, line 103: why do you need to hard code this? seems strange to me

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
```diff
@@ -261,6 +267,15 @@ public ListTablesResponse doListTables(BlockAllocator blockAllocator, ListTables
...
      * Returns a list of tables from AWS Glue DataCatalog with optional filtering for the requested schema (aka database)
+     * The request (ListTablesRequest) includes a nextToken (String) and pageSize (int) attributes that can
+     * be used to paginate the response. The nextToken (request.getNextToken()) will be null if this is the first
+     * paginated request, or will contain a value of the starting point for the current request. pageSize (request
+     * .getPageSize()) will be -1 (ListTablesRequest.UNLIMITED_PAGE_SIZE_VALUE) for non-paginated requests (full
+     * list of tables should be returned), or it will be an integer value > 0. The latter indicates the maximum
+     * number of tables to return in the response (ListTablesResponse).
+     * The response (ListTablesResponse) also has a nextToken (String) attribute used for pagination. Its value
+     * should be set with the starting point for the next paginated request. If all tables have been processed, the
+     * value returned for nextToken should be null.
```
avirtuos, line 278: same comment as before about copy pasting large blobs of documentation

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
```diff
@@ -275,9 +290,17 @@ protected ListTablesResponse doListTables(BlockAllocator blockAllocator, ListTab
...
+        int pageSize = request.getPageSize();
         do {
             getTablesRequest.setNextToken(nextToken);
+            if (pageSize != UNLIMITED_PAGE_SIZE_VALUE) {
+                // Paginated requests will include the maxResults argument determined by the minimum value between the
+                // pageSize and the maximum results supported by Glue (as defined in the Glue API docs).
+                int maxResults = Math.min(pageSize, GET_TABLES_REQUEST_MAX_RESULTS);
+                getTablesRequest.setMaxResults(maxResults);
+                pageSize -= maxResults;
+            }
```
avirtuos, line 303: I dont understand the -= being done here.  While not go with a simpler approach and keep calling Glue until you accomuilate pageSize or run out of results. This seems entirely unnecessary just to save a few rows from Glue.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDe.java
```diff
@@ -59,6 +63,10 @@ protected void doRequestSerialize(FederationRequest federationRequest, JsonGener
             ListTablesRequest listTablesRequest = (ListTablesRequest) federationRequest;
 
             jgen.writeStringField(SCHEMA_NAME_FIELD, listTablesRequest.getSchemaName());
+
```
avirtuos, line 66: why empty lines between?

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandlerTest.java
```diff
@@ -192,38 +195,96 @@ public void doListTables()
             throws Exception
     {
         List<Table> tables = new ArrayList<>();
-        tables.add(new Table().withName("table1"));
+        tables.add(new Table().withName("table3"));
         tables.add(new Table().withName("table2"));
+        tables.add(new Table().withName("table5"));
+        tables.add(new Table().withName("table4"));
+        tables.add(new Table().withName("table1"));
```
avirtuos, line 202: why the ordering change here? seems significant but there is no comment.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandlerTest.java
```diff
@@ -192,38 +195,96 @@ public void doListTables()
...
+        logger.info("doListTables - {}", actualResponse);
+        assertEquals("Lists do not match.", fullListResponse, actualResponse);
+
+        logger.info("doListTables - Large page-size request");
+        req = new ListTablesRequest(IdentityUtil.fakeIdentity(),
+                queryId, catalog, schema, null, GET_TABLES_REQUEST_MAX_RESULTS + 50);
+        logger.info("doListTables - {}", req);
+        actualResponse = handler.doListTables(allocator, req);
+        logger.info("doListTables - {}", actualResponse);
+        assertEquals("Lists do not match.", fullListResponse, actualResponse);
```
avirtuos, line 260: should these be different tests? this one test is getting rather large.

avirtuos (changes requested):
Good cut, only minor questions remain. Also, you haven't provided testing evidence for the compatibility testing requested on the preview review. Also, what is the plan for updating the connectors? the change you made in the validator highlighted the catch-22 with releasing this to the SDK without updating the connectors. namely that we aren't enforcing new connectors be written with the proper pagination behavoir.

athena-example/src/test/java/com/amazonaws/connectors/athena/example/ExampleMetadataHandlerTest.java
shurvitz: This is the ExampleHandler for the athena-example connector (not the SDK). I've updated the tests for the SDK ExampleMetadataHandler. I thought we agreed not to change the actual connectors in this iteration.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
shurvitz: Glue has its own pagination. In order to satisfy the `pageSize` I needed to specify in the Glue Request the maximum results to retrieve (Glue allows `1-100`). Failure to do so may result in too many tables returned from Glue. If that happens, either we won't be able to satisfy the size contract in the **SDK**, or if we truncate the list (which **WS** will do), the `nextToken` from Glue will be invalidated by our logic. The solution is to retrieve the minimum between the `pageSize` and the max results that can be returned from Glue, and iterate until either the `nextToken` is null or the `pageSize` has been satisfied. Rather than hard-code the max results allowed in Glue, I opted for the above solution.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
shurvitz: 1. See comment above about Glue pagination.
2. This is the most straight-forward way to guarantees that no more than pageSize results are accumulated. I could have included a separate counter for the number of results retrieved, but that would require an additional variable that would would need to be subtracted from pageSize in order to determine the next max results size to retrieve in this iterative process. That seems a lot less intuitive than my solution.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDe.java
shurvitz: I didn't change that - it was like that before. But I will change.

athena-aws-cmdb/src/test/java/com/amazonaws/athena/connectors/aws/cmdb/AwsCmdbMetadataHandlerTest.java
shurvitz: Your comment for me from before was to only have one constructor for ListTablesRequest and ListTablesResponse. This comment seems contradictory.

athena-example/src/test/java/com/amazonaws/connectors/athena/example/ExampleMetadataHandlerTest.java
shurvitz: Added pagination to athena-example.

athena-federation-sdk-tools/src/main/java/com/amazonaws/athena/connector/validation/LambdaMetadataProvider.java
shurvitz: added TODO

athena-federation-sdk/README.md
shurvitz: Reduced.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/examples/ExampleMetadataHandler.java
shurvitz: Reduced.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
shurvitz: Reduced

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandlerTest.java
shurvitz: added comment

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandlerTest.java
shurvitz: moved some of the mock dependencies and reduced the size substantially.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDe.java
shurvitz: I wouldn't say that the extra separation is absolutely needed, but to me it makes it look cleaner to be able to see each field being written.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListTablesRequest.java
avirtuos: evidence provided offline

athena-aws-cmdb/src/test/java/com/amazonaws/athena/connectors/aws/cmdb/AwsCmdbMetadataHandlerTest.java
avirtuos: previously you had two constructors - one with pagination and one without. I'm proposing a 2nd constructor that has a convince of 'null' for the NextToken but still requires the pageSize. This doesn't leave an avenue for ignoring pagination support like your constructors did. That is the difference here. Its a suggestion, not a requirement so won't hold back the PR. Just something to think about.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
avirtuos: i don't follow. What example?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDe.java
avirtuos: this is not consistent with our styling, can you remove the extra lines?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
avirtuos: lets remove this. getting a few extra rows (max 99) from Glue seems preferable to have this hard coded incase Glue's behavior ever changes in the future.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandlerTest.java
```diff
@@ -100,6 +103,28 @@
...
+    // the number of tables in the unPaginatedTables list (or pageSize has UNLIMITED_PAGE_SIZE_VALUE).
+    private final ListTablesResponse fullListResponse = new ListTablesResponse(catalog,
+            new ImmutableList.Builder<TableName>()
+                    .add(new TableName(schema, "table1"))
+                    .add(new TableName(schema, "table2"))
+                    .add(new TableName(schema, "table3"))
+                    .add(new TableName(schema, "table4"))
+                    .add(new TableName(schema, "table5"))
+                    .build(), null);
+
```
avirtuos, line 127: this is exactly what I asked for. I asked for us to split up the tests. Moving this that are only needed by 1 function onto the class itself does shorten that function but it doesn't accomplish the goal of having smaller independent tests.

avirtuos (changes requested):
Pretty much ready, couple minor things that were called out in previous iterations. You may be rushing a bit and that is creating some churn for you as you take half measures to get this shipped instead of targeting the actual feedback (e.g. copy stuff from a function to the class in order to shorten the function when the feedback was to break up the test). Regardless, you have only minor suggestions left. Let's see if we can get it over the line today or as your time allows. Thanks for working on this. It will be a good addition.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v2/ListTablesRequestSerDe.java
shurvitz: Removed extra blank lines.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandlerTest.java
shurvitz: Broke up the tests into three: doListTablesWithUnlimitedPageSize, doListTablesWithLargePageSize, and doListTablesWithPagination.

avirtuos (approved):
Ship-It

burhan94 (approved):

PR #392, opened 2021-04-13
Publish Federation SDK to Maven Central

jobrandw:
*Description of changes:*

This adds a git action to publish the Federation SDK to Maven Central for new releases. 
Some additional metadata and plugins had to be added to satisfy the requirements. https://central.sonatype.org/publish/publish-maven/.

All artifacts published to the central repository need to be signed with [PGP](https://central.sonatype.org/publish/requirements/gpg/), which the `maven-gpg-plugin` handles. The private key will be uploaded to github as a [encrypted secret](https://docs.github.com/en/actions/reference/encrypted-secrets) that can then be referenced in the workflow.

### DIFF
--- a/.github/workflows/publish_to_maven_central.yml
+++ b/.github/workflows/publish_to_maven_central.yml
@@ -1,0 +1,34 @@
+name: Release
+
+on:
+  release:
+    types: [ published ]
+
+jobs:
+  release:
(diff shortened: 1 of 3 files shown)


burhan94 (approved):
LGTM

Issue #410, opened 2021-05-12
[BUG] Pagination token is not handled properly with connector athena-timestream


ningqian-dev:
**Describe the bug**
Pagination token is not handled properly in athena-timestream connector which leads to empty query result when the dataset is large

**To Reproduce**
1. Create a TimeStream table and ingest 500MB data
2. Setup the TimeStream Athena Connector
3. Query from Athena Connector with an expensive query: "Select * from your_table"
4. Empty result returned

**Expected behavior**
All the data in the TimeStream should be returned properly

**Screenshots / Exceptions / Errors**
If applicable, add screenshots, exception stack traces, or error messages to help explain your problem.  Stack traces for exceptions thrown by your Connector can be found in CloudWatch Logs under the 'aws/lambda/<Lambda_function_name>' log group.

**Connector Details (please complete the following information):**
 - Version: [2021.18.1]
 - Name [timestream]
 - Athena Query IDs [114b28d1-70ce-49d3-a22c-e90611890e2b] us-east-1 region

**Additional context**
I have made a fix locally and tested. The issue seems related to the behavior that TimeStream can return empty result with a nextToken when the query is in progress and will take a long period of time to complete.

burhan94:
hi ningqian, 

thank you for reporting this and submitting the PR. I will take a look in the AM

ningqian-dev:
Hey,

Updates here?

Thanks,
Qian

From: burhan94 ***@***.***>
Reply-To: awslabs/aws-athena-query-federation ***@***.***>
Date: Thursday, May 13, 2021 at 8:02 PM
To: awslabs/aws-athena-query-federation ***@***.***>
Cc: Qian Ning ***@***.***>, Author ***@***.***>
Subject: Re: [awslabs/aws-athena-query-federation] [BUG] Pagination token is not handled properly with connector athena-timestream (#410)


hi ningqian,

thank you for reporting this and submitting the PR. I will take a look in the AM

avirtuos:
An earlier comment from @burhan94 suggests you raised a PR but I couldn't find a PR from you. Is this still an issue? if so, are you sure the table you are querying had data in it? I wasn't able to reproduce.

ningqian-dev:
Hey,

I have raised the PR and it has been approved.

Thanks

avirtuos:
ah i see. it. will merge shortly.

PR #412, opened 2021-05-13
[Issue-410] Continue query when the nextToken is returned

ningqian-dev:
*Issue #, if available:*
Pagination token is not handled properly in athena-timestream connector which leads to empty query result when the dataset is large. See issue-410

*Description of changes:*
1. Continue the query as long as the pagination token is returned even though the result is empty. TimeStream returns empty result with a pagination token when the query takes a longer time to finish  
2. Pass the query string even the pagination token is supplied because this is required per query API documentation: https://docs.aws.amazon.com/timestream/latest/developerguide/API_query_Query.html#timestream-query_Query-request-QueryString

*Test*
I deployed the code to you AWS account and verified the fix


burhan94 (approved):

avirtuos:
this is for #410

PR #414, opened 2021-05-18
Athena neptune bugfix - Null value handling and Type casting 

sandeepveldi:
*Issue #, if available:*

*Description of changes:*
Fixed the null value handling and type casting issues in the Row Writer.


athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/TypeRowWriter.java
```diff
@@ -38,18 +38,19 @@
 
 import java.util.ArrayList;
 import java.util.Map;
+
 /**
  * This class is a Utility class to create Extractors for each field type as per
  * Schema
  */
-public final class TypeRowWriter
+public final class TypeRowWriter
```
avirtuos, line 46: i think your IDE is using a different style than we are. can you remove these extra whitespace changes? there are many of them

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneRecordHandlerTest.java
```diff
@@ -167,7 +167,7 @@ private void buildGraphTraversal() {
                 Vertex vertex1 = tinkerGraph.addVertex(T.label, "default");
                 vertex1.property("property1", 10);
                 vertex1.property("property2", "string1");
-                vertex1.property("property3", 12.4);
+                vertex1.property("property3", 12);
```
avirtuos, line 170: why this change? can you add another property instead of changing existing?

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneRecordHandlerTest.java
```diff
@@ -181,7 +181,6 @@ private void buildGraphTraversal() {
                 Vertex vertex3 = tinkerGraph.addVertex(T.label, "default");
                 vertex3.property("property1", 9);
                 vertex3.property("property2", "string3");
-                vertex3.property("property3", 15.4);
                 vertex3.property("property4", true);
```
avirtuos, line 184: why removing?

avirtuos (changes requested):
looks pretty good but the test case changes were unclear to me. can you give an example of what caused the issue before?

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneRecordHandlerTest.java
sandeepveldi: We want to add test case to test the null value handling when a particular field value is null/empty

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/TypeRowWriter.java
sandeepveldi: Will update the code by removing white spaces.

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneRecordHandlerTest.java
sandeepveldi: Also, the bugs that we found are:

1. Failure to detect null values returned from the data store and handling those gracefully before returning the result set to Athena.
2. Type casting issues for certain data types such as Double.

avirtuos: i'm not sure this answers my question of why you removed the 12.4 test case as opposed to adding a new one.

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneRecordHandlerTest.java
avirtuos: you removed a test case here, can you add a new property to test the null/empty case?

avirtuos (changes requested):

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneRecordHandlerTest.java
abhishekpradeepmishra: added new test cases

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneRecordHandlerTest.java
abhishekpradeepmishra: added new test cases

avirtuos (approved):

PR #419, opened 2021-05-24
2021.21.1 release

jobrandw:
Updates the poms and yaml files for the 2021.21.1 release

### DIFF
--- a/athena-aws-cmdb/pom.xml
+++ b/athena-aws-cmdb/pom.xml
@@ -20,7 +20,7 @@
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-athena-federation-sdk</artifactId>
-            <version>2021.21.1</version>
+            <version>${aws-athena-federation-sdk.version}</version>
         </dependency>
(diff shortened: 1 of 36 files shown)


athena-aws-cmdb/pom.xml
```diff
@@ -20,7 +20,7 @@
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-athena-federation-sdk</artifactId>
-            <version>2021.18.1</version>
+            <version>2021.21.1</version>
```
janmran, line 23: Would it be a good idea to update all the connector poms that have hardcoded sdk versions to be like the vertica pom that uses <aws-athena-federation-sdk.version> ?

jobrandw: thats a good idea.  i'll update them

jobrandw: I updated the connectors and integ test to use `<aws-athena-federation-sdk.version>`

janmran (approved):

PR #445, opened 2021-06-22
Athena Neptune Connector - support to connect IAM enabled Neptune cluster

abhishekpradeepmishra:
*Issue #, if available:*

*Description of changes:*

Added support to connect IAM enabled Neptune cluster


athena-neptune/athena-neptune.yaml
```diff
@@ -72,6 +75,8 @@ Resources:
           neptune_endpoint: !Ref NeptuneClusterEndpoint
           neptune_port: !Ref NeptunePort
           neptune_cluster_res_id: !Ref NeptuneClusterResourceID
+          iam_enabled: !Ref IAMEnabled
+          SERVICE_REGION: !Ref AWS::Region
```
avirtuos, line 79: Why is this needed? Can't this be derived from the Lambda runtime env?

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneConnection.java
```diff
@@ -21,21 +21,27 @@
...
 import org.apache.tinkerpop.gremlin.process.traversal.AnonymousTraversalSource;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
 
 public class NeptuneConnection
 {
-     private static Cluster cluster = null;
+    private static Cluster cluster = null;
 
-    NeptuneConnection(String neptuneEndpoint, String neptunePort)
+    NeptuneConnection(String neptuneEndpoint, String neptunePort, Boolean enabledIAM)
```
avirtuos, line 33: Why Boolean (boxed type) as opposed to Boolean (non-boxed)?

avirtuos: You may need to add a null check somewhere in order to use the non-boxed form but I think that is preferred for how you are using this as null seems to be akin to false for this usage.

avirtuos (commented):
Two minor comments but this seems ready to ship otherwise.

athena-neptune/athena-neptune.yaml
abhishekpradeepmishra: SERVICE_REGION is needed by the SignV4 SDK SDK

https://docs.aws.amazon.com/neptune/latest/userguide/iam-auth-connecting-gremlin-java.html
https://github.com/aws/amazon-neptune-gremlin-java-sigv4/blob/master/src/main/java/com/amazon/neptune/gremlin/driver/sigv4/ChainedSigV4PropertiesProvider.java

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneConnection.java
abhishekpradeepmishra: can convert this to primitive type boolean.  yes, the default value is false, even if user doesn't supply a value. Boolean.parseBoolean(System.getenv("iam_enabled")) will return false.

avirtuos (approved):

PR #452, opened 2021-06-30
Sm doc update

janmran:
*Description of changes:*
Updated READMEs with AWS Secret Manager to include:

> To use the Athena Federated Query feature with AWS Secrets Manager, you must configure an Amazon VPC private endpoint for Secrets Manager. For more information, see Create a Secrets Manager VPC Private Endpoint in the AWS Secrets Manager User Guide.

Updated HBase README's connection string
from:

> master_hostname:zookeeper_port:hbase_port
to

> master_hostname:hbase_port:zookeeper_port


janmran:
> Some of these comments are NOT accurate. You only need a private endpoint if the VPC you run your lambda function in does not have access to these services via an internet gateway.

Oh I see, will make an update. Was referencing https://docs.aws.amazon.com/athena/latest/ug/connect-to-a-data-source-lambda.html

jobrandw:
The hbase connection string url needs to be updated in the [HBaseConnectionFactory](https://github.com/awslabs/aws-athena-query-federation/blob/20bfdd4b2cb3dcef4337dee469909faec10bf41b/athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/connection/HbaseConnectionFactory.java) and [yaml](https://github.com/awslabs/aws-athena-query-federation/blob/f0739105177a40b31a7f86681d81958e4a44ac02/athena-hbase/athena-hbase.yaml#L49) too.

jobrandw (approved):
lgtm

PR #453, opened 2021-07-01
2021.27.1 release

jobrandw:
Updates the pom and yaml files for the 2021.21.1 release.

I updated the `httpclient` dependency because of the security vulnerability found [here](https://github.com/awslabs/aws-athena-query-federation/security/dependabot). 

Github also suggest updating `io.netty:netty-all` to `4.1.42`, which is what we're currently [using](https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-neptune/pom.xml#L103).  I checked [maven](https://mvnrepository.com/artifact/io.netty/netty-all) and all versions are suffixed with `.Final`.  I'm going to dismiss the security alert after this merges, since there is no issue.  

### DIFF
--- a/athena-aws-cmdb/pom.xml
+++ b/athena-aws-cmdb/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>aws-athena-query-federation</artifactId>
         <groupId>com.amazonaws</groupId>
-        <version>${aws-athena-query-federation.version</version>
+        <version>1.2</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/athena-aws-cmdb/pom.xml
+++ b/athena-aws-cmdb/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>aws-athena-query-federation</artifactId>
         <groupId>com.amazonaws</groupId>
-        <version>1.1</version>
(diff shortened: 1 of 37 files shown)


pom.xml
```diff
@@ -7,15 +7,15 @@
     <groupId>com.amazonaws</groupId>
     <artifactId>aws-athena-query-federation</artifactId>
     <packaging>pom</packaging>
-    <version>1.1</version>
+    <version>1.2</version>
```
jobrandw, line 10: I tried making this a property, so we could just reference it in the connectors but got the error `Properties in parent definition are prohibited`.

janmran (dismissed):

burhan94:
Lets change the version from 1.2 to 1.1.1

athena-aws-cmdb/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>aws-athena-query-federation</artifactId>
         <groupId>com.amazonaws</groupId>
-        <version>1.1</version>
+        <version>1.2</version>
```
burhan94, line 8: lets do 1.1.1 instead.

burhan94 (changes requested):

janmran (approved):

burhan94 (approved):

PR #454, opened 2021-07-08
Add snowflake JDBC Connector

rs-alexis:
Creating a Snowflake connector


athena-jdbc/src/main/java/com/amazonaws/connectors/athena/jdbc/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -0,0 +1,169 @@
...
+        return super.escapeNamePattern(name, escape).toUpperCase();
+    }
+
+    @Override
+    protected ResultSet getColumns(final String catalogName, final TableName tableHandle, final DatabaseMetaData metadata)
+            throws SQLException
+    {
+        String escape = metadata.getSearchStringEscape();
+        LOGGER.warn(
+          "[SNOWFLAKE DEBUG] catalogName: {}, SchemaName: {}, TableName: {}",
```
avirtuos, line 97: Logging to warm but text says debug

athena-jdbc/src/main/java/com/amazonaws/connectors/athena/jdbc/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -0,0 +1,169 @@
...
+                catalogName,
+                escapeNamePattern(tableHandle.getSchemaName(), escape),
+                escapeNamePattern(tableHandle.getTableName(), escape),
+                null);
+    }
+
+    @Override
+    public Schema getPartitionSchema(final String catalogName)
+    {
+        LOGGER.warn("[SNOWFLAKE DEBUG] getPartitionSchema ");
```
avirtuos, line 112: Same about log channel

athena-jdbc/src/main/java/com/amazonaws/connectors/athena/jdbc/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -0,0 +1,169 @@
...
+    @Override
+    public GetSplitsResponse doGetSplits(BlockAllocator blockAllocator, GetSplitsRequest getSplitsRequest)
+    {
+        LOGGER.info("doGetSplits: enter - " + getSplitsRequest);
+
+        String catalogName = getSplitsRequest.getCatalogName();
+        Set<Split> splits = new HashSet<>();
+
+        Split split = Split.newBuilder(makeSpillLocation(getSplitsRequest), makeEncryptionKey()).build();
+        splits.add(split);
```
avirtuos, line 130: Does snowflake support any partitioning or parallelism that we can exploit to create multiple splits?

avirtuos (changes requested):
Overall seems reasonable, few comments to address. Also didn't see any unit tests (maybe I missed them as I reviewed this on mobile) but both unit and integration tests will be required before we can accept this into the project as a supported connector.

avirtuos:
Also, thanks so much for the contribution!

rs-alexis:
> Overall seems reasonable, few comments to address. Also didn't see any unit tests (maybe I missed them as I reviewed this on mobile) but both unit and integration tests will be required before we can accept this into the project as a supported connector.

@avirtuos Thanks for your feedback, This is still a draft no the final PR yet
I'll add Unit Tests

athena-jdbc/src/main/java/com/amazonaws/connectors/athena/jdbc/snowflake/SnowflakeMetadataHandler.java
rs-alexis: Yeah the PR is not ready yet, shouldn't have pushed my debug lol

athena-jdbc/src/main/java/com/amazonaws/connectors/athena/jdbc/snowflake/SnowflakeMetadataHandler.java
rs-alexis: That's the big question, Snowflake supports micro-partitioning but not sure yet how that would fit in the structure of the code yet (https://docs.snowflake.com/en/user-guide/tables-clustering-keys.html)

I'm open to feedback on that

avirtuos:
is this ready for final review? or are you still working on it?

rs-sathya:
@avirtuos Yes, this should be ready for final review btw

henrymai:
Thank you for this contribution but we now have a snowflake connector: https://github.com/awslabs/aws-athena-query-federation/tree/master/athena-snowflake

If there is a feature in this pull request that is missing from the connector above, please open a new pull request to update the referenced connector. Thanks.

PR #467, opened 2021-08-11
Add Cluster and SSL support to Redis Connector

janmran:
*Description of changes:*
Adding Cluster and SSL/TLS support to the Redis Connector. The Jedis Client was replaced with [Lettuce](https://lettuce.io/core/release/reference/#basic-usage) Client which provided better cluster support. Lettuce has separate standalone and cluster api objects (same as Jedis) so a wrapper was added to contain either a standalone or cluster api.

New flags were added:
**redis-ssl-flag**: optional, default false. True will use ssl for the connection to a Redis instance
**redis-cluster-flag**: optional, default false. True will enable cluster support for a Redis instance


athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/lettuce/RedisConnectionFactory.java
```diff
@@ -0,0 +1,115 @@
...
+          connection = getOrCreateStandaloneCon(endpointParts[0], Integer.parseInt(endpointParts[1]), endpointParts[2],
+                                                sslEnabled);
+        }
+      }
+      else {
+        throw new IllegalArgumentException("Redis endpoint format error.");
+      }
+      clientCache.put(conKey, connection);
+    }
+    return connection;
```
avirtuos, line 71: do we need to 'test' this connection to ensure its still usable?

avirtuos (changes requested):
looks good overall, only one question.

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/lettuce/RedisConnectionFactory.java
janmran: good question, like something similar to the HBaseConnection? I'll take a look

janmran: Just checked, Lettuce has a [ConnectionWatchdog](https://lettuce.io/core/release/api/io/lettuce/core/protocol/ConnectionWatchdog.html) that monitors each connection and automatically reconnects if the connection is lost.

avirtuos (approved):
LGTM

burhan94:
Hey, so I just wanted to add one comment on the connection string. If I remember correctly, lettuce allows users to specify SSL enabled clusters by adding an extra s to the front of the URI scheme (eg: rediss:// for ssl and redis:// for non-ssl)

can you check if it’s better to keep it this way vs using the ssl flag in your implementation?

janmran:
> Hey, so I just wanted to add one comment on the connection string. If I remember correctly, lettuce allows users to specify SSL enabled clusters by adding an extra s to the front of the URI scheme (eg: rediss:// for ssl and redis:// for non-ssl)
> 
> can you check if it’s better to keep it this way vs using the ssl flag in your implementation?

Lettuce does but, I think things would get messier. Lettuce URI Syntax:
`redis :// [[username :] password@] host [:port][/database]
          [?[timeout=timeout[d|h|m|s|ms|us|ns]]`
`rediss :// [[username :] password@] host [: port][/database]
           [?[timeout=timeout[d|h|m|s|ms|us|ns]]`
Our current syntax: `host:port[:password]`
So either customers would have to change existing URIs or we would have to infer. Also, password comes before host in the Lettuce URI

PR #475, opened 2021-08-22
Fix typ, use AES-GCM in README

pingleig:
*Issue #, if available:*

N/A

*Description of changes:*

Some places are using `AES-GMC` instead of `AES-GCM`, and `GCM` seems to be the right one based on wiki https://en.wikipedia.org/wiki/Galois/Counter_Mode


janmran (approved):
lgtm

PR #478, opened 2021-08-27
Fix Default Glue Catalog

janmran:
*Description of changes:*

If Env variable `glue_catalog` is blank, the glue catalog used should be the account where the lambda function exists. This isn't case for Cross account federated queries:
1. Account A executes cross account federated query to use Account B's function
2. Account B's function will then try to read the glue catalog of Account A by default

This change will update step 2 so that Account B's function, when invoked by Account A, will read Account B's glue catalog by default.

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
@@ -59,6 +59,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Matcher;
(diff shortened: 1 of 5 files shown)


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/GlueMetadataHandler.java
```diff
@@ -199,6 +202,14 @@ protected String getCatalog(MetadataRequest request)
     {
         String override = System.getenv(CATALOG_NAME_ENV_OVERRIDE);
         if (override == null) {
+            if (request.getContext() != null) {
```
jobrandw, line 205: Looks good, but can you add some tests for `getCatalog`?

janmran: good looks, done

jobrandw (approved):
lgtm

PR #479, opened 2021-09-01
Fix JDBC password encoding

janmran:
*Description of changes:*

Special characters in Secrets Manager passwords are causing password authentication failures from the resulting jdbc connection string with urlencoded password.

This change fixes that by not including the credentials in the connection string and having to wonder which special characters need to be url encoded for each implementation (redshift, mysql, postgres).

Was able to test with a Secrets manager test password `Ja123%20!&1(){}[]^*`~-_<>?` on all three (redshift, mysql, postgres) And Jdbc Integ tests passed


jobrandw (commented):
Do we need to change any documentation in the readme about this change?

janmran:
> Do we need to change any documentation in the readme about this change?

Checked doc for password restrictions, limitations, and none are listed. No references to the previous encoding done either. So I would say no change is needed

jobrandw (approved):

PR #482, opened 2021-09-09
Update S3 Bucket Policy

janmran:
*Description of changes:*

Updating the S3 bucket policy in the publish script to contain `aws:SourceAccount` context key to address confused deputy issue. 
Reference to SAR docs with example policy using `aws:SourceAccount`: https://docs.aws.amazon.com/serverlessrepo/latest/devguide/serverlessrepo-how-to-publish.html#publishing-application-through-aws-console

### DIFF
--- a/tools/publish.sh
+++ b/tools/publish.sh
@@ -39,7 +39,6 @@ if [ "$#" -lt 2 ]; then
     echo "\n1. S3_BUCKET used for publishing artifacts to Lambda/Serverless App Repo.\n"
     echo "\n2. The connector module to publish (e.g. athena-exmaple or athena-cloudwatch) \n"
     echo "\n3. The AWS REGION to target (e.g. us-east-1 or us-east-2) \n"
-    echo "\n4. The AWS PARTITION to target (aws, aws-cn, aws-us-gov) Defaults to aws \n"
     echo "\n\n USAGE from the module's directory: ../tools/publish.sh my_s3_bucket athena-example \n"
     exit;
 fi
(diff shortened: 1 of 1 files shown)


tools/publish.sh
```diff
@@ -39,7 +39,6 @@ if [ "$#" -lt 2 ]; then
     echo "\n1. S3_BUCKET used for publishing artifacts to Lambda/Serverless App Repo.\n"
     echo "\n2. The connector module to publish (e.g. athena-exmaple or athena-cloudwatch) \n"
     echo "\n3. The AWS REGION to target (e.g. us-east-1 or us-east-2) \n"
-    echo "\n4. The AWS PARTITION to target (aws, aws-cn, aws-us-gov) Defaults to aws \n"
```
janmran, line 42: PARTITION isn't an input

tools/publish.sh
```diff
@@ -76,6 +75,7 @@ if ! aws s3api get-bucket-policy --bucket $1 --region $REGION| grep 'Statement'
         read -p "Do you wish to proceed? (yes or no) " yn
         case $yn in
             [Yy]* ) echo "Proceeding..."
+                read -p "Enter the AWS Account ID that will be used to publish to Serverless Application Repository (limits SAR access to applications on the specified account: " account
```
jobrandw, line 78: Should we validate that account is non empty?  I can see users pressing enter and missing this.

jobrandw, line 78: Does this mean that only users from the uploading account will be able to deploy the application?

janmran: true, let me add that

janmran: Just checked, as application owner (with the updated s3 bucket policy), I was able to share my published application with another account. And the account I shared to was able to deploy it.

jobrandw (approved):

PR #489, opened 2021-09-23
Add test for bucket policy in publish.sh

janmran:
*Description of changes:*
Adding a unit test to verify the bucket policy in `publish.sh` is correct. Specifically it should contain the `aws:SourceAccount` since the SAM service principal will be interacting with the bucket on behalf of the calling identity.


janmran:
related to #482

jobrandw:
So this is just checking to make sure that the policy exists in the publish script?

janmran:
> So this is just checking to make sure that the policy exists in the publish script?

yep checking that the policy exists and that is has the sourceAccount condition.

jobrandw (approved):

PR #499, opened 2021-10-15
release 2021.41.1

burhan94:
cuttting new release.


janmran:
Think you're missing updates to the aws-athena-query-federation version number in all the poms

burhan94:
the child poms are versioned through a variable from the parent pom

janmran (approved):
lgtm

PR #521, opened 2021-11-24
Fix example query UDF syntax

kparwal:
*Description of changes:*
Example query was using incorrect `LAMBDA_INVOKE` syntax, rather than just `LAMBDA`


janmran (approved):
lgtm

PR #525, opened 2021-12-09
Add support for custom redis db number

janmran:
*Description of changes:*
Adding support to Redis Connector to allow input of custom redis logical database number (1,2,3,...) via new glue table property flag `redis-db-number`

Also fixing integ tests. Previous shortened lambda function arn input is no longer valid when creating Athena Catalog.

### DIFF
--- a/athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/RedisMetadataHandler.java
+++ b/athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/RedisMetadataHandler.java
@@ -114,6 +114,7 @@ public class RedisMetadataHandler
     protected static final String REDIS_CLUSTER_FLAG = "redis-cluster-flag";
     //Defines the redis database to use
     protected static final String REDIS_DB_NUMBER = "redis-db-number";
+    protected static final String DEFAULT_REDIS_DB_NUMBER = "0";
 
     //Used to filter out Glue tables which lack a redis endpoint.
     private static final TableFilter TABLE_FILTER = (Table table) -> table.getParameters().containsKey(REDIS_ENDPOINT_PROP);
@@ -265,6 +266,10 @@ public class RedisMetadataHandler
             throw new RuntimeException("Table is missing " + VALUE_TYPE_TABLE_PROP + " table property");
         }
 
+        if (dbNumber == null) {
+            dbNumber = DEFAULT_REDIS_DB_NUMBER; // default redis logical database
+        }
+
         logger.info("doGetSplits: Preparing splits for {}", BlockUtils.rowToString(partitions, 0));
 
         KeyType keyType;
@@ -353,7 +358,7 @@ public class RedisMetadataHandler
                     .add(SPLIT_END_INDEX, String.valueOf(endIndex))
                     .add(REDIS_SSL_FLAG, String.valueOf(sslEnabled))
                     .add(REDIS_CLUSTER_FLAG, String.valueOf(isCluster))
-                    .add(REDIS_DB_NUMBER, String.valueOf(dbNumber))
+                    .add(REDIS_DB_NUMBER, dbNumber)
                     .build();
 
             splits.add(split);
--- a/athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/RedisMetadataHandler.java
+++ b/athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/RedisMetadataHandler.java
@@ -73,7 +73,7 @@ import static io.lettuce.core.ScanCursor.INITIAL;
  * For more detail, please see the module's README.md, some notable characteristics of this class include:
  * <p>
  * 1. Uses Glue table properties (redis-endpoint, redis-value-type, redis-key-prefix, redis-keys-zset, redis-ssl-flag,
- * and redis-cluster-flag) to provide schema as well as connectivity details to Redis.
+ * redis-cluster-flag, and redis-db-number) to provide schema as well as connectivity details to Redis.
  * 2. Attempts to resolve sensitive fields such as redis-endpoint via SecretsManager so that you can substitute
  * variables with values from by doing something like hostname:port:password=${my_secret}
  */
@@ -112,6 +112,8 @@ public class RedisMetadataHandler
     protected static final String REDIS_SSL_FLAG = "redis-ssl-flag";
     //Defines if redis instance is a cluster
     protected static final String REDIS_CLUSTER_FLAG = "redis-cluster-flag";
+    //Defines the redis database to use
+    protected static final String REDIS_DB_NUMBER = "redis-db-number";
 
     //Used to filter out Glue tables which lack a redis endpoint.
     private static final TableFilter TABLE_FILTER = (Table table) -> table.getParameters().containsKey(REDIS_ENDPOINT_PROP);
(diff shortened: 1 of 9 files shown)


athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/integ/RedisIntegTestHandler.java
```diff
@@ -47,6 +47,8 @@
 
     private static final Logger logger = LoggerFactory.getLogger(RedisIntegTestHandler.class);
 
+    private static final String STANDALONE_REDIS_DB_NUMBER = "10";
```
kparwal, line 50: What exactly happens if you specify a redis db number in clustered mode?
1. If it's an error case for redis, athena should just pass that upstream
2. If it's a silent failure, do we warn the user or is it caveat emptor?
3. (Most importantly) should we add tests for it and handle it, or is it "undefined behavior" and we leave it alone?

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/integ/RedisIntegTestHandler.java
```diff
@@ -66,8 +68,8 @@ public final void handleRequest(InputStream inputStream, OutputStream outputStre
         RedisConnectionWrapper<String, String> standaloneConnection = null;
         RedisConnectionWrapper<String, String> clusterConnection = null;
         try {
-            standaloneConnection = connectionFactory.getOrCreateConn(standaloneConnectionString, false, false);
-            clusterConnection = connectionFactory.getOrCreateConn(clusterConnectionString, true, true);
+            standaloneConnection = connectionFactory.getOrCreateConn(standaloneConnectionString, false, false, STANDALONE_REDIS_DB_NUMBER);
+            clusterConnection = connectionFactory.getOrCreateConn(clusterConnectionString, true, true, null);
```
kparwal, line 72: nit: Related to above comment -- I think `null` here is a little weird. It's really "default 0" rather than "nonexistent".

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/lettuce/RedisConnectionFactory.java
```diff
@@ -38,19 +38,26 @@
...
 
-  public synchronized RedisConnectionWrapper<String, String> getOrCreateConn(String conStr, boolean sslEnabled,
-                                                                             boolean isCluster)
+  public synchronized RedisConnectionWrapper<String, String> getOrCreateConn(String conStr,
+                                                                             boolean sslEnabled,
+                                                                             boolean isCluster,
+                                                                             String dbNumber)
   {
     String conKey = conStr + sslEnabled + isCluster;
+    if (dbNumber != null && !"null".equals(dbNumber)) {
```
kparwal, line 47: Similarly relates -- if clustered mode just **doesn't support dbnum at all** then this is correct. if it's default 0, we should emulate that behavior instead.

kparwal (commented):
Some thoughts on clustered vs standalone usage of dbnumber, everything else lgtm

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/integ/RedisIntegTestHandler.java
janmran: Absolutely nothing happens. Originally I didn't know cluster wasn't supported for db number, and during integ testing I was creating the cluster connection with a db number when writing test data, but the read connection without db number was still able to read the data. So theres. a notice in redis readme, that this flag only applies to standalone. But no error occurs if it was provided for a clustered instance

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/integ/RedisIntegTestHandler.java
janmran: Well this is just for integ test. Could really be anything since cluster doesn't do anything with the dbnumber

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/lettuce/RedisConnectionFactory.java
janmran: hmm i see. if user has multiple tables to the same clustered instances but with the same db number, we should treat them all as the same connection

janmran: will make that change

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/integ/RedisIntegTestHandler.java
janmran: fixed

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/RedisMetadataHandler.java
```diff
@@ -346,6 +353,7 @@ public GetSplitsResponse doGetSplits(BlockAllocator blockAllocator, GetSplitsReq
                     .add(SPLIT_END_INDEX, String.valueOf(endIndex))
                     .add(REDIS_SSL_FLAG, String.valueOf(sslEnabled))
                     .add(REDIS_CLUSTER_FLAG, String.valueOf(isCluster))
+                    .add(REDIS_DB_NUMBER, String.valueOf(dbNumber))
```
kparwal, line 356: `.valueOf` seems redundant

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/lettuce/RedisConnectionFactory.java
```diff
@@ -38,19 +38,33 @@
...
+    return getOrCreateConn(conStr, sslEnabled, isCluster, "0");
+  }
+
+  public synchronized RedisConnectionWrapper<String, String> getOrCreateConn(String conStr,
+                                                                             boolean sslEnabled,
+                                                                             boolean isCluster,
+                                                                             String dbNumber)
   {
     String conKey = conStr + sslEnabled + isCluster;
+    if (!isCluster && dbNumber != null && !"null".equals(dbNumber)) {
```
kparwal, line 54: I think if we get rid of the `.valueOf` we can get rid of this string equality check for "null"

kparwal (commented):
nitpicks

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/lettuce/RedisConnectionFactory.java
janmran: removed .valueOf and "null" eq check

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/lettuce/RedisConnectionFactory.java
```diff
@@ -38,19 +38,33 @@
   private static final int CONNECTION_TIMEOUT_MS = 2_000;
   private final Map<String, RedisConnectionWrapper<String, String>> clientCache = new HashMap<>();
 
-  public synchronized RedisConnectionWrapper<String, String> getOrCreateConn(String conStr, boolean sslEnabled,
+  public synchronized RedisConnectionWrapper<String, String> getOrCreateConn(String conStr,
+                                                                             boolean sslEnabled,
                                                                              boolean isCluster)
+  {
+    return getOrCreateConn(conStr, sslEnabled, isCluster, "0");
```
kparwal, line 45: nit: Use the constant you created earlier

janmran: done

kparwal (approved):
great work

Issue #606, opened 2021-12-18
[FEATURE] Feature request: I want to filter DynamoDB tables into Athena


vaquarkhan:
**Is your feature request related to a problem? If yes, please describe.** Yes User want control to display only selected tables into Athena .

User has multiple teams and tables also contain KMS key few are configured via Lake formation , now I can see following issues.
-	Asking permission for KMS key which own by other team is big issue as they are not comfortable to share this keys and since I have my 4 -5 tables which I want to use in Athena blocked by others team.
-	Without keys Athena connector will give error and cannot fetch any tables , so as per current functionality either all or non tables can fetch using connector.
-	Same issue we have with Lake formation 
-	User want to display only few logs table in quicksight and since Dynamodb has no direct connectivity want to use only few logs table via Athena to display in Quicksight.

**Describe the solution you'd like**
A clear and concise description of what you want to happen. Want a filter where i can configure my tables which will display in Athena,

**Describe alternatives you've considered**

I have fixed code in following branch , raising pull request please review .
https://github.com/vaquarkhan/aws-athena-query-federation/tree/master/athena-dynamodb

**Additional context**

I have tested all possible scenario in committed code.
![image](https://user-images.githubusercontent.com/3491908/146627481-57f33fd1-c63f-47ae-8028-c9396f58bdde.png)
![image](https://user-images.githubusercontent.com/3491908/146627501-77053fba-3cdb-4bad-a017-7dc0661b39a6.png)

I have no expertise in cloud formation so i am adding Lambda parameter manually if you could add into your template cloud formation would be great help

vaquarkhan:
Pull request for review 

- https://github.com/awslabs/aws-athena-query-federation/pull/607

PR #607, opened 2021-12-18
[FEATURE] Feature request: I want to filter dynamoDB tables into Athena as configurable  list  #606

vaquarkhan:
*Issue #, if available:* [FEATURE] Feature request: I want to filter tables to list into DynamoDB #606

*Description of changes:*

User has multiple teams and tables also contain KMS key few are configured via Lake formation , now I can see following issues.

- Added new filter method 
- Updated existing method to filter tables 
- Lambda need to create manual env variable and add values  "dynamo_connector_filter"
- Requested to add cloud formation so future will be part of AWS code

### DIFF
--- a/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
+++ b/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
@@ -217,8 +217,50 @@ public class DynamoDBMetadataHandler
         if (DynamoDBConstants.DEFAULT_SCHEMA.equals(request.getSchemaName())) {
             combinedTables.addAll(tableResolver.listTables());
         }
-        return new ListTablesResponse(request.getCatalogName(), new ArrayList<>(combinedTables), null);
+        return new ListTablesResponse(request.getCatalogName(), new ArrayList<>(filterTables(combinedTables)), null);
     }
+
+ /**
+     * @author vaquar khan
+     * This method is used to filter DyanamoDB tables in Athena 
+     * We enable users to configure optional environment variables with commaseparated List
+     * dynamo_connector_filter (Optional) Can be used to fetch selected tables from DynamoDB.
+     * default the connector will attempt to get all DynamoDB tables while connecting via Athena.(e.g. *  or tbl1,tbl2,tbl3)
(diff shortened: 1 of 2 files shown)


athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
```diff
@@ -217,9 +218,51 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
...
+     * We enable users to configure optional environment variables with commaseparated List
+     * dynamo_connector_filter (Optional) Can be used to fetch selected tables from DynamoDB.
+     * default the connector will attempt to get all DynamoDB tables while connecting via Athena.(e.g. *  or tbl1,tbl2,tbl3)
+     * @param combinedTables
+     * @return
+     */
+    
+	private Set<TableName> filterTables(Set<TableName> combinedTables)
+	{
+		String dynamoDBFilter = "*";
```
henrymai, line 236: It looks like dynamoDBFilter is immediately overwritten after being initialized to `"*"` (on line 242).

This line should just be replaced with:
```
String dynamoDBFilter = System.getenv("dynamo_connector_filter");
```

And then remove line 242

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
```diff
@@ -217,9 +218,51 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
...
+		boolean flag = false;
+
+		dynamoDBFilter = System.getenv("dynamo_connector_filter");
+
+		if (null == combinedTables || null == dynamoDBFilter) {
+			return combinedTables;
+		}
+
+		if (null != dynamoDBFilter) {
+			flag = dynamoDBFilter.toUpperCase().contains("*");
```
henrymai, line 249: What if the filter is set to: `sometable1,sometable2*,sometable3` ?

I think either we should throw an exception here or implement actual wildcard or regex matching if you want to support the case where a `*` is contained in the filter rather than equals the filter.

AbdulR3hman:
closing this as it has been stale for long time.

Issue #648, opened 2021-12-27
[FEATURE] Updating the README for `athena-dynamodb`


RajasGujarathi:
**Is your feature request related to a problem? If yes, please describe.**
It is not a problem per say, but an updated and correctly reflecting [athena-dynamodb - README](https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-dynamodb/README.md) will surely help fix logical issue while making use

**Describe the solution you'd like**
Someone with working knowledge of this library to update the [athena-dynamodb - README](https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-dynamodb/README.md) to reflect to the current latest implementation

**Describe alternatives you've considered**
Nothing as such as it is more relating to documentation

**Additional context**
Not applicable

PR #649, opened 2021-12-27
 Updated Documentation and the default LambdaMemory

RajasGujarathi:
*Issue #, if available:* #648

*Description of changes:*

1. Updated the default LambdaMemory to 10240 MB from 3008 MB
2. Documentation: "disable_glue" and "glue_catalog" are not seen to be used and hence removed
3. Documentation: Updated Parameter names to match "[athena-dynamodb.yaml](https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-dynamodb/athena-dynamodb.yaml)" parameters

### DIFF
--- a/athena-dynamodb/README.md
+++ b/athena-dynamodb/README.md
@@ -10,15 +10,17 @@ This connector enables Amazon Athena to communicate with DynamoDB, making your t
 
 The Athena DynamoDB Connector exposes several configuration options via Lambda environment variables. More detail on the available parameters can be found below.
 
-1. **spill_bucket** - When the data returned by your Lambda function exceeds Lambda’s limits, this is the bucket that the data will be written to for Athena to read the excess from. (e.g. my_bucket)
-2. **spill_prefix** - (Optional) Defaults to sub-folder in your bucket called 'athena-federation-spill'. Used in conjunction with spill_bucket, this is the path within the above bucket that large
+1. **SpillBucket** - When the data returned by your Lambda function exceeds Lambda’s limits, this is the bucket that the data will be written to for Athena to read the excess from. (e.g. my_bucket)
+2. **SpillPrefix** - (Optional) Defaults to sub-folder in your bucket called 'athena-federation-spill'. Used in conjunction with spill_bucket, this is the path within the above bucket that large
 responses are spilled to. You should configure an S3 lifecycle on this location to delete old spills after X days/Hours.
-3. **kms_key_id** - (Optional) By default any data that is spilled to S3 is encrypted using AES-GCM and a randomly generated key. Setting a KMS Key ID allows your Lambda function to use KMS for key
+3. **KMSKeyId** - (Optional) By default any data that is spilled to S3 is encrypted using AES-GCM and a randomly generated key. Setting a KMS Key ID allows your Lambda function to use KMS for key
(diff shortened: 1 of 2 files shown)


athena-dynamodb/README.md
```diff
@@ -10,15 +10,17 @@ This connector enables Amazon Athena to communicate with DynamoDB, making your t
 
 The Athena DynamoDB Connector exposes several configuration options via Lambda environment variables. More detail on the available parameters can be found below.
 
-1. **spill_bucket** - When the data returned by your Lambda function exceeds Lambda’s limits, this is the bucket that the data will be written to for Athena to read the excess from. (e.g. my_bucket)
-2. **spill_prefix** - (Optional) Defaults to sub-folder in your bucket called 'athena-federation-spill'. Used in conjunction with spill_bucket, this is the path within the above bucket that large
+1. **SpillBucket** - When the data returned by your Lambda function exceeds Lambda’s limits, this is the bucket that the data will be written to for Athena to read the excess from. (e.g. my_bucket)
```
janmran, line 13: None of these had to be renamed, and doing so will make the connector not pick these environmental variables up. Ex. As an environment variable its supposed to be `spill_prefix` but in the yaml template its named `SpillPrefix`. And these config params are environmental variables the user can set.

athena-dynamodb/README.md
```diff
@@ -10,15 +10,17 @@ This connector enables Amazon Athena to communicate with DynamoDB, making your t
...
 generation for a stronger source of encryption keys. (e.g. a7e63k4b-8loc-40db-a2a1-4d0en2cd8331)
-4. **disable_spill_encryption** - (Optional) Defaults to False so that any data that is spilled to S3 is encrypted using AES-GCM either with a randomly generated key or using KMS to generate keys.
+4. **DisableSpillEncryption** - (Optional) Defaults to False so that any data that is spilled to S3 is encrypted using AES-GCM either with a randomly generated key or using KMS to generate keys.
 Setting this to false will disable spill encryption. You may wish to disable this for improved performance, especially if your spill location in S3 uses S3 Server Side Encryption. (e.g. True or False)
 5. **disable_glue** - (Optional) If present, with any value except false, the connector will no longer attempt to retrieve supplemental metadata from Glue.
 6. **glue_catalog** - (Optional) Can be used to target a cross-account Glue catalog. By default the connector will attempt to get metadata from its own Glue account.
+7. **LambdaMemory** - (Optional) Memory is the amount of memory available to your Lambda function at runtime. Set a value between 128 MB and 10240 MB
```
janmran, line 22: These are not environmental variables and shouldn't be here.

athena-dynamodb/athena-dynamodb.yaml
```diff
@@ -29,8 +29,8 @@ Parameters:
     Default: 900
     Type: Number
   LambdaMemory:
-    Description: 'Lambda memory in MB (min 128 - 3008 max).'
-    Default: 3008
+    Description: 'Lambda memory in MB (min 128 - 10240 max).'
+    Default: 10240
```
janmran, line 33: I don't think we should change the default, as that will incur more cost

RajasGujarathi: Yes, agree.

Will take back the PR.

RajasGujarathi:
Ok. Agree

Will take back the PR.

Issue #658, opened 2022-01-20
[BUG] Using Dimension name to check constraint for dimension value in connector Cloudwatch Metrics connector


Abhinav2604:
**Describe the bug**
Cloudwatch metrics connector applies wrong constraints on Dimension Value field when running queries to get metrics.
[Line in code](https://github.com/awslabs/aws-athena-query-federation/blob/2e5aa7f0fe0880856cde32c9ed528894613a82ed/athena-cloudwatch-metrics/src/main/java/com/amazonaws/athena/connectors/cloudwatch/metrics/MetricsRecordHandler.java#L214) this line is causing this bug, it uses Dimension Name to create a constraint for Dimension Value.
**To Reproduce**
Use the Cloudwatch  Metrics connector and write a Athena Query like this in
SELECT *
FROM "lambda:<athena_connector_name>"."default".metrics_sample
WHERE namespace="<desired namespace>" AND DIM_NAME="<name_of dimension to filter upon>" AND DIM_VALUE="<desired_value>"
LIMIT 10;

**Expected behavior**
Ideally it should give the row matching our query, but we would get 0 rows always.

**Connector Details (please complete the following information):**
 - Version: 2022.2.1
 - Name: Cloudwatch Metrics

henrymai:
Closing this out since it was already fixed.

PR #659, opened 2022-01-20
Corrected the filter for DIMENSION_VALUE_FIELD

Abhinav2604:
fixes issue #658 

*Description of changes:*
DIMENSION_VALUE_FIELD field was taking DIMENSION_NAME_FIELD to create a constraint for dimension value filed which was resulting in 0 results. Corrected that, now DIMENSION_VALUE_FIELD is used to create the required constraint.


burhan94:
Thanks for the contribution and for catching this bug. approved.

burhan94 (approved):

PR #676, opened 2022-02-17
Enhance integration test suite

kparwal:
Add datatype conversion integration test suite and implementation for PostgreSQL.

### DIFF
--- a/athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DynamoDbIntegTest.java
+++ b/athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DynamoDbIntegTest.java
@@ -20,16 +20,21 @@
 package com.amazonaws.athena.connectors.dynamodb;
 
 import com.amazonaws.athena.connector.integ.IntegrationTestBase;
+import com.amazonaws.services.athena.model.Datum;
 import com.amazonaws.services.athena.model.Row;
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.testng.AssertJUnit;
 import org.testng.annotations.Test;
 import software.amazon.awscdk.core.Stack;
 import software.amazon.awscdk.services.iam.Effect;
 import software.amazon.awscdk.services.iam.PolicyDocument;
 import software.amazon.awscdk.services.iam.PolicyStatement;
 
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -38,6 +43,7 @@ import java.util.UUID;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import static org.testng.AssertJUnit.assertNull;
 
 /**
  * Integration-tests for the DynamoDB connector using the Integration-test module.
@@ -47,8 +53,11 @@ public class DynamoDbIntegTest extends IntegrationTestBase {
 
     private final String dynamodbDbName;
     private final String lambdaFunctionName;
-    private final String tableName;
-    private final DdbTableUtils ddbTableUtils;
+    private final String movieTableName;
+    private final DdbTableUtils tableUtils;
+
+    private static final String DDB_PARTITION_KEY = "title";
+    private static final String DDB_SORT_KEY = "year";
 
     public DynamoDbIntegTest()
     {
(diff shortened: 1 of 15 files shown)


athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DynamoDbIntegTest.java
```diff
@@ -47,20 +53,87 @@ public class DynamoDbIntegTest extends IntegrationTestBase {
...
+     * @param genre List of movie generes.
+     */
+    private Map<String, AttributeValue> createMovieItem(String year, String title, String director,
+            List<String> cast, List<String> genre)
+    {
+        logger.info("Add item: year=[{}], title=[{}], director=[{}], cast={}, genre={}",
+                year, title, director, cast, genre);
+
+        List<AttributeValue> castAttrib = new ArrayList<>();
+        cast.forEach(value -> castAttrib.add(new AttributeValue(value)));
```
henrymai, line 121: Haven't used Java in a while, but could this be done as follows instead? (coming from scala so not sure on the exact java syntax)

List<AttributeValue> castAttrib = cast.stream().map(value -> new AttributeValue(value)).collect(Collectors.toList());

athena-elasticsearch/src/test/java/com/amazonaws/athena/connectors/elasticsearch/integ/ElasticsearchIntegTest.java
```diff
@@ -151,6 +152,7 @@ protected Optional<PolicyDocument> getConnectorAccessPolicy()
     protected void setConnectorEnvironmentVars(final Map environmentVars)
     {
         // No-op.
+//        environmentVars.put("auto_discover_endpoint", "true");
```
henrymai, line 155: Is this meant to be here?

athena-elasticsearch/src/test/java/com/amazonaws/athena/connectors/elasticsearch/integ/ElasticsearchIntegTest.java
```diff
@@ -184,6 +186,33 @@ protected void setUpTableData()
                             "David Gyasi", "Michael Caine", "Jessica Chastain", "Matt Damon", "Casey Affleck")
             ));
         }
+        try (ElasticsearchIndexUtils indexUtils = new ElasticsearchIndexUtils(domainName, TEST_DATATYPES_TABLE_NAME))
+        {
+            Map<String, Object> item = new ImmutableMap.Builder<String, Object>()
```
henrymai, line 191: Not sure what our guidelines are, but if they are to prefer Java standard library stuff over Guava, you may want to use Map.of() or Map.ofEntries instead of this Guava builder.

https://docs.oracle.com/javase/9/docs/api/java/util/Map.html#immutable

athena-elasticsearch/src/test/java/com/amazonaws/athena/connectors/elasticsearch/integ/ElasticsearchIntegTest.java
```diff
@@ -249,4 +278,48 @@ public void selectColumnWithPredicateIntegTest()
...
+        logger.info("--------------------------------------");
+
+        String query = String.format("select float8_type from %s.%s.%s;",
+                lambdaFunctionName, INTEG_TEST_DATABASE_NAME, TEST_DATATYPES_TABLE_NAME);
+        List<Row> rows = startQueryExecution(query).getResultSet().getRows();
+        if (!rows.isEmpty()) {
+            // Remove the column-header row
+            rows.remove(0);
+        }
+        List<Double> values = new ArrayList<>();
```
henrymai, line 296: Similar comment here as earlier about using .map() instead of .forEach() for these kinds of situations.

athena-hbase/src/test/java/com/amazonaws/athena/connectors/hbase/integ/HbaseIntegTest.java
```diff
@@ -384,4 +386,214 @@ public void selectColumnWithPredicateIntegTest()
...
+        logger.info("--------------------------------------");
+
+        String query = String.format("select \"datatype:int_type\" from %s.%s.%s;",
+                lambdaFunctionName, INTEG_TEST_DATABASE_NAME, TEST_DATATYPES_TABLE_NAME);
+        List<Row> rows = startQueryExecution(query).getResultSet().getRows();
+        if (!rows.isEmpty()) {
+            // Remove the column-header row
+            rows.remove(0);
+        }
+        List<Integer> values = new ArrayList<>();
```
henrymai, line 404: Similar comment here about using .map() instead of .forEach()

henrymai (approved):
Looks good to me other than the small comments on:
- map() instead forEach()
- Map.of() instead of Guava builder

athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DynamoDbIntegTest.java
```diff
@@ -47,20 +53,87 @@ public class DynamoDbIntegTest extends IntegrationTestBase {
...
+        tableUtils = new DdbTableUtils();
+    }
+
+    /**
+     * Insert rows into the newly created DDB table.
+     */
+    protected void addMovieItems()
+    {
+        tableUtils.putItem(movieTableName, createMovieItem("2014", "Interstellar", "Christopher Nolan",
+                ImmutableList.of("Matthew McConaughey", "John Lithgow", "Ann Hathaway",
```
henrymai, line 78: If we end up preferring the Java standard library over Guava, replace this with: `List.of()`

https://docs.oracle.com/javase/9/docs/api/java/util/List.html#immutable

kparwal: We use Guava everywhere here, so I kept that pattern.

athena-elasticsearch/src/test/java/com/amazonaws/athena/connectors/elasticsearch/integ/ElasticsearchIntegTest.java
kparwal: We use Guava pretty much everywhere, so I've kept that pattern for now. If we want to remove Guava, we'd do it everywhere as a separate PR.

athena-elasticsearch/src/test/java/com/amazonaws/athena/connectors/elasticsearch/integ/ElasticsearchIntegTest.java
kparwal: nope

athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DynamoDbIntegTest.java
kparwal: This code is moved not changed, so I didn't really touch it. In general yeah we could probably make this more "functional/streaming java" but since this isn't technically a change but a refactor here, I will defer that to a future PR.

athena-elasticsearch/src/test/java/com/amazonaws/athena/connectors/elasticsearch/integ/ElasticsearchIntegTest.java
kparwal: removed

athena-elasticsearch/src/test/java/com/amazonaws/athena/connectors/elasticsearch/integ/ElasticsearchIntegTest.java
henrymai: I don't feel strongly as to whether or not you should change this, which is why I approved the pull request in addition to leaving the comments before.

But if you're wondering about the reasoning:

Its a declarative vs imperative style (where declarative is preferred when possible).

More formally this (creating the list and then looping through it to add to it) is an antipattern called ITM (initialize then modify): https://www.youtube.com/watch?v=VV9vwFsaQ6U

In scala a .map() is always used for situations like this instead (just transforming one collection into another) vs a .foreach() which is meant for side effecting operations: https://stackoverflow.com/a/44042773

The youtube link above is mostly about c++ (although he has examples in python as well) and stackoverflow link is on scala, so its not a concept that is isolated to any single language.

It only seems roundabout because its not something that most Java developers are used to (until Java 8 and beyond), but this is also why they added all of this stuff to Java.

henrymai (approved):

PR #679, opened 2022-03-10
Fix redshift POM

kparwal:
We were missing a driver, adding it here. Will cut a new release shortly after. I took the version from the last known good redshift connector from release 2022.4.1.


janmran (approved):
lgtm

PR #695, opened 2022-05-06
Add PartitionLimit option to configure the snowflake connector via SAR

trianz-azam:
*Issue #, if available:*
None

*Description of changes:*
In a recently deployed [AthenaSnowflakeConnector](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/create/app?applicationId=arn:aws:serverlessrepo:us-east-1:292517598671:applications/AthenaSnowflakeConnector), there is no option named **PartitionLimit**, which is a required option to run queries from Athena Query Console. 

This option takes an integer value to limit the number of partitions and sets it to an environment variable named **partitionlimit** in the lambda configuration. 

In the PR, the issue is fixed and tested.

### DIFF
--- a/athena-snowflake/README.md
+++ b/athena-snowflake/README.md
@@ -106,10 +106,23 @@ spill_bucket    Spill bucket name. Required.
 spill_prefix    Spill bucket key prefix. Required.
 ```
 
+### PageCount parameter
+Limits the number of records per partition
+
+The default value is 500000
+
+### PartitionLimit parameter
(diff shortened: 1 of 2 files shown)


kparwal (changes requested):
Please add some notes in the README about this.

kparwal:
Looks good -- why did we go from 1000 partitions to 10 partitions though? Does 1000 often time out?

kparwal (approved):

PR #698, opened 2022-05-13
Fix for External table access issue in Azure Server less environment

trianz-raviboyina:
*Issue # INC00001062 : External Table Access Issue In Azure Server Less Environment (Synapse) (Issue Raised Directly By Client)

The issue has been raised from JDBC Driver when we try to access table metadata, so avoided getColumns() method call in case of Azure Server less environment. 
After fixing this, got another issue from driver i.e @@TRANCOUNT' is not supported
In JdbcRecordHandler, refactored conn.commit() to if condition to resolve it.

### DIFF
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtil.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtil.java
@@ -22,19 +22,16 @@ package com.amazonaws.athena.connectors.jdbc.manager;
 import com.amazonaws.athena.connectors.jdbc.connection.DatabaseConnectionConfig;
 import com.amazonaws.athena.connectors.jdbc.connection.DatabaseConnectionConfigBuilder;
 import com.google.common.collect.ImmutableMap;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Validate;
 
 import java.util.List;
 import java.util.Map;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 public final class JDBCUtil
 {
     private static final String DEFAULT_CATALOG_PREFIX = "lambda:";
     private static final String LAMBDA_FUNCTION_NAME_PROPERTY = "AWS_LAMBDA_FUNCTION_NAME";
-    private static final Pattern SYNAPSE_CONN_STRING_PATTERN = Pattern.compile("([a-zA-Z]+)://([^;]+);(.*)");
+
     private JDBCUtil() {}
 
     /**
@@ -131,21 +128,5 @@ public final class JDBCUtil
         }
 
         return recordHandlerMap.build();
-    }
-
-    public static String checkEnvironment(String databaseEngine, String url)
-    {
-        if ("synapse".equals(databaseEngine)) {
-            // checking whether it's Azure serverless environment or not based on host name
-            Matcher m = SYNAPSE_CONN_STRING_PATTERN.matcher(url);
-            String hostName = "";
-            if (m.find() && m.groupCount() == 3) {
-                hostName = m.group(2);
-            }
-            if (StringUtils.isNotBlank(hostName) && hostName.contains("ondemand")) {
-                return "azureServerless";
-            }
-        }
-        return null;
     }
 }
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtil.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtil.java
@@ -22,10 +22,13 @@ package com.amazonaws.athena.connectors.jdbc.manager;
 import com.amazonaws.athena.connectors.jdbc.connection.DatabaseConnectionConfig;
 import com.amazonaws.athena.connectors.jdbc.connection.DatabaseConnectionConfigBuilder;
 import com.google.common.collect.ImmutableMap;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Validate;
 
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public final class JDBCUtil
 {
@@ -127,5 +130,21 @@ public final class JDBCUtil
         }
 
         return recordHandlerMap.build();
+    }
+
+    public static String checkEnvironment(String databaseEngine, String url)
+    {
+        if ("synapse".equals(databaseEngine)) {
+            // checking whether it's Azure serverless environment or not based on host name
+            Matcher m = Pattern.compile("([a-zA-Z]+)://([^;]+);(.*)").matcher(url);
+            String hostName = "";
+            if (m.find() && m.groupCount() == 3) {
+                hostName = m.group(2);
+            }
+            if (StringUtils.isNotBlank(hostName) && hostName.contains("ondemand")) {
+                return "azureServerless";
+            }
+        }
+        return null;
     }
 }
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtil.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtil.java
@@ -34,6 +34,7 @@ public final class JDBCUtil
 {
     private static final String DEFAULT_CATALOG_PREFIX = "lambda:";
     private static final String LAMBDA_FUNCTION_NAME_PROPERTY = "AWS_LAMBDA_FUNCTION_NAME";
+    private static final Pattern SYNAPSE_CONN_STRING_PATTERN = Pattern.compile("([a-zA-Z]+)://([^;]+);(.*)");
     private JDBCUtil() {}
 
     /**
@@ -136,7 +137,7 @@ public final class JDBCUtil
     {
         if ("synapse".equals(databaseEngine)) {
             // checking whether it's Azure serverless environment or not based on host name
-            Matcher m = Pattern.compile("([a-zA-Z]+)://([^;]+);(.*)").matcher(url);
+            Matcher m = SYNAPSE_CONN_STRING_PATTERN.matcher(url);
             String hostName = "";
             if (m.find() && m.groupCount() == 3) {
                 hostName = m.group(2);
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
@@ -86,8 +86,6 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 /**
  * Abstracts JDBC record handler and provides common reusable split records handling.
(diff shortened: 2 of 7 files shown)


athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
```diff
@@ -295,36 +297,128 @@ private Schema getSchema(Connection jdbcConnection, TableName tableName, Schema
...
             stmt.setString(1, tableName.getSchemaName() + "." + tableName.getTableName());
             try (ResultSet dataTypeResultSet = stmt.executeQuery()) {
                 while (dataTypeResultSet.next()) {
-                    dataType = dataTypeResultSet.getString("DATA_TYPE");
-                    columnName = dataTypeResultSet.getString("COLUMN_NAME");
-                    hashMap.put(columnName.trim(), dataType.trim());
+                    columnDetails = new ArrayList<>();
+                    columnDetails.add(dataTypeResultSet.getString("DATA_TYPE").trim());
+                    columnDetails.add(dataTypeResultSet.getString("PRECISION"));
+                    columnDetails.add(dataTypeResultSet.getString("SCALE"));
```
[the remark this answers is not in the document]
trianz-raviboyina: Sure Henry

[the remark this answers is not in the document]
trianz-raviboyina: will add .trim() on these two as well

athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
```diff
@@ -295,36 +299,131 @@ private Schema getSchema(Connection jdbcConnection, TableName tableName, Schema
...
+        // add partition columns
+        partitionSchema.getFields().forEach(schemaBuilder::addField);
+        return schemaBuilder.build();
+    }
+
+    private SchemaBuilder doDataTypeConversion(HashMap<String, List<String>> columnNameAndDataTypeMap, String schemaName)
+    {
+        String columnName;
+        List<String> dataTypeDetails;
+        String dataType;
```
henrymai, line 349: Sorry for being nitpicky, but these variables (and also `columnType` on line 352) are only used inside the for loop, so don't declare them outside of that scope.

This is the general principle of "Minimize the scope of variables":

https://wiki.sei.cmu.edu/confluence/display/java/DCL53-J.+Minimize+the+scope+of+variables

https://refactoring.com/catalog/reduceScopeOfVariable.html

https://gist.github.com/kpgalligan/4cf93abd88044df71da5

trianz-raviboyina: sure Henry, will do.

athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
```diff
@@ -295,36 +299,131 @@ private Schema getSchema(Connection jdbcConnection, TableName tableName, Schema
...
+        SchemaBuilder schemaBuilder = SchemaBuilder.newBuilder();
+        // Initializing varchar type as default
+        ArrowType columnType = Types.MinorType.VARCHAR.getType();
```
henrymai, line 352: Additionally, this won't work as intended, because the default for the `n+1` iteration will be whatever the value was set to in the `nth` iteration.

Fixing the variable scope should also fix this problem.

trianz-raviboyina: agreed.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
```diff
@@ -163,7 +165,23 @@ public void readWithConstraint(BlockSpiller blockSpiller, ReadRecordsRequest rea
                 }
                 LOGGER.info("{} rows returned by database.", rowsReturnedFromDatabase);
 
-                connection.commit();
+                boolean isAzureServerless = false;
+                // checking whether it's Azure serverless environment or not based on host name
+                if("synapse".equals(this.databaseConnectionConfig.getEngine())) {
```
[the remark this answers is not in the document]
trianz-raviboyina: This is done.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
```diff
@@ -163,7 +165,23 @@ public void readWithConstraint(BlockSpiller blockSpiller, ReadRecordsRequest rea
...
+                    Matcher m = Pattern.compile("([a-zA-Z]+)://([^;]+);(.*)").matcher(connection.getMetaData().getURL());
```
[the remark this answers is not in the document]
trianz-raviboyina: Taken care.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
```diff
@@ -163,7 +163,10 @@ public void readWithConstraint(BlockSpiller blockSpiller, ReadRecordsRequest rea
                 }
                 LOGGER.info("{} rows returned by database.", rowsReturnedFromDatabase);
 
-                connection.commit();
+                // Not performing commit() in case of Azure serverless environment
```
henrymai, line 166: Can this comment expand further as to why we should not commit if we are in an Azure serverless environment?

athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
```diff
@@ -143,59 +144,59 @@ public void getPartitions(BlockWriter blockWriter, GetTableLayoutRequest getTabl
...
+             Statement st2 = connection.createStatement();
+             ResultSet resultSet = st.executeQuery(getPartitionsSt.render());
+             ResultSet resultSet2 = st2.executeQuery(rowCountSt.render())) {
+            // check whether the table have partitions or not using ROW_COUNT_QUERY
+            if (resultSet2.next()) {
+                rowCount = resultSet2.getInt("ROW_COUNT");
+                LOGGER.info("rowCount: {}", rowCount);
+            }
+            // create a single split for view/non-partition table
+            if (rowCount == 0) {
```
henrymai, line 158: `rowCount` is only used by this block, so move the if == 0 if statement inside here and move the declaration of `rowCount` into here as well.

henrymai (commented):
Just a few more comments.
I think we are getting close for this patch to be ready to be merged; apologies for so many comments.
I think @kparwal should also review this as well just in case I missed anything.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtil.java
```diff
@@ -128,4 +131,20 @@ public static Map<String, JdbcRecordHandler> createJdbcRecordHandlerMap(final Ma
 
         return recordHandlerMap.build();
     }
+
+    public static String checkEnvironment(String databaseEngine, String url)
+    {
+        if ("synapse".equals(databaseEngine)) {
+            // checking whether it's Azure serverless environment or not based on host name
+            Matcher m = Pattern.compile("([a-zA-Z]+)://([^;]+);(.*)").matcher(url);
```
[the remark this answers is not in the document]
trianz-raviboyina: It won't be compiled every time as we already made a condition to do that in case of only Synapse.
But as a standard I've made the necessary changes.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
trianz-raviboyina: Taken care.

henrymai (approved):
Approving these changes, but I think @kparwal should also take a look just in case.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
```diff
@@ -163,7 +163,12 @@ public void readWithConstraint(BlockSpiller blockSpiller, ReadRecordsRequest rea
                 }
                 LOGGER.info("{} rows returned by database.", rowsReturnedFromDatabase);
 
-                connection.commit();
+                // SqlServer jdbc driver is using @@TRANCOUNT while performing commit(), it results below RuntimeException.
```
kparwal, line 166: instead of putting this detail in the superclass, just override `readWithConstraint` to not `commit()`. Also, it's unclear to me that just removing `commit()` is enough -- do we start the transaction, and then simply never end it? Should we remove transactions entirely for this connector? all the more reason to override this function.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtil.java
```diff
@@ -128,4 +132,20 @@ public static Map<String, JdbcRecordHandler> createJdbcRecordHandlerMap(final Ma
 
         return recordHandlerMap.build();
     }
+
+    public static String checkEnvironment(String databaseEngine, String url)
```
kparwal, line 136: same as comment below, don't invert responsibilities

kparwal:
I also see from the original issue that "type `table` is not supported"

`table` appears to be a datatype in synapse, the error being that we have no conversion from native datatype to arrow. https://docs.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql?view=sql-server-ver16

can you help me understand what this has to do with the serverless pool?

trianz-raviboyina:
> I also see from the original issue that "type `table` is not supported"
> 
> `table` appears to be a datatype in synapse, the error being that we have no conversion from native datatype to arrow. https://docs.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql?view=sql-server-ver16
> 
> can you help me understand what this has to do with the serverless pool?

@kparwal 
While accessing the external table from Synapse server less SQL pool using Synapse connector, client got the below error -

image

we simulated the issue in trianz environment and understood that the issue has been raised from SQL Server JDBC Driver when we try to access table metadata through getColumns() method call.

As a fix, we are getting data type info from metadata query instead calling getColumns() and performing explicit data type conversion for all data types in case of azure server less environment.

After fixing this, got below RuntimeException on execution of connection.commit().
com.microsoft.sqlserver.jdbc.SQLServerException:  '@@TRANCOUNT' is not supported.

Solution: Evaded this line as well by doing the same environment check.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
trianz-raviboyina: Will do override in sub class.
We are not removing the transactions for this connector. Rather avoiding execution of connection.commit() only in case of Azure Serverless sql environment.

trianz-raviboyina:
> I also see from the original issue that "type `table` is not supported"
> 
> `table` appears to be a datatype in synapse, the error being that we have no conversion from native datatype to arrow. https://docs.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql?view=sql-server-ver16
> 
> can you help me understand what this has to do with the serverless pool?

kparwal (approved):

PR #707, opened 2022-05-31
Fix for Sql Server partition issue

trianz-raviboyina:
*Issue #INC00001064, Sqlserver Connector Generating $PARTITION Clauses With Nulls.*

*Included partition information to the block to fix null issue while forming the query for partitioned table,
modified corresponding mockito test case as well.*

### DIFF
--- a/athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
+++ b/athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
@@ -183,7 +183,6 @@ public class SqlServerMetadataHandler extends JdbcMetadataHandler
                  PreparedStatement preparedStatement2 = new PreparedStatementBuilder().withConnection(connection).withQuery(ROW_COUNT_QUERY).withParameters(parameters).build();
                  ResultSet resultSet = preparedStatement.executeQuery();
                  ResultSet resultSet2 = preparedStatement2.executeQuery()) {
-
                 int rowCount = 0;
                 // check whether the table have partitions or not using ROW_COUNT_QUERY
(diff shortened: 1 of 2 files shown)


athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
```diff
@@ -208,7 +208,7 @@ public void getPartitions(BlockWriter blockWriter, GetTableLayoutRequest getTabl
                     // Include the first partition because it's not retrieved from GET_PARTITIONS_QUERY
                     blockWriter.writeRows((Block block, int rowNum) ->
                     {
-                        block.setValue(PARTITION_NUMBER, rowNum, "1");
+                        block.setValue(PARTITION_NUMBER, rowNum, "1" + ":::" + partitionFunction + ":::" + partitioningColumn);
```
henrymai, line 211: `partitionFunction` and `partitioningColumn` are only ever used inside this function, so there is no need for them to be class level member variables.

Update `getPartitionFunction` to return these values rather than having it set them as class level member variables.

Additionally, what happens if `getPartitionFunction` doesn't set the value of the `partitionFunction` or `partitioningColumn` variables? (meaning when `resultSet.next()` on line 302 doesn't return true).

If this is a possible scenario, add a test to cover this scenario in addition to fixing the code to handle this scenario.

trianz-raviboyina: @henrymai 
I got your point, there will not be case wherein the resultSet.next() become false. I have tested the code thoroughly.
But if we get such scenario will analyze it and required changes will be done.

kparwal (approved):

PR #708, opened 2022-05-31
Transition to AES-GCM for UDFs

kparwal:
*Description of changes:*
Use AES-GCM per security audit findings.

What is AES with GCM?

GCM (Galois Counter Mode) is described [here](https://en.wikipedia.org/wiki/Galois/Counter_Mode). Basically instead of just regular ECB where each chunk of plaintext is encrypted with the same cipher, in GCM mode plaintext chunk is assigned a unique "count". This count is then combined with an initialization vector (IV) and encrypted with AES. This is `xor'd` with the plaintext chunk to create a chunk of ciphertext.

Here, we take the following steps in `encrypt` and `decrypt`.
### `encrypt`
1. Create an IV with `SecureRandom`.
2. Use the IV to create an AES-GCM cipher.
3. Use the cipher to encrypt the plaintext input.
4. Concatenate the IV + encrypted input.
5. Convert that to base64 and return.

### `decrypt`
1. Convert encoded text from base64 to byte array of IV + ciphertext.
2. Slice off the IV (predefined length + offset) and create a cipher.
3. Use that cipher to decrypt the ciphertext (which is what remains after the offset).
4. Return plaintext.


athena-udfs/src/test/java/com/amazonaws/athena/connectors/udfs/AthenaUDFHandlerTest.java
```diff
@@ -141,45 +130,41 @@ public void testKmsEncryption() throws Exception
...
 
     /**
-     * This UT is used to test {@link AthenaUDFHandler#encrypt(String, String)} method end-to-end.
+     * This UT is used to test encryption and decryption end-to-end.
      * It requires AWS Secret Manager setup and AWS credential setup.
      * @throws Exception
      */
     @Ignore("Enabled as needed to do end-to-end test")
     @Test
-    public void testKmsEncryptionEndToEnd() throws Exception
```
kparwal, line 168: This test was identical to `testKmsDecryptionEndToEnd` so I repurposed it to test both methods end-to-end.

athena-udfs/src/test/java/com/amazonaws/athena/connectors/udfs/AthenaUDFHandlerTest.java
```diff
@@ -105,29 +107,16 @@ public void testDecompressTruncatedInput()
...
+        byteBuffer.put(cipher.getIV());
+        byteBuffer.put(encryptedString);
 
-        String result = athenaUDFHandler.decrypt(encryptedString, DUMMY_SECRET_NAME);
-
-        assertEquals(expected, result);
-    }
-
-    @Test
-    public void testKmsEncryption() throws Exception
```
kparwal, line 121: This test basically became unworkable, as there was no way to pass in the IV to `encrypt` so that they would yield the same result without some horrific monkeypatching (which I'm not entirely sure is possible).

kparwal: Good idea, I'll take a look -- one thing I don't really want to do is leave secure random generation up to the user here, we should just do correct defaults in this case. I'm not yet sure if what you're saying implies this, rather just stating a principle I'm following.

kparwal (commented):
Looking back on this, I see the obvious issue here. GCM needs to be an input parameter, as if a customer encrypted data with the old version of this lambda and tried to decrypt with their new version, it would fail.

chngpe (approved):

henrymai (approved):

PR #718, opened 2022-06-10
Add support for turning on server side encryption for S3 spills

henrymai:
Add support for turning on server side encryption for S3 spills.


athena-aws-cmdb/athena-aws-cmdb.yaml
```diff
@@ -36,6 +36,9 @@ Parameters:
     Description: "WARNING: If set to 'true' encryption for spilled data is disabled."
     Default: 'false'
     Type: String
+  SpillSSEAlgorithm:
+    Description: 'SSE algorithm for S3 spills'
+    Type: String
```
nauy2697, line 41: Since we are allowing only aes256, maybe we should only allow the value aes256 here?

kparwal: Defaults that we may remove later can cause problems later imo. I think it's fine as-is. In theory, the other option is `KMS`, which would require another parameter specifying which key to use. That would mean we remove the check here as well.

We also do this pattern of "anything other than `x` means false` in other parameters, so overall I'm ok with this.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/data/S3BlockSpillerTest.java
```diff
@@ -155,10 +162,9 @@ public Object answer(InvocationOnMock invocationOnMock)
             assertEquals(prefix + "/" + requestId + "/" + splitId + ".1", ((S3SpillLocation) blockLocation2).getKey());
         }
 
-        verify(mockS3, times(1))
-                .putObject(eq(bucket), eq(prefix + "/" + requestId + "/" + splitId + ".0"), anyObject(), anyObject());
-        verify(mockS3, times(1))
-                .putObject(eq(bucket), eq(prefix + "/" + requestId + "/" + splitId + ".1"), anyObject(), anyObject());
+        verify(mockS3, times(2)).putObject(argument.capture());
```
henrymai, line 165: Not sure if this is correct (not familiar with mockito)

henrymai:
Removed the commit for the yaml files.

Just realized that the params in the yaml file are intended to be common params that the customers will fill out during a cloudformation deployment of a connector.

This param isn't a common param that should be set by most customers during the cloudformation deployment of the connector.

If someone feels otherwise, I can add the patch back in, but for now I think its best to leave it out since it will just set the param to "" for most customers all the time, which is wasteful.

hackett123:
Would recommend to squash your commits before the PR just to keep mainline as clean as we can.

kparwal (approved):

PR #727, opened 2022-07-01
Feature/snowflake datasource

namratas-trianz:
*Issue # INC00001069 : Snowflake Connector Doesn't Respect Schema Definition In JDBC String (Snowflake) (Issue Raised Directly By Client)*

*Description of changes:*
The issue was after adding snowflake data source, all other schemas present were coming in the database dropdown. Now we have restricted this as per the connection string, if user is selecting specific schema, then only that schema will be available in database dropdown. If no schema has been chosen, then schemas present in the database mentioned in connection string will be available.

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
@@ -24,43 +24,46 @@ import org.apache.arrow.vector.types.Types;
 import org.apache.arrow.vector.types.pojo.ArrowType;
 
 import java.time.ZoneId;
-import java.util.HashMap;
-import java.util.HashSet;
+import java.util.Arrays;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static java.util.Map.entry;
 
 /**
  * Defines the default mapping of AWS Glue Data Catalog types to Apache Arrow types. You can override these by
  * overriding convertField(...) on GlueMetadataHandler.
  */
-public enum DefaultGlueType
+public class DefaultGlueType
 {
-    INT("int", Types.MinorType.INT.getType()),
-    VARCHAR("varchar", Types.MinorType.VARCHAR.getType()),
-    STRING("string", Types.MinorType.VARCHAR.getType()),
-    BIGINT("bigint", Types.MinorType.BIGINT.getType()),
-    DOUBLE("double", Types.MinorType.FLOAT8.getType()),
-    FLOAT("float", Types.MinorType.FLOAT4.getType()),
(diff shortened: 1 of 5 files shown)


athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -506,4 +509,39 @@ protected TableName getTableFromMetadata(final String catalogName, final TableNa
...
+                String catalogName = resultSet.getString("TABLE_CATALOG");
+                // skip internal schemas
+                if (inputSchemaName != null && schemaName.equals(inputSchemaName)  && !schemaName.equals("information_schema") && catalogName.equals(inputCatalogName)) {
+                    schemaNames.add(schemaName);
+                    catalogNames.add(catalogName);
+                }
+                else if (inputSchemaName == null && !schemaName.equals("information_schema") && catalogName.equals(inputCatalogName)) {
+                    schemaNames.add(schemaName);
+                    catalogNames.add(catalogName);
+                }
```
henrymai, line 542: I think this can be simplified to:
```
boolean shouldAddSchema =
    ((inputSchemaName == null) || schemaName.equals(inputSchemaName)) && 
        (!schemaName.equals("information_schema") && catalogName.equals(inputCatalogName));

if (shouldAddSchema) {
    schemaNames.add(schemaName);
}

```

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -506,4 +509,39 @@ protected TableName getTableFromMetadata(final String catalogName, final TableNa
...
+        catch (SQLException sqlException) {
+            throw new RuntimeException(sqlException.getErrorCode() + ": " + sqlException.getMessage());
+        }
+    }
+    protected static Set<String>  listDatabaseNames(final Connection jdbcConnection)
+            throws SQLException
+    {
+        try (ResultSet resultSet = jdbcConnection.getMetaData().getSchemas()) {
+            ImmutableSet.Builder<String> schemaNames = ImmutableSet.builder();
+            ImmutableSet.Builder<String> catalogNames = ImmutableSet.builder();
```
henrymai, line 528: I don't see this set `catalogNames` being used anywhere other than being added to.

Meaning `.build()` is never called on it and it is also never read from anywhere.

So I think this can just be deleted right?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
namratas-trianz: Hi Henry, i have addressed this now.

namratas-trianz:
there is another PR raised.
https://github.com/awslabs/aws-athena-query-federation/pull/731

PR #728, opened 2022-07-01
GlueFieldLexer: handle Glue set type and fix lexComplex logic

henrymai:
Fix 1:
    GlueFieldLexer: handle Glue set type

    GlueFieldLexer before would fail on the set type, but it should really just be treated as an array would be treated.

Fix 2:
    GlueFieldLexer: refactor and fix lexing logic

    The previous version of the code hardcoded the number of inner end fields (">") to consume in the parent enclosing struct.
    This doesn't work for all situations (see the test added).

    The right way to do this is to actually just consume the closing fields properly whenever we build a complex type.
    This avoids us having to try to figure out how many of these to consume later in an enclosing struct.

    Unified and simplified the calling paths as well.

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/GlueFieldLexer.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/GlueFieldLexer.java
@@ -37,7 +37,8 @@ public class GlueFieldLexer
     private static final Logger logger = LoggerFactory.getLogger(GlueFieldLexer.class);
 
     private static final String STRUCT = "struct";
-    private static final String LIST = "array";
+    private static final String ARRAY = "array";
+    private static final String SET = "set";
 
     private static final BaseTypeMapper DEFAULT_TYPE_MAPPER = (String type) -> DefaultGlueType.toArrowType(type);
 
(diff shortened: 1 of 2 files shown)


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/GlueFieldLexer.java
```diff
@@ -89,19 +90,22 @@ private static Field lexComplex(String name, GlueTypeParser.Token startToken, Gl
...
         }
 
-        if (startToken.getValue().toLowerCase().equals(STRUCT)) {
+        final String startTokenValueLower = startToken.getValue().toLowerCase();
+
+        if (startTokenValueLower.equals(STRUCT)) {
             fieldBuilder = FieldBuilder.newBuilder(name, Types.MinorType.STRUCT.getType());
         }
-        else if (startToken.getValue().toLowerCase().equals(LIST)) {
+        else if (startTokenValueLower.equals(ARRAY) || startTokenValueLower.equals(SET)) {
```
nauy2697, line 98: what happened to list?

henrymai: What do you mean?
LIST was misnamed before, see line 40 in athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/GlueFieldLexer.java in the diff

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/GlueFieldLexer.java
```diff
@@ -60,102 +66,85 @@ default Field getField(String name, String type)
 
     public static Field lex(String name, String input)
     {
-        ArrowType typeResult = DEFAULT_TYPE_MAPPER.getType(input);
-        if (typeResult != null) {
-            return FieldBuilder.newBuilder(name, typeResult).build();
-        }
```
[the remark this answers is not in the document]
henrymai: Its not removed, see lines 91-92 in the new version of the file on the right hand side.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/GlueFieldLexer.java
```diff
@@ -59,99 +66,85 @@ default Field getField(String name, String type)
...
-        else if (typeToken.getMarker().equals(GlueTypeParser.FIELD_SEP) ||
-                typeToken.getMarker().equals(GlueTypeParser.FIELD_END)
-        ) {
-            logger.debug("lex: exit - {}", nameToken.getValue());
-            return mapper.getField(name, typeToken.getValue());
+    }
+
+    private static void expectTokenMarkerIsFieldEnd(GlueTypeParser.Token token)
+    {
+        boolean isFieldEnd = (token.getMarker() == null ||
```
nauy2697, line 143: why would null be field end? Isn't it going to come with GlueTypeParser.FIELD_END?

henrymai: FIeld end being null happens at the very end of the type string.

hackett123 (approved):

PR #729, opened 2022-07-05
Added Timestamp and Date Datatype conversion For Snowflake

Jithendar12:
Issue#INC00001071:Data types conversion (TIMESTAMP_NTZ, DATE, TIMESTAMP --> TIMESTAMPMILLI) is not working.
These columns are getting converted to VARCHAR

Description of changes:
Previsouly TIMESTAMP and DATE Datetypes are casted into TIMESTAMPMILLI which converting those DataTypes to VARCHAR On Athena. Now we have casted TIMESTAMP to DATEMILLI and DATE to DATEDAY which converts into appropriate DataType on Athena.

Please refer attached document for more details and sample queries.

[snowflake_datatype_issue.docx](https://github.com/awslabs/aws-athena-query-federation/files/9044412/snowflake_datatype_issue.docx)

### DIFF
--- a/athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
+++ b/athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
@@ -365,19 +365,26 @@ public class SnowflakeMetadataHandler extends JdbcMetadataHandler
                         columnType = Types.MinorType.INT.getType();
                     }
                     if (dataType != null && (dataType.equalsIgnoreCase("DATE"))) {
-                        columnType = Types.MinorType.TIMESTAMPMILLI.getType();
+                        columnType = Types.MinorType.DATEDAY.getType();
                     }
                     /**
-                     * Converting TIMESTAMP data type into TIMESTAMPMILLI
+                     * Converting TIMESTAMP data type into DATEMILLI
                      */
                     if (dataType != null && (dataType.equalsIgnoreCase("TIMESTAMP"))
                     ) {
-                        columnType = Types.MinorType.TIMESTAMPMILLI.getType();
+                        columnType = Types.MinorType.DATEMILLI.getType();
+                    }
+                    if (dataType != null && (dataType.equalsIgnoreCase("TIMESTAMP_LTZ"))
+                    ) {
+                        columnType = Types.MinorType.DATEMILLI.getType();
                     }
                     if (dataType != null && (dataType.equalsIgnoreCase("TIMESTAMP_NTZ"))
                     ) {
-                        LOGGER.debug(" inside TIMESTAMP ");
-                        columnType = Types.MinorType.TIMESTAMPMILLI.getType();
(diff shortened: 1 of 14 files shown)


athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -365,19 +365,26 @@ public Schema getSchema(Connection jdbcConnection, TableName tableName, Schema p
...
                     ) {
-                        columnType = Types.MinorType.TIMESTAMPMILLI.getType();
+                        columnType = Types.MinorType.DATEMILLI.getType();
+                    }
+                    if (dataType != null && (dataType.equalsIgnoreCase("TIMESTAMP_LTZ"))
+                    ) {
+                        columnType = Types.MinorType.DATEMILLI.getType();
                     }
                     if (dataType != null && (dataType.equalsIgnoreCase("TIMESTAMP_NTZ"))
                     ) {
```
henrymai, line 382: Can all of these if statements just be replaced with a static final class level Map (even higher up before my selection here), like:

    static final Map<String, ArrowType> stringToArrowTypeMap = Map.of(
        entry("DECIMAL", Types.MinorType.BIGINT.getType()),
        entry("INTEGER", Types.MinorType.INT.getType()),
        // ....
     );

And then in this code you can just do: `ArrowType columnType = stringToArrowTypeMap(dataType)`
Deleting the original definition of columnType where it just gets overwritten anyway.

Jithendar12: Sure @henrymai, will update with this

henrymai: Also I just noticed something, the original code is mapping "DECIMAL" to Arrow BIGINT, is this intentional?
Usually "DECIMAL" types should be mapped to Arrow DECIMAL instead.

henrymai: Looking at the Snowflake documentation confirms this too.

DECIMAL is the same as NUMBER and NUMBER is definitely not just a BIGINT:
https://docs.snowflake.com/en/sql-reference/data-types-numeric.html#number

Additionally the code doesn't parse the parameters to DECIMAL (or NUMBER) which specify the precision and scale.

Note that these comments aren't suggesting that you fix this in this pull request, its just something that I noticed and is a potential bug that should be fixed at some point.

Jithendar12: HI Henry,done changes as per your suggestions.please review

henrymai: Looked through it, I'm letting tests run on it now.

henrymai (approved):

Issue #664, opened 2022-01-28
[BUG] glue parse error with connector dynamodb


womblep:
**Describe the bug**
A glue table format with a column type of decimal or double causes a parse error and Athena uses the implied format. 
Changing the type to string parses correctly and Athena gets all the columns.

java.lang.RuntimeException: Parse error, expected < but found null
	at com.amazonaws.athena.connector.lambda.metadata.glue.GlueFieldLexer.lexComplex(GlueFieldLexer.java:89) ~[task/:?]
	at com.amazonaws.athena.connector.lambda.metadata.glue.GlueFieldLexer.lex(GlueFieldLexer.java:68) ~[task/:?]
	at com.amazonaws.athena.connectors.dynamodb.DynamoDBMetadataHandler.convertField(DynamoDBMetadataHandler.java:475) ~[task/:?]
	at com.amazonaws.athena.connector.lambda.handlers.GlueMetadataHandler.doGetTable(GlueMetadataHandler.java:361) ~[task/:?]
	at com.amazonaws.athena.connector.lambda.handlers.GlueMetadataHandler.doGetTable(GlueMetadataHandler.java:308) ~[task/:?]
	at com.amazonaws.athena.connectors.dynamodb.DynamoDBMetadataHandler.doGetTable(DynamoDBMetadataHandler.java:232) [task/:?]
	at com.amazonaws.athena.connector.lambda.handlers.MetadataHandler.doHandleRequest(MetadataHandler.java:250) [task/:?]
	at com.amazonaws.athena.connector.lambda.handlers.CompositeHandler.handleRequest(CompositeHandler.java:132) [task/:?]
	at com.amazonaws.athena.connector.lambda.handlers.CompositeHandler.handleRequest(CompositeHandler.java:100) [task/:?]
	at lambdainternal.EventHandlerLoader$2.call(EventHandlerLoader.java:903) [LambdaSandboxJava-byol.jar:?]
	at lambdainternal.AWSLambda.startRuntime(AWSLambda.java:349) [LambdaSandboxJava-byol.jar:?]
	at lambdainternal.AWSLambda.<clinit>(AWSLambda.java:70) [LambdaSandboxJava-byol.jar:?]
	at java.lang.Class.forName0(Native Method) ~[?:1.8.0_312]
	at java.lang.Class.forName(Class.java:348) [?:1.8.0_312]
[...]

PR #730, opened 2022-07-05
GlueFieldLexer: Add decimal support

henrymai:
Adds support for all of the following forms of decimal:
"decimal"
"decimal(12,123)"
"decimal(23,111,128)"

This change required a refactoring of DefaultGlueType that removes its use of enum.
The usage was enum seemed to be only for the purposes of the static initialization of the hash map, which is not necessary since we can just use Map.of which achieves the same thing with less code.

An extra level of indirection is also removed now that we just return "ArrowType"s from the DefaultGlueType methods rather than returning "DefaultGlueType".

Issue #664

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
@@ -24,43 +24,46 @@ import org.apache.arrow.vector.types.Types;
 import org.apache.arrow.vector.types.pojo.ArrowType;
 
 import java.time.ZoneId;
-import java.util.HashMap;
-import java.util.HashSet;
+import java.util.Arrays;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static java.util.Map.entry;
 
 /**
  * Defines the default mapping of AWS Glue Data Catalog types to Apache Arrow types. You can override these by
  * overriding convertField(...) on GlueMetadataHandler.
  */
-public enum DefaultGlueType
+public class DefaultGlueType
 {
-    INT("int", Types.MinorType.INT.getType()),
-    VARCHAR("varchar", Types.MinorType.VARCHAR.getType()),
-    STRING("string", Types.MinorType.VARCHAR.getType()),
-    BIGINT("bigint", Types.MinorType.BIGINT.getType()),
-    DOUBLE("double", Types.MinorType.FLOAT8.getType()),
-    FLOAT("float", Types.MinorType.FLOAT4.getType()),
-    SMALLINT("smallint", Types.MinorType.SMALLINT.getType()),
-    TINYINT("tinyint", Types.MinorType.TINYINT.getType()),
-    BIT("boolean", Types.MinorType.BIT.getType()),
-    VARBINARY("binary", Types.MinorType.VARBINARY.getType()),
-    TIMESTAMP("timestamp", Types.MinorType.DATEMILLI.getType()),
-    // ZoneId.systemDefault().getId() is just a place holder, each row will have a TZ value
-    // otherwise fall back to the table configured default TZ
-    TIMESTAMPMILLITZ("timestamptz", new ArrowType.Timestamp(org.apache.arrow.vector.types.TimeUnit.MILLISECOND, ZoneId.systemDefault().getId())),
-    DATE("date", Types.MinorType.DATEDAY.getType());
+    private static final String TIMESTAMPMILLITZ = "timestamptz";
+    private static final Set<String> NON_COMPARABALE_SET = Set.of("TIMESTAMPMILLITZ");
 
-    private static final Map<String, DefaultGlueType> TYPE_MAP = new HashMap<>();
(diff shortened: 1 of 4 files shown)


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
```diff
@@ -83,12 +116,12 @@ public static DefaultGlueType fromId(String id)
 
     public static ArrowType toArrowType(String id)
     {
-        DefaultGlueType result = TYPE_MAP.get(id.toLowerCase());
+        ArrowType result = TYPE_MAP.get(id.toLowerCase());
```
nauy2697, line 119: Why do we need `toArrowType` and `fromId` both seems doing the same thing

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
```diff
@@ -71,10 +74,40 @@
...
-    public static DefaultGlueType fromId(String id)
+    private static ArrowType getDecimalArrowType(String in)
     {
-        DefaultGlueType result = TYPE_MAP.get(id.toLowerCase());
+        Matcher decimalMatcher = decimalPattern.matcher(in);
+        if (!decimalMatcher.matches()) {
+            return null;
+        }
+        try {
+            int[] params = Arrays.stream(decimalMatcher.group(2).split(","))
```
hackett123, line 84: what does decimalMatcher.group(2) represent?

nauy2697:
Before we have adding this decimal support to glue schema. What was the behavior before when customer set decimal in glue? 
Just double check here as this is a behavior change

henrymai:
> Before we have adding this decimal support to glue schema. What was the behavior before when customer set decimal in glue? Just double check here as this is a behavior change

See this Issue #664

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
```diff
@@ -71,10 +74,40 @@
         this.arrowType = arrowType;
     }
 
-    public static DefaultGlueType fromId(String id)
```
hackett123, line 74: Is this just the diff cutting weird or are we replacing all ids -> DefaultGlueType to just a string to decimal?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
henrymai: It means the 2nd capture group (things surrounded by ()) in the regex.
See this: https://i.imgur.com/yLRFRAC.png

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
henrymai: Just the diff cutting weird, this function "fromId" still exists. See line 103 in the new version of the file on the right hand side.

hackett123 (approved):

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/DefaultGlueType.java
henrymai: Yeah, as mentioned on slack, there are callers to both of these methods:

```
~/workspace/aws-athena-query-federation$ git grep "DefaultGlueType\.fromId" .
athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/RedisMetadataHandler.java:        return FieldBuilder.newBuilder(name, DefaultGlueType.fromId(type)).build();
~/workspace/aws-athena-query-federation$ git grep "DefaultGlueType\.toArrowType" .
athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchGlueTypeMapper.java:     * Gets the Arrow type equivalent for the Glue type string representation using the DefaultGlueType.toArrowType
athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchGlueTypeMapper.java:     * conversion routine. The implementation of DefaultGlueType.toArrowType(type) invokes toLowerCase() on the type.
athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchGlueTypeMapper.java:        return DefaultGlueType.toArrowType(type);
athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/glue/GlueFieldLexer.java:    private static final BaseTypeMapper DEFAULT_TYPE_MAPPER = (String type) -> DefaultGlueType.toArrowType(type);
```

henrymai: Just updated the PR with having one of these call the other while retaining both methods (since one of them is used to throw on null and there are callers for this interface).

nauy2697 (approved):

PR #731, opened 2022-07-07
Feature/teradata datasource

namratas-trianz:
*Issue #, if available:*
1.Teradata Connector Errors On Non-Existent DatabaseTerm Parameter In JDBC String
2.Snowflake Connector Doesn't Respect Schema Definition In JDBC String (Snowflake) (Issue Raised Directly By Client)
*Description of changes:*
Issue 1. There was an extra default parameter added in jdbcproperties which was causing the issue. The issue was coming only above 17.x.x.x jdbc version of java.
Issue 2.The issue was after adding snowflake data source, all other schemas present were coming in the database dropdown. Now we have restricted this as per the connection string, if user is selecting specific schema, then only that schema will be available in database dropdown. If no schema has been chosen, then schemas present in the database mentioned in connection string will be available.

### DIFF
--- a/athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataMetadataHandler.java
+++ b/athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataMetadataHandler.java
@@ -348,6 +348,8 @@ public class TeradataMetadataHandler extends JdbcMetadataHandler
                     LOGGER.info("Column Name: " + columnName);
                     if (columnType != null) {
                         LOGGER.info("Column Type: " + columnType.getTypeID());
+                    }else {
+                        LOGGER.warn("Column Type is null for Column Name" + columnName);
                     }
 
                     /**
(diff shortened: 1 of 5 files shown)


athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataMetadataHandler.java
```diff
@@ -349,7 +346,9 @@ private Schema getSchema(Connection jdbcConnection, TableName tableName, Schema
                     String columnName = resultSet.getString("COLUMN_NAME");
 
                     LOGGER.info("Column Name: " + columnName);
-                    LOGGER.info("Column Type: " + columnType.getTypeID());
+                    if (columnType != null) {
+                        LOGGER.info("Column Type: " + columnType.getTypeID());
+                    }
```
henrymai, line 351: I think we should log the case where the `columnType` is null, like:

```
else
{
    Logger.warn("Column Type is null for Column Name" + columnName);
}

henrymai (approved):
Just one small comment that I left on logging that the columnType is null. Otherwise it looks good to me.

athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataMetadataHandler.java
namratas-trianz: Hi Henry, added the statement.

henrymai (approved):

PR #738, opened 2022-07-26
Add `disable_projection_and_casing` to deal with casing issues. Also undoes the revert of the Glue Lexer revert.

henrymai:
This `disable_projection_and_casing` variable supports 3 modes:
- auto
    - This disables projection and casing when we see a previously unsupported type
      and we see that the user does not have column name mapping on their table.
- true
    - This disables projection and casing unconditionally.
      This is useful when users have casing in their ddb column names but do not want to
      specify a column name mapping at all.
- false
    - This enables projection and casing unconditionally.
      This is useful when the users want projection enabled (mainly for bandwidth and latency)
      and are not worried about their columns with casing or already have a column name mapping
      for those columns.

Caveats with setting this to true or auto:

- May incur higher bandwidth usage depending on your query.
This not a problem if your lambda is in the same region as your ddb table.

- Overall latency may increase because there is a larger number of bytes being transferred and also
higher deserialization time given the larger amount of bytes.

### DIFF
--- a/athena-dynamodb/README.md
+++ b/athena-dynamodb/README.md
@@ -20,6 +20,8 @@ generation for a stronger source of encryption keys. (e.g. a7e63k4b-8loc-40db-a2
 Setting this to false will disable spill encryption. You may wish to disable this for improved performance, especially if your spill location in S3 uses S3 Server Side Encryption. (e.g. True or False)
 6. **disable_glue** - (Optional) If present, with any value except false, the connector will no longer attempt to retrieve supplemental metadata from Glue.
 7. **glue_catalog** - (Optional) Can be used to target a cross-account Glue catalog. By default the connector will attempt to get metadata from its own Glue account.
+8. **disable_projection_and_casing** - (Optional) Defaults to: auto. Can be used to disable projection and casing in order to be able to query DynamoDB tables with casing in their column names without having to specify a "columnMapping" property on their Glue table. Go here for more info [disable_projection_and_casing](#disable_projection_and_casing).
+
 
 ### Setting Up Databases & Tables in Glue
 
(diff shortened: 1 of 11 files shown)


athena-dynamodb/README.md
```diff
@@ -20,6 +20,8 @@ generation for a stronger source of encryption keys. (e.g. a7e63k4b-8loc-40db-a2
 Setting this to false will disable spill encryption. You may wish to disable this for improved performance, especially if your spill location in S3 uses S3 Server Side Encryption. (e.g. True or False)
 6. **disable_glue** - (Optional) If present, with any value except false, the connector will no longer attempt to retrieve supplemental metadata from Glue.
 7. **glue_catalog** - (Optional) Can be used to target a cross-account Glue catalog. By default the connector will attempt to get metadata from its own Glue account.
+8. **DISABLE_PROJECTION_AND_CASING_ENV** - (Optional) Can be used to disable projection and casing in order to be able to query DynamoDB tables with casing in their column names without having to specify a "columnMapping" property on their Glue table. Go here for more info [DISABLE_PROJECTION_AND_CASING_ENV](#DISABLE_PROJECTION_AND_CASING_ENV).
```
kparwal, line 23: I am not sure we should expose this option. This was a bug, it should just get fixed. If someone wants the old behavior (instead of just fixing your glue definitions) then they can use an old version of the connector. These connectors have a problem with already being too difficult to use, this just adds to that mental strain.

kparwal, line 23: nit: is there a reason this is all caps? if this is a feature, can we make this look like the other vars?

henrymai: Sure I can lowercase it instead.

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBRecordHandler.java
```diff
@@ -134,21 +136,55 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
         invokerCache.get(tableName).setBlockSpiller(spiller);
-        Iterator<Map<String, AttributeValue>> itemIterator = getIterator(split, tableName, recordsRequest.getSchema());
         DDBRecordMetadata recordMetadata = new DDBRecordMetadata(recordsRequest.getSchema());
+
+        String disableProjectionAndCasingEnvValue = System.getenv().getOrDefault(DISABLE_PROJECTION_AND_CASING_ENV, "auto").toLowerCase();
+        logger.info(DISABLE_PROJECTION_AND_CASING_ENV + " environment variable set to: " + disableProjectionAndCasingEnvValue);
+
+        boolean disableProjectionAndCasing = false;
+        if (disableProjectionAndCasingEnvValue.equals("auto")) {
+            // In the automatic case, we will try to mimic the behavior prior to the support of `set` and `decimal` types as much
```
hackett123, line 146: sorry to nit here, but why not a multi-line comment for this section?

henrymai: There are no style guidelines dictating one vs the other here.

You could imagine that I could have started out with a multi line comment and then someone else asks me to use // comments instead (which has happened before in other projects at other companies).

hackett123: okay, in that case, no problem.

hackett123 (dismissed):

nauy2697 (approved):

hackett123 (approved):

PR #741, opened 2022-07-27
Resolve Misc Issues in Integration Tests Package

hackett123:
- Integ Test Config File Fixes
- Add helper python script to validate json content of test config files
- timestream fixes for integ tests

*Issue #, if available:*
735
*Description of changes:*

### DIFF
--- a/tools/config_json_validator.py
+++ b/tools/config_json_validator.py
@@ -1,0 +1,29 @@
+import os
+import json
+import glob
+from pathlib import Path
+
+# set working directory to root dir
+root_dir = Path(__file__).parent.parent
+os.chdir(root_dir)
+
+file_paths = glob.glob('**/**/test-config.json')
+
+def remove_comments_from_json(file):
+    with open(file) as f:
+        lines = f.readlines()
+        formatted_lines = []
+        for line in lines:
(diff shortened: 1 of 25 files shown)


tools/config_json_validator.py
```diff
@@ -0,0 +1,26 @@
+import subprocess
+import os
+import json
+
+
+result = subprocess.run(['find', '../', '-name', 'test-config.json'], capture_output=True, text=True)
```
henrymai, line 6: Before this line, do this:

```
os.chdir(Path(__file__).parent)
```

This is the ensure that the working directory for the script is the script's actual parent path.

Otherwise your `find ..` command won't work as expected for people invoking the script from the root directory instead of `cd`ing into the tools directory first (a lot of people don't cd all the way into a directory to invoke a script and they shouldn't have to).

henrymai: Also consider using glob instead of find:
`glob.glob("../**/**/test-config.json")`

henrymai (dismissed):
Approved with some minor comments.

henrymai (approved):

kparwal (approved):

PR #749, opened 2022-08-03
Implemented SSL based connection logic in the oracle connector

trianz-raviboyina:
*Issue INC00001066 #,Improve Oracle Connector To Support SSL Connections To The Oracle Db*

Implemented SSL based connection logic in the connector.  
The current version of oracle connector supports ssl based connection for Amazon RDS instances only.

### DIFF
--- a/athena-oracle/README.md
+++ b/athena-oracle/README.md
@@ -32,11 +32,11 @@ A JDBC Connection string is used to connect to a database instance. Following fo
 
 Multiplexer provides a way to connect to multiple database instances using a single Lambda function. Requests are routed depending on catalog name. Use following classes in Lambda for using multiplexer.
 
-|Handler|Class|
-|---	|---	|
-|Composite Handler|OracleMuxCompositeHandler|
-|Metadata Handler|OracleMuxMetadataHandler|
-|Record Handler|OracleMuxRecordHandler|
+| Handler           | Class                      |
+|-------------------|----------------------------|
+| Composite Handler | OracleMuxCompositeHandler  |
+| Metadata Handler  | OracleMuxMetadataHandler   |
+| Record Handler    | OracleMuxRecordHandler     |
 
 
 **Parameters:**
@@ -50,13 +50,11 @@ default                         Default connection string. Required. This will b
 
 Example properties for a Oracle Mux Lambda function that supports two database instances, oracle1(default) and oracle2:
 
-|Property|Value|
-|---|---|
-|default|oracle://jdbc:oracle:thin:${Test/RDS/Oracle1}@//oracle1.hostname:port/servicename|
-|	|	|
-|oracle_catalog1_connection_string|oracle://jdbc:oracle:thin:${Test/RDS/Oracle2}@//oracle1.hostname:port/servicename|
-|	|	|
-|oracle_catalog2_connection_string|oracle://jdbc:oracle:thin:${Test/RDS/Oracle2}@//oracle2.hostname:port/servicename|
+| Property                          | Value                                                                             |
+|-----------------------------------|-----------------------------------------------------------------------------------|
+| default                           | oracle://jdbc:oracle:thin:${Test/RDS/Oracle1}@//oracle1.hostname:port/servicename |
+| oracle_catalog1_connection_string | oracle://jdbc:oracle:thin:${Test/RDS/Oracle2}@//oracle1.hostname:port/servicename |
+| oracle_catalog2_connection_string | oracle://jdbc:oracle:thin:${Test/RDS/Oracle2}@//oracle2.hostname:port/servicename |
 
 Oracle Connector supports substitution of any string enclosed like *${Test/RDS/Oracle}* with *username* and *password* retrieved from AWS Secrets Manager. Example:
 
@@ -87,15 +85,23 @@ Record Handler       OracleRecordHandler
 
 ```
 default         Default connection string. Required. This will be used when a catalog is not recognized.
+ssl             SSL. Required. Based on this flag connection will be encrypted over ssl.
 ```
 
 These handlers support one database instance and must provide `default` connection string parameter. All other connection strings are ignored.
+The current version of oracle connector supports ssl based connection for Amazon RDS instances only.
 
 **Example property for a single oracle instance supported by a Lambda function:**
 
-|Property|Value|
-|---|---|
-|default|oracle://jdbc:oracle:thin:${Test/RDS/Oracle}@//hostname:port/servicename|
+| Property | Value                                                                    |
+|----------|--------------------------------------------------------------------------|
+| default  | oracle://jdbc:oracle:thin:${Test/RDS/Oracle}@//hostname:port/servicename |
+| ssl      | no                                                                       |
+
+| Property | Value                                                                                                                                                                                |
+|----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| default  | oracle://jdbc:oracle:thin:@(DESCRIPTION=(ADDRESS=(PROTOCOL=TCPS)(HOST=<HOSTNAME>)(PORT=<PORT>))(CONNECT_DATA=(SID=<SID>))(SECURITY=(SSL_SERVER_CERT_DN=<SSL_SERVER_CERT_DN_VALUE>))) |
+| ssl      | yes                                                                                                                                                                                  |
 
 ### Spill parameters:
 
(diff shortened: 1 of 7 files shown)


athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactory.java
```diff
@@ -60,22 +70,52 @@ public Connection getConnection(final JdbcCredentialProvider jdbcCredentialProvi
...
+                        "\\(port=([1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])\\)\\)" +
+                        "\\(connect_data=\\(sid=[a-zA-Z_]+\\)\\)\\(security=\\(ssl_server_cert_dn=\\\"[=a-zA-Z,0-9-.,]+\\\"\\)\\)\\)");
+                Matcher connStringMatcher = connStringPattern.matcher(derivedJdbcString);
+                if (!connStringMatcher.matches()) {
+                    throw new RuntimeException("Invalid connection string to establish connection over SSL, Please check.");
+                }
+
+                JdbcCredential credential = getCredentialsFromSecretsManager(envVariables.get("secret_name"));
+                this.jdbcProperties.put("user", credential.getUser());
+                this.jdbcProperties.put("password", credential.getPassword());
```
henrymai, line 87: Is there a reason that we can't just reuse the existing mechanism for getting the username/password secret?

If the reason is because of the uri validation, I would suggest to validate only the part after the secret.

Like just split the derivedJdbcString using "@" and then only run the last part part through the pattern matcher.

The pattern must also be modified accordingly to validate the part without the username and password of course.

Then you can continue to make use of the `jdbcCredentialProvider`.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactory.java
```diff
@@ -60,22 +70,52 @@ public Connection getConnection(final JdbcCredentialProvider jdbcCredentialProvi
...
-                Matcher secretMatcher = SECRET_NAME_PATTERN.matcher(databaseConnectionConfig.getJdbcConnectionString());
-                final String secretReplacement = String.format("%s/%s", jdbcCredentialProvider.getCredential().getUser(), jdbcCredentialProvider.getCredential().getPassword());
-                derivedJdbcString = secretMatcher.replaceAll(Matcher.quoteReplacement(secretReplacement));
+            final Map<String, String> envVariables = System.getenv();
+
+            if (OracleConstants.YES.equalsIgnoreCase(envVariables.get("ssl"))) {
+                derivedJdbcString = databaseConnectionConfig.getJdbcConnectionString();
+                Pattern connStringPattern = Pattern.compile("jdbc:oracle:thin:@\\((?i)description=\\(address=\\(protocol=tcps\\)\\(host=[a-zA-Z0-9-.]+\\)" +
+                        "\\(port=([1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])\\)\\)" +
+                        "\\(connect_data=\\(sid=[a-zA-Z_]+\\)\\)\\(security=\\(ssl_server_cert_dn=\\\"[=a-zA-Z,0-9-.,]+\\\"\\)\\)\\)");
```
henrymai, line 79: This should be a `static final SSLConnStringPattern` at the class level.

I know this is just in a lambda so it will be compiled every time the lambda runs anyway, but its just a code quality thing, where we want things that are logically `static final` as a `static final` into the class level.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactory.java
```diff
@@ -60,22 +70,52 @@ public Connection getConnection(final JdbcCredentialProvider jdbcCredentialProvi
...
-            throw new RuntimeException(ex);
-        }
+    }
+
+    public JdbcCredential getCredentialsFromSecretsManager(String secretName)
+    {
+        AWSSecretsManager secretsManager = AWSSecretsManagerClient.builder().build();
+        GetSecretValueResult secretValueResult = secretsManager.getSecretValue(new GetSecretValueRequest()
+                .withSecretId(secretName));
+        return new RdsSecretsCredentialProvider(secretValueResult.getSecretString()).getCredential();
```
henrymai, line 119: This shouldn't be necessary anymore after the change above is made.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleMetadataHandler.java
```diff
@@ -98,7 +95,7 @@ public OracleMetadataHandler()
      */
     public OracleMetadataHandler(final DatabaseConnectionConfig databaseConnectionConfig)
     {
-        this(databaseConnectionConfig, new OracleJdbcConnectionFactory(databaseConnectionConfig, JDBC_PROPERTIES, new DatabaseConnectionInfo(OracleConstants.ORACLE_DRIVER_CLASS, OracleConstants.ORACLE_DEFAULT_PORT)));
+        this(databaseConnectionConfig, new OracleJdbcConnectionFactory(databaseConnectionConfig, null, new DatabaseConnectionInfo(OracleConstants.ORACLE_DRIVER_CLASS, OracleConstants.ORACLE_DEFAULT_PORT)));
```
henrymai, line 98: if this param is always null now, can we just get rid of the param from the `OracleJdbcConnectionFactory` constructor?

athena-oracle/athena-oracle.yaml
```diff
@@ -58,6 +61,8 @@ Resources:
           spill_bucket: !Ref SpillBucket
           spill_prefix: !Ref SpillPrefix
           default: !Ref DefaultConnectionString
+          ssl: !Ref Ssl
+          secret_name: !Ref SecretNamePrefix
```
henrymai, line 65: Shouldn't need this once the code is updated to make use of the existing jdbc credential provider

athena-oracle/README.md
```diff
@@ -137,9 +143,9 @@ We support two ways to input database username and password:
 # Partitions and Splits
 A partition is represented by a single partition column of type varchar. We leverage partitions defined on a Oracle table, and this column contains partition names. For a table that does not have partition names, * is returned which is equivalent to a single partition. A partition is equivalent to a split.
 
-|Name|Type|Description
-|---|---|---|
-|PARTITION_NAME|Varchar|Named partition in MySql. E.g. P_2006_DEC,P_2006_AUG ..|
+| Name           | Type    | Description                                              |
+|----------------|---------|----------------------------------------------------------|
+| PARTITION_NAME | Varchar | Named partition in Oracle. E.g. P_2006_DEC,P_2006_AUG .. |
```
henrymai, line 148: Good catch

trianz-raviboyina:
Henry, 
I have addressed all the review comments, below are the connection strings for your reference.

SSL: yes
Connection String: 
oracle://jdbc:oracle:thin:${<SECRET_NAME>}@(DESCRIPTION=(ADDRESS=(PROTOCOL=TCPS)(HOST=<HOST_NAME>)(PORT=<PORT_NO>))(CONNECT_DATA=(SID=<SID_NAME>))(SECURITY=(SSL_SERVER_CERT_DN="<PARAMETERS>")))

SSL: no
Connection String:
oracle://jdbc:oracle:thin:${<SECRET_NAME>}@//<HOST_NAME>:<PORT_NO>/<SID_NAME>

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactory.java
trianz-raviboyina: yes Henry,

we used to provide secret name along with connection string.
I see below SSL based connection string from [oracle documentation](https://www.oracle.com/technetwork/database/enterprise-edition/wp-oracle-jdbc-thin-ssl-130128.pdf)  which does not have credentials as part of url.
`String url = "jdbc:oracle:thin:@(DESCRIPTION=(ADDRESS=(PROTOCOL=tcps)(HOST=servername)(PORT=2484))(CONNECT_DATA=(SERVICE_NAME=servicename)))";`

As the earlier logic has failed in SSL scenario, I have added utility method which will fetch credentials from aws secret manager.

Today I have tried sample java program which uses SSL based connection string along with credentials as part of the url. It worked fine, now I've reused the existing mechanism as suggested.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactory.java
```diff
@@ -25,57 +25,78 @@
...
+                    }
+                    properties.put("javax.net.ssl.trustStoreType", "JKS");
+                    properties.put("javax.net.ssl.trustStorePassword", "changeit");
+                    properties.put("oracle.net.ssl_server_dn_match", "true");
+
+                    LOGGER.info("Establishing connection over SSL..");
+                }
+                else {
+                    if (connStringMatcher.matches()) {
+                        throw new RuntimeException("Invalid connection string, Please check.");
```
henrymai, line 83: Actually, this makes me wonder if we need the ssl environment variable at all?

I think that since we're already pattern matching against the connection string, we can infer ssl, right?

The main reason I'd want to get rid of the extra environment variable is because the way we support SSL in the future may change if we extend beyond RDS and it may be a pain for backwards compatibility later on.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactory.java
henrymai: Great thanks for checking and updating the patch.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactory.java
trianz-raviboyina: Yes Henry, got it.

Done with the required changes, please check once.

henrymai (approved):
Looks good.

Issue #759, opened 2022-08-16
[BUG/FEATURE] Case Sensitive and Date/timestamp in Neptune Athena Connector


jinman:
**Is your feature request related to a problem? If yes, please describe.**
We have several nodes and properties that are camel case and several date timestamps as datatypes. All the columntypes appear as blank and we are unable to consume the Athena neptune connector.

**Describe the solution you'd like**
Amazon DynamoDB provides casing flag which allows customers to query case sensitive column names for Quicksight consumption

**Describe alternatives you've considered**
Changing column names and datatypes is an option but typically not possible

**Additional context**
Abhishek has developed https://github.com/abhishekpradeepmishra/aws-athena-query-federation/tree/neptune-edge-support is there any chance we can merge this request as this will truly help us tremenduously.

jinman:
Is there an update to move this to main branch/public serverless repo as we are unable to build - lots of test failures.

henrymai:
This has been merged: https://github.com/awslabs/aws-athena-query-federation/commit/8fcad1b6d1014392fcde7158d22456035be4ac6f

Closing this out now.

PR #761, opened 2022-08-22
Neptune edge support

abhishekpradeepmishra:
*Issue #, if available:*

https://github.com/awslabs/aws-athena-query-federation/issues/759

*Description of changes:*

Added support for Neptune date field type and Case Insensitive search (defaults to true)


avirtuos:
LGTM but would like @burhan94's take before we merge.

avirtuos (approved):
approved but would like to get @burhan94  's take before merging.

athena-neptune/athena-neptune.yaml
```diff
@@ -65,6 +65,10 @@ Parameters:
     Description: 'If set to ''true'' the connector uses Signature Version 4 Signing'
     Default: false
     Type: String
+  EnableCaseInsensitiveMatch:
+    Description: 'If set to ''false'' the connector does a case sensetive match for keys'
```
burhan94, line 69: nit: sensetive -> sensitive

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/EdgeRowWriter.java
```diff
@@ -77,33 +82,33 @@ public static void writeRowTemplate(RowWriterBuilder rowWriterBuilder, Field fie
                 rowWriterBuilder.withExtractor(field.getName(),
                         (VarCharExtractor) (Object context, NullableVarCharHolder value) -> {
                             Map<Object, Object> obj = (Map<Object, Object>) context;
+                            String fieldName = getFieldName(field.getName().toLowerCase().trim(), obj);
```
burhan94, line 85: maybe I'm missing something, but how does a case sensitive match happen if I'm always lower casing the fieldName? Can you add some unit tests to show how this works?

henrymai: Yup, this is error prone.

I would suggest that the author of this patch look into DDBTypeUtils.java:contextAsMap and also see how it gets used and follow that pattern instead:
https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTypeUtils.java#L368

abhishekpradeepmishra: The challenge here is to pick the correct key from underlying data store because Glue converts everything into lower case. So whether you store FirstName, FiRstName or firstname Glue always sends FirstName. I need to pick correct key match from Neptune properties. The final match of properties anyways has to be converted into Lowercase

abhishekpradeepmishra: If you know a way to read properties from Glue Table the way it was stored let me know, I couldn't find any reference for it.

henrymai: This is error prone because it requires every site that performs a lookup into the obj map:
1) Cast `Object obj` into `Map<Object, Object> map`.

2) To lowercase the `fieldName` field before passing it into this function.

3) Pass in the resolved `fieldName` back into the `map` to look up the value.

If you look at the `DDBTypeUtils.java:contextAsMap` way of doing things instead, the map itself becomes case insensitive so you never have to worry about the calling sites lowercasing before passing in.

Using the `DDBTypeUtils.java:contextAsMap` pattern it becomes:

1) Convert `Object obj` into `Map<String, Object> map`  using `contextAsMap(obj, caseInsensitive)`. This returns you a Map that is either case sensitive or not depending on the parameter.

2) Simply lookup `fieldName` from `map` without lowercasing or anything.

This means that if you did things like `DDBTypeUtils.java:contextAsMap` it just works without the developer ever having to think about manually lowecasing `fieldName` when `enableCaseinsensitivematch` is set to true (this just replaces the straight cast with a function call instead of the cast).

Think about this from the aspect of maintainability too, people that add code to this function won't have to think about lowercasing the `fieldName` before looking things up if you just had a case insensitive map to begin with.

The pattern using `DDBTypeUtils.java:contextAsMap` is the most direct expression of "case insensitive lookup" rather than what is currently being done in this pull request, which is: "resolve the original cased key from a lowercased key by searching the map keys and then look up the value from the map using the resolved original cased key."

So even while your assumption that the fieldNames are always going to be lowercased is correct, it is strictly simpler and less error prone to never have to make that assumption or think about it to begin with.

abhishekpradeepmishra: Makes sense, so if I understand it correctly. A call to obj.get  would return value irrespective of key being uppercase or lowercase or mixedcase. Will make changes and test

henrymai: Yeah, that's the idea.

abhishekpradeepmishra: updated code to use contextAsMap.

Neptune doesn't return simple key value map. For keys ID, IN and OUT the returned key is of type struct

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/EdgeRowWriter.java
```diff
@@ -76,34 +82,34 @@ public static void writeRowTemplate(RowWriterBuilder rowWriterBuilder, Field fie
...
+                            } 
                             else if (fieldName.equals(SpecialKeys.OUT.toString().toLowerCase())) {
-                                Object fieldValue = ((LinkedHashMap) obj.get(Direction.OUT)).get(T.id);
+                                Object fieldValue = obj.get(SpecialKeys.OUT.toString());
                                 if (fieldValue != null) {
                                     value.value = fieldValue.toString();
                                     value.isSet = 1;
                                 }
                             } 
                             else {
```
henrymai, line 111: It looks like all of these `if` and `else if` statements can be deleted now since they all just do the same thing as the `else` case (lines 112-116), right?

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/EdgeRowWriter.java
```diff
@@ -167,4 +190,32 @@ else if (fieldName.equals(SpecialKeys.OUT.toString().toLowerCase())) {
...
+            Object fieldValueOUT = ((LinkedHashMap) contextAsMap.get(Direction.OUT)).get(T.id);
+
+            contextAsMap.remove(T.id);
+            contextAsMap.remove(T.label);
+            contextAsMap.remove(Direction.IN);
+            contextAsMap.remove(Direction.OUT);
+
+            contextAsMap.put(SpecialKeys.ID.toString(), fieldValueID);
+            contextAsMap.put(SpecialKeys.IN.toString(), fieldValueIN);
+            contextAsMap.put(SpecialKeys.OUT.toString(), fieldValueOUT);
```
henrymai, line 214: Just checking, is it safe to mutate `Object context` here?

Meaning, is this object ever used beyond the scope of just the lambda where `contextAsMap` is called?

If so, you may want to perform these mutations on the TreeMap copy of this object instead.

I'm not sure though, that's why I'm just asking to double check.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/VertexRowWriter.java
```diff
@@ -157,4 +182,24 @@ public static void writeRowTemplate(RowWriterBuilder rowWriterBuilder, Field fie
...
+
+        if (!caseInsensitive) {
+            return contextAsMap;
+        }
+
+        Object fieldValueID = contextAsMap.get(T.id);
+        if (fieldValueID != null) {
+            contextAsMap.remove(T.id);
+            contextAsMap.remove(T.label);
+            contextAsMap.put(SpecialKeys.ID.toString(), fieldValueID);
```
henrymai, line 198: Same question here as above in regards to mutating `Object context`

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/EdgeRowWriter.java
abhishekpradeepmishra: Yes, seems like so, let me make changes and test

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/EdgeRowWriter.java
abhishekpradeepmishra: The tree map copy fails because, the keys are not String in the original context map. The first call to the context map per record, transforms the records else it will keep failing. And then the next call just use the modified context map. But this is limited to per record processing

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/VertexRowWriter.java
abhishekpradeepmishra: same as above, my initial thought was it would be straight forward but it fails because the keys are not Strings

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/EdgeRowWriter.java
henrymai: Sounds fine to me. Just wanted to double check.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/EdgeRowWriter.java
henrymai: Actually one thing you may want to do is perform the flattening in `contextAsMap` even when `caseInsensitive` is false to maintain the correct behavior regardless of casing.

abhishekpradeepmishra: updated

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/PropertyGraphHandler.java
```diff
@@ -109,9 +109,15 @@ public void executeQuery(ReadRecordsRequest recordsRequest, final QueryStatusChe
...
         GeneratedRowWriter.RowWriterBuilder builder = GeneratedRowWriter.newBuilder(recordsRequest.getConstraints());
         String type = recordsRequest.getSchema().getCustomMetadata().get("componenttype");
+        String glabel = recordsRequest.getSchema().getCustomMetadata().get("glabel");
         TableSchemaMetaType tableSchemaMetaType = TableSchemaMetaType.valueOf(type.toUpperCase());
-        
+
         logger.debug("readWithConstraint: schema type is " + tableSchemaMetaType.toString());
+        
+        //AWS Glue converts table name to lowercase, table property 'glabel' stores Amazon Neptune Vertex/Edge labels to be used in Gremlin query
+        if (glabel != null && !glabel.trim().isEmpty()) {
```
abhishekpradeepmishra, line 118: added this to  handle AWS Glue converting table name to lowercase, the gremlin query fails as Property Graph label names are case sensitive

henrymai: Just making sure: this is something that a customer has to set on their Glue Table, right?

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/EdgeRowWriter.java
henrymai: No, I mean that the deleted `if else` statements were basically doing an ad-hoc flattening of the map.

This isn't being done anymore when `caseInsensitive` is set to `false` (it doesn't matter what the default is).

So in order to maintain the same behavior as these deleted `if else` statements, regardless of the value of `caseInsensitive`, `contextAsMap` should always "flatten" out the map.

henrymai: Additionally, did you test to make sure that when `caseInsensitive` is set to false, that everything works as expected?

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/PropertyGraphHandler.java
abhishekpradeepmishra: Yes, I have updated sample Nodejs script on how to do it

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/EdgeRowWriter.java
abhishekpradeepmishra: thanks for pointing this, refactored code to make sure contextMap rewrite is consistent whether flag is set to true or false

henrymai (approved):
Looks good. Approved.

nauy2697 (approved):

PR #777, opened 2022-09-12
Upgrade apache arrow version to 9.0.0

henrymai:
Includes upgrades to jackson and netty as well (required for the arrow upgrade).


burhan94:
I don't expect there to be any issues, but please double check that we're not breaking backwards compatibility. 

When I did this last time, I checked the following:
1. Engine contains newer sdk, customer lambda has older
2. Engine contains older sdk, customer lambda has newer

Arrow had breaking changes the last time it was upgraded (.14 to 3.0.0), but it looks like its been pretty stable since then.

hackett123 (approved):
all looks good.

PR #779, opened 2022-09-16
Bugfix/snowflake partitionlimit

ankita-yadav12:
*Issue # INC00001080,Snowflake Connector Partitioning Logic Is Resulting In A Single Large Partition, Slowing Down Queries.

*Description of changes:*
We have removed pagecount and partitionlimit variables from yaml file(environment variables).It is handled internally to make it user friendly and for better performance.
Maximum partition limit we have limited to 50 to handle timeout or slow performance issues because of huge partitions.
If the limit of partitions will increase beyond 50 partitions,it will handle internally to make maximum 50 partitions now(which was single partition before).

**Content needs to be removed from Readme:**
PageCount parameter
Limits the number of records for each partition. The default value is 500000

PartitionLimit parameter
Limits the number of partitions. The default value is 10. A large number may cause a timeout issue. If you encounter a timeout error, set this parameter to a lower value.


akuzin1:
For SnowflakeMetadataHandlerTest.java, I think it would make sense to alter the doGetTableLayoutMaxPartition() test to not have hardcoded values, and instead use a similar approach to when creating expectedValues. That way the test can be more dynamic if looking to test for a different total record count.

akuzin1 (changes requested):
For **SnowflakeMetadataHandlerTest.java**, I think it would make sense to alter the **doGetTableLayoutMaxPartition()** test to not have hardcoded **_values_**, and instead use a similar approach to when creating **_expectedValues_**. That way the test can be more dynamic if looking to test for a different total record count.

PR #780, opened 2022-09-21
Data Integrity: Add generative testing and fixes

ykrystal:
Dependency changes:
  - Upgraded JUnit to Jupiter
  - Added JQwik

Added BlockUtilsPropertiesTest

Found and fixed the following issues with BlockUtils (via the added test above):
  - Null values not being preserved in any container types (Lists, Maps, Structs)
  - List of maps was not supported (threw an exception)
  - Struct of maps was not supported (thew an exception)
  - Incorrect casting of primitive types (thew an exception)
  - Struct of DateMilli fields was not working (threw an exception)

Remaining issues:
  - List and Map of TimeStamp*TZ fields are not working right now because of a bug in Arrow. - This is an existing problem. - Test is not generating and testing these fields right now (commented out).

### DIFF
--- a/athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/data/BlockUtilsPropertiesTest.java
+++ b/athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/data/BlockUtilsPropertiesTest.java
@@ -60,13 +60,19 @@ class BlockUtilsPropertiesTest {
         return fieldsGenerator.field();
     }
 
+    // Using the default number of tries here (1000)
+    // With a higher recursion level, we will not be able to cover all possible combinations
+    // This will do a random sampling of scenarios
+    // The purpose of this test is to cover depth
     @Property
-    boolean setComplexValuesSetsAllFieldsCorrectlyGivenAnyInputShort(@ForAll("fieldHighRecursion") Field field) {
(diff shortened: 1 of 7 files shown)


henrymai (approved):
Looks great to me. But lets also get one more person to review.

athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/lambda/data/BlockUtilsPropertiesTest.java
```diff
@@ -0,0 +1,188 @@
...
+        return fieldsGenerator.field();
+    }
+
+    @Provide
+    private Arbitrary<Field> fieldHighRecursion() {
+        FieldsGenerator fieldsGenerator = new FieldsGenerator(5);
+        return fieldsGenerator.field();
+    }
+
+    @Property
```
hackett123, line 63: Curious why we are only running this one once, and the other one 2000 times.

ykrystal: The default is actually 1000. Since the recursion depth is higher here, it is not possible to cover all the possible combinations so we're using the default number of tries to do a sampling of combinations (whereas the other test with a lower recursion will cover all combinations but cover less depth). I will add a comment for this.

Issue #491, opened 2021-09-30
[BUG] Limit not working with connector ElasticSearch


BhuviTheDataGuy:
I have deployed the connector and ran some queries, it was working fine. Then I started scanning a large index (100k records). Then I got an error like this.

```
GENERIC_USER_ERROR: Encountered an exception[org.elasticsearch.ElasticsearchStatusException] from your 
LambdaFunction[arn:aws:lambda:ap-south-1:123123123:function:esconnector] 
executed in context[S3SpillLocation{bucket='dev-de-team',
key='athena-spill/06a182de-4450-456d-924e-4217d7653198/72899ad3-8277-4e14-a70b-db460405f0cd', directory=true}] 
with message[Elasticsearch exception [type=search_phase_execution_exception, reason=all shards failed]]
```

I checked the cloudwatch log and got this.

```
Elasticsearch exception [type=search_phase_execution_exception, reason=all shards failed]:
org.elasticsearch.ElasticsearchStatusException
ElasticsearchStatusException[Elasticsearch exception [type=search_phase_execution_exception, reason=all shards failed]];
nested: ElasticsearchException[Elasticsearch exception [type=illegal_argument_exception, reason=Result window is too large,
from + size must be less than or equal to: [10000] but was [10100]. See the scroll api for a more efficient way to request large data
sets. This limit can be set by changing the [index.max_result_window] index level setting.]]; nested:
ElasticsearchException[Elasticsearch exception [type=illegal_argument_exception, reason=Result window is too large, from + size
must be less than or equal to: [10000] but was [10100]. See the scroll api for a more efficient way to request large data sets. This
limit can be set by changing the [index.max_result_window] index level setting.]];
```

I tried to use `limit 100` in SQL, but no luck.

burhan94:
Hi @BhuviTheDataGuy 

So the connectors dont support limit pushdown at the moment. It's something we're actively working on and we are planning on releasing it during the next Athena Engine upgrade :)

nauy2697:
Hi,
Our ES connectors just got an update to process the data more the default max window size, which will resolved below exception.
`Elasticsearch exception [type=search_phase_execution_exception, reason=all shards failed]:
org.elasticsearch.ElasticsearchStatusException
ElasticsearchStatusException[Elasticsearch exception [type=search_phase_execution_exception, reason=all shards failed]];
nested: ElasticsearchException[Elasticsearch exception [type=illegal_argument_exception, reason=Result window is too large,
[...]

Issue #461, opened 2021-07-22
[BUG] error with large number of records with connector to Elasticsearch


madhavvishnubhatta:
**Describe the bug**
When I query an index with a large number of rows (roughly greater than about 50,000), I get an error:

```
GENERIC_USER_ERROR: Encountered an exception[org.elasticsearch.ElasticsearchStatusException] from your LambdaFunction[es-connector] executed in context[S3SpillLocation{bucket='<spill-bucket-name>', key='athena-spill/6519b7e0-eaf3-45ce-849b-a2c3d8acd9e3/5d3c9931-bd11-4a23-9d64-04b2f1556a89', directory=true}] with message[Elasticsearch exception [type=search_phase_execution_exception, reason=all shards failed]]
```

**To Reproduce**
1. Create an index in Elasticsearch and load more than 50,000 records (or more, I am not sure what the threshold is)
2. Count the records using the following query on Kibana Dev Tools
```
POST _opendistro/_sql
{
  "query":""" select count(*) from testindex"""
}
```

The result will be like this
```json
{
  "schema": [
    {
      "name": "count(*)",
      "type": "integer"
    }
  ],
  "datarows": [
    [
      161564
    ]
  ],
  "total": 1,
  "size": 1,
  "status": 200
}
```
3. Now deploy the connector application and query the same index from Athena

```sql
select count(*) from "lambda:es-connector"."es-alleninstitute-data"."cord19"
```

We get the error
```
GENERIC_USER_ERROR: Encountered an exception[org.elasticsearch.ElasticsearchStatusException] from your LambdaFunction[es-connector] executed in context[S3SpillLocation{bucket='<spill-bucket-name>', key='athena-spill/6519b7e0-eaf3-45ce-849b-a2c3d8acd9e3/5d3c9931-bd11-4a23-9d64-04b2f1556a89', directory=true}] with message[Elasticsearch exception [type=search_phase_execution_exception, reason=all shards failed]]
```

**Expected behavior**
The returns the proper count.

**Screenshots / Exceptions / Errors**
N/A

**Connector Details (please complete the following information):**
 - Version: 2021.27.1
 - Name elasticsearch
 - Athena Query IDs: 6519b7e0-eaf3-45ce-849b-a2c3d8acd9e3.

**Additional context**
The exact query works fine if the number of records is about 47K

avirtuos:
What happens if you run this query on elasticsearch without using Athena? e.g. using some other elastic client?

[...]

PR #783, opened 2022-09-22
Elastic search connector enable pagination query data from get records

nauy2697:
Elastic search connector enable pagination query data from get records. This will allow use to bypass max_result_window settings.

By default, max_result_window set to 10,000. We can't query data more than 10,000. Enable pagination we can read all rows.

Test done through integration test and manually test on sample cluster with max_result_window = 1,000 and data rows 4907

https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html

Link to related tickets:
https://github.com/awslabs/aws-athena-query-federation/issues/461
https://github.com/awslabs/aws-athena-query-federation/issues/491

### DIFF
--- a/athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchRecordHandler.java
+++ b/athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchRecordHandler.java
@@ -34,9 +34,14 @@ import com.amazonaws.services.secretsmanager.AWSSecretsManager;
 import com.amazonaws.services.secretsmanager.AWSSecretsManagerClientBuilder;
 import org.apache.arrow.util.VisibleForTesting;
 import org.apache.arrow.vector.types.pojo.Field;
+import org.elasticsearch.action.search.ClearScrollRequest;
 import org.elasticsearch.action.search.SearchRequest;
 import org.elasticsearch.action.search.SearchResponse;
+import org.elasticsearch.action.search.SearchScrollRequest;
+import org.elasticsearch.client.RequestOptions;
+import org.elasticsearch.common.Strings;
 import org.elasticsearch.common.unit.TimeValue;
+import org.elasticsearch.search.Scroll;
 import org.elasticsearch.search.SearchHit;
 import org.elasticsearch.search.builder.SearchSourceBuilder;
 import org.slf4j.Logger;
@@ -68,7 +73,11 @@ public class ElasticsearchRecordHandler
 
     // Env. variable that holds the query timeout period for the Search queries.
     private static final String QUERY_TIMEOUT_SEARCH = "query_timeout_search";
+    // Env. variable that holds the scroll timeout for the Search queries.
+    private static final String SCROLL_TIMEOUT = "query_scroll_timeout";
+
     private final long queryTimeout;
+    private final long scrollTimeout;
 
     // Pagination batch size (100 documents).
     private static final int QUERY_BATCH_SIZE = 100;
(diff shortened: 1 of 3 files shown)


athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchRecordHandler.java
```diff
@@ -147,38 +151,54 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
+                Scroll scroll = new Scroll(TimeValue.timeValueSeconds(60));
+
+                //Read the returned scroll id, which points to the search context that’s being kept alive and will be needed in the following search scroll call
+                SearchResponse searchResponse = client.search(searchRequest, RequestOptions.DEFAULT);
+                String scrollId = searchResponse.getScrollId();
+                SearchHit[] searchHits = searchResponse.getHits().getHits();
+                Iterator<SearchHit> iterator = searchResponse.getHits().iterator();
+
+                while (searchHits != null && searchHits.length > 0 && queryStatusChecker.isQueryRunning()) {
+                    Iterator<SearchHit> finalIterator = iterator;
```
henrymai, line 176: Is this variable necessary? It looks like you can just use `iterator` as is without assigning it to `finalIterator` right?

henrymai: Also I don't think you need to even initialize `iterator` outside of the while loop. You can just do it inside where you're assigning to `finalIterator`, but instead you can just do:

`Iterator<SearchHit> iterator = searchResponse.getHits().iterator();`
on line 176 and then get rid of line 196 where you assign to it again.

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchRecordHandler.java
```diff
@@ -147,38 +151,54 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
-                do {
-                    // Process the search request injecting the search-source, and setting the from position
-                    // used for pagination of results.
-                    SearchResponse searchResponse = client
-                            .getDocuments(searchRequest.source(searchSource.from(currPosition)));
-
-                    // Throw on query timeout.
+                SearchRequest searchRequest = new SearchRequest(index)
+                        .preference(shard)
+                        .scroll(TimeValue.timeValueSeconds(60))
```
hackett123, line 163: This means the ElasticSearch search request will stop maintaining context of the state of our search after 60 seconds (docs say `The scroll parameter indicates how long Elasticsearch should retain the [search context](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#scroll-search-context) for the request.` [link](https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html#scroll-api)). What is the expected behavior if a query to ES takes more than a minute? Have we tested this with very large ES indices to verify we don't approach this limit in our handlers?

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchRecordHandler.java
henrymai: Same comment about `scrollId`, you can just move line 171 into the while loop body and then get rid of line 194.

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchRecordHandler.java
```diff
@@ -147,38 +151,54 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
-                    while (hitIterator.hasNext() && queryStatusChecker.isQueryRunning()) {
-                        ++numRows;
-                        spiller.writeRows((Block block, int rowNum) ->
-                                rowWriter.writeRow(block, rowNum, client.getDocument(hitIterator.next())) ? 1 : 0);
+                    if (searchResponse.getHits() == null) {
+                        break;
                     }
-                    // if hitsNum < QUERY_BATCH_SIZE, then this is the last batch of documents.
-                } while (hitsNum == QUERY_BATCH_SIZE && queryStatusChecker.isQueryRunning());
+                    scrollId = searchResponse.getScrollId();
```
hackett123, line 194: wondering if we should put this at the top of the loop and remove the assignments that occur before the loop. That way we don't repeat ourselves and risk a bug introduced from the copy. The only thing is that you then need to break up the while condition into a break clause.

style wise, deferring to your judgment, just raising the question

henrymai (approved):

PR #787, opened 2022-09-28
DDB: Add BETWEEN support

henrymai:
This patch converts inclusive ranges for DDB key conditions to use BETWEEN.

Also tests have been added to ensure that we do not convert the mixed inclusion ranges to BETWEEN.

Also fixes a long standing buggy test that has been masked over by
automatic type inference failover in: DynamoDBRecordHandlerTest.java

This patch is a safer version of this prior work.: https://github.com/awslabs/aws-athena-query-federation/pull/375


kparwal (approved):
looks good to me

PR #791, opened 2022-10-06
Data Integrity: Add generative tests for FieldResolver, DynamoDBFieldResolver, ElasticsearchFieldResolver.

ykrystal:
Added FiedResolverTest, DDBFieldResolverTest and ElasticsearchFieldResolverGenerativeTest

Found and fixed the following issues in FieldResolver:
1) Values not being set correctly for maps (throws exception)

Found and fixed the following issues in DynamoDBFieldResolver and ElasticsearchFieldResolver: 
1) Null values not being set correctly for nested lists (eg. Lists of lists) (incorrect behavior where a null list would be represented as a list with a null value) 
2) Values not being set correctly for maps (throws exceptiton)

### DIFF
--- a/athena-dynamodb/pom.xml
+++ b/athena-dynamodb/pom.xml
@@ -88,6 +88,13 @@
             <version>${aws-cdk.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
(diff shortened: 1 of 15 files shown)


henrymai (approved):
Looks good to me.

PR #794, opened 2022-10-11
Bugfix/snowflake partitionlimit issue

ankita-yadav12:
*Issue # INC00001080,Snowflake Connector Partitioning Logic Is Resulting In A Single Large Partition, Slowing Down Queries.

### Description of changes:
We have removed pagecount and partitionlimit variables from yaml file(environment variables).It is handled internally to make it user friendly and for better performance.
Maximum partition limit we have limited to 50 to handle timeout or slow performance issues because of huge partitions.
If the limit of partitions will increase beyond 50 partitions,it will handle internally to make maximum 50 partitions now(which was single partition before).

### Content needs to be removed from Readme:
PageCount parameter
Limits the number of records for each partition. The default value is 500000

PartitionLimit parameter
Limits the number of partitions. The default value is 10. A large number may cause a timeout issue. If you encounter a timeout error, set this parameter to a lower value.

### DIFF
--- a/athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeConstants.java
+++ b/athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeConstants.java
@@ -7,9 +7,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -25,6 +25,8 @@ public final class SnowflakeConstants
     public static final String SNOWFLAKE_NAME = "snowflake";
     public static final String SNOWFLAKE_DRIVER_CLASS = "com.snowflake.client.jdbc.SnowflakeDriver";
     public static final int SNOWFLAKE_DEFAULT_PORT = 1025;
+    public static final int PARTITION_RECORD_COUNT = 500000;
+    public static final int MAX_PARTITION_COUNT = 50;
 
     private SnowflakeConstants() {}
 }
--- a/athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeConstants.java
+++ b/athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeConstants.java
@@ -25,7 +25,15 @@ public final class SnowflakeConstants
     public static final String SNOWFLAKE_NAME = "snowflake";
     public static final String SNOWFLAKE_DRIVER_CLASS = "com.snowflake.client.jdbc.SnowflakeDriver";
     public static final int SNOWFLAKE_DEFAULT_PORT = 1025;
+    /**
+     * This constant limits the number of records for each partition. The default set to 500000
+     * We arrived at this number after performance testing with datasets of different size
+     */
     public static final int PARTITION_RECORD_COUNT = 500000;
+    /**
+     * This constant limits the number of partitions. The default set to 50. A large number may cause a timeout issue.
+     * We arrived at this number after performance testing with datasets of different size
+     */
     public static final int MAX_PARTITION_COUNT = 50;
 
     private SnowflakeConstants() {}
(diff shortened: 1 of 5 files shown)


athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeConstants.java
```diff
@@ -25,6 +25,8 @@
     public static final String SNOWFLAKE_NAME = "snowflake";
     public static final String SNOWFLAKE_DRIVER_CLASS = "com.snowflake.client.jdbc.SnowflakeDriver";
     public static final int SNOWFLAKE_DEFAULT_PORT = 1025;
+    public static final int PARTITION_RECORD_COUNT = 500000;
+    public static final int MAX_PARTITION_COUNT = 50;
```
kparwal, line 29: I may have asked this before, but did we try this with a higher default? It's pretty common to have more than 25M records in a table.

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -151,24 +151,15 @@ public void getPartitions(BlockWriter blockWriter, GetTableLayoutRequest getTabl
...
          */
-        String partitionlimit = properties.get("partitionlimit");
-        Long totalPartitionlimit = Long.valueOf(partitionlimit);
-
-        LOGGER.info(" Total Partition Limit" + totalPartitionlimit);
-        LOGGER.info(" Total Page  Count" +  totalpagecount);
+        LOGGER.info(" Total Partition Limit" + SnowflakeConstants.MAX_PARTITION_COUNT);
+        LOGGER.info(" Total Page  Count" +  SnowflakeConstants.PARTITION_RECORD_COUNT);
         long offset = 0;
         double limit = 0;
```
kparwal, line 164: declare variables in the scope that they're being used.

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -191,30 +182,41 @@ public void getPartitions(BlockWriter blockWriter, GetTableLayoutRequest getTabl
...
-                        {
-                            block.setValue(BLOCK_PARTITION_COLUMN_NAME, rowNum, partitionVal);
-                            return 1;
-                        });
+                    // if number of partitions are more than defined limit "MAX_PARTITION_COUNT"
+                    // it will do maximum 50 partitions,49 partitions will have 500000 records each and last partition will have the remaining number of records.
+                    if (limit > SnowflakeConstants.MAX_PARTITION_COUNT) {
+                        for (int i = 1; i <= SnowflakeConstants.MAX_PARTITION_COUNT; i++) {
+                            int partitionRecord = SnowflakeConstants.PARTITION_RECORD_COUNT;
+                            if (i > 1) {
```
kparwal, line 193: You could probably set `offset` after it's used and remove this condition

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -191,30 +182,41 @@ public void getPartitions(BlockWriter blockWriter, GetTableLayoutRequest getTabl
...
                     else {
                         /**
                          * Custom pagination based partition logic will be applied with limit and offset clauses.
-                         * the partition values we are setting the limit and offste values like p-limit-3000-offset-0
+                         * the partition values we are setting the limit and offset values like p-limit-3000-offset-0
                          */
                         for (int i = 1; i <= limit; i++) {
                             if (i > 1) {
-                                offset = offset + totalpagecount;
+                                offset = offset + SnowflakeConstants.PARTITION_RECORD_COUNT;
```
kparwal, line 217: You can just static import these, no need to prefix them all the time.

kparwal (changes requested):
Needs some code cleanup and polish. Also, please try testing other numbers for these defaults or provide a justification for them. Lastly, document the reasoning for this specific partition/split strategy in the code so that future readers can understand why this is being done.

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
ankita-yadav12: Hi Keshav,I have addressed all the review comments.Please review

kparwal (approved):

PR #808, opened 2022-10-21
Add support for optional LambdaRole parameter for AthenaSynapseConnector

ankita-yadav12:
*Issue # INC00001084, Being able to set the Lambda Role in AthenaSynapseConnector.

*Description of changes:*
Added support for optional LambdaRole parameter for AthenaSynapseConnector. The Lambda function created will now allow users to select own role.

### DIFF
--- a/athena-synapse/athena-synapse.yaml
+++ b/athena-synapse/athena-synapse.yaml
@@ -40,11 +40,11 @@ Parameters:
     Description: 'Lambda memory in MB (min 128 - 3008 max).'
     Default: 3008
     Type: Number
-  LambdaRole:
+  LambdaRoleARN:
     Description: "(Optional) A custom role to be used by the Connector lambda"
     Default: ""
     Type: String
-  PermissionBoundary:
+  PermissionsBoundaryARN:
     Description: "(Optional) A custom Permission Boundary to be used by the Connector lambda"
     Default: ""
     Type: String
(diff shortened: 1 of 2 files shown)


henrymai:
Hi Ankita,

Can you look into adding support for setting an optional `PermissionsBoundary`: https://aws.amazon.com/premiumsupport/knowledge-center/lambda-sam-template-permissions/

It will need to be implemented in a way such that only one or the other (`PermissionsBoundary` or `Role`) is set in the `Properties` (both should not be set at the same time).

Thanks.

athena-synapse/athena-synapse.yaml
```diff
@@ -40,6 +40,14 @@ Parameters:
     Description: 'Lambda memory in MB (min 128 - 3008 max).'
     Default: 3008
     Type: Number
+  LambdaRole:
+    Description: "(Optional) A custom role to be used by the Connector lambda"
+    Default: ""
+    Type: String
+  PermissionBoundary:
```
henrymai, line 47: One last nit:
Is it possible to name this `PermissionsBoundary` to be consistent with how this is named normally?

If not (because the name conflicts), can you name it `PermissionsBoundaryARN`?

Similarly, if you end up naming it `PermissionsBoundaryARN` can you rename `LambdaRole` to `LambdaRoleARN` ?

Sorry for being nitpicky about this.

ankita-yadav12: Yes Henry, I will check on this.

henrymai (approved):
Looks good. Thanks.

PR #811, opened 2022-10-23
Don't catch just to rethrow

henrymai:
There are several instances in the code where we caught an exception just to rethrow it as a RuntimeException.

Sometimes this was due to the interfaces not declaring that they throw, so the implementations were forced to catch exceptions and then rethrow them as RuntimeExceptions.

This patch fixes the interfaces and implementations that inherit from those interfaces to not catch just to rethrow.


henrymai:
Note to reviewers, make sure to turn on "Hide whitespace", otherwise the diff will show excessive changes when not much has changed.
diff_settings

ykrystal (approved):

Issue #184, opened 2020-05-29
[BUG] Queries fail on tables with array-of-struct type with DynamoDB Connector


SeanPrendi:
**Describe the bug**
It seems like the DynamoDB connector breaks when it encounters a field with the array-of-struct type. This seems to be intimately connected with the issue that I was facing in #182. We currently managed to bypass this issue to an extent by using DDL to define copies of the offending tables with the array-of-struct fields ignored. Interestingly, based on CloudWatch logs it seems like running the connector on tables that have columns with this type is fine by itself as they're not parsed as array-of-struct, but instead list-of-struct, which works. However, the glue schema inference marks these fields as being of type array-of-struct, which breaks the connector. The connector then falls back to its inference, which brings us back to the null reference exception that was the original reason for using the glue connector. 

**To Reproduce**
Create a table in DynamoDB with an array of structs
Crawl the table in glue
Query the glue table with the connector

**Expected behavior**
The query should be executed successfully

**Screenshots / Exceptions / Errors**
First the connector encounters the array-of-struct type, and CloudWatch logs:

>INFO  GlueMetadataHandler:360 - Column [offending column] with registered type array<struct<[fields]>>

Then fails and falls back to the Connector's inference schema

>doGetTable: Unable to retrieve table [table] from AWSGlue in database/schema [database]. Falling back to schema inference

However, when running a table with similar data (but no empty values, so inference can be performed successfully by the dynamo connector), we see 

>INFO  RecordHandler:154 - doHandleRequest: request[ReadRecordsRequest{queryId=[query id], tableName=TableName{schemaName=[schema], tableName=[table]}, schema=Schema<..., [field]: List<[field].element: Struct<[struct fields]>>, ...>, ...]

**Connector Details (please complete the following information):**
 - Version: 2020.19.1
 - Name DynamoDB

avirtuos:
@atennak1 and @soojinj any thoughts on this? I think you are most familiar with this code path.

atennak1:
Was there a stacktrace printed with `doGetTable: Unable to retrieve table [table] from AWSGlue in database/schema [database]. Falling back to schema inference`?  If so could you provide it?

SeanPrendi:
Sorry for the late response, here is the stacktrace for that warning:

> doGetTable: Unable to retrieve table [table] from AWSGlue in database/schema [database]. Falling back to schema inference. If inferred schema is incorrect, create a matching table in Glue to define schema (see README)
[...]

PR #817, opened 2022-10-27
Added table_schema condition in COUNT_RECORDS_QUERY to get the correct row count.

ankita-yadav12:
*Issue # INC00001086, Check only last row count with connector snowflake.

*Description of changes:*
Added table_schema condition in COUNT_RECORDS_QUERY,to check for the table in that particular schema and give the correct row count.

### DIFF
--- a/athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
+++ b/athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
@@ -93,6 +93,7 @@ public class SnowflakeMetadataHandler extends JdbcMetadataHandler
     static final String COUNT_RECORDS_QUERY = "SELECT row_count\n" +
             "FROM   information_schema.tables\n" +
             "WHERE  table_type = 'BASE TABLE'\n" +
+            "AND table_schema= ?\n" +
             "AND TABLE_NAME = ? ";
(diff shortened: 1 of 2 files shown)


athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -173,7 +174,7 @@ public void getPartitions(BlockWriter blockWriter, GetTableLayoutRequest getTabl
         else {
             double totalRecordCount = 0;
             LOGGER.info(COUNT_RECORDS_QUERY);
-            List<String> parameters = Arrays.asList(getTableLayoutRequest.getTableName().getTableName());
+            List<String> parameters = Arrays.asList(getTableLayoutRequest.getTableName().getSchemaName(), getTableLayoutRequest.getTableName().getTableName());
```
rs-alexis, line 177: Related to line #184: Do you know why we only store the `totalRecordCount` of the latest ResultSet?

Shouldn't it be sum of the RowCount?

namratas-trianz: Hi, The query is a metadata query, to fetch the total records count for a table in snowflake. It will return single row with row_count value being retrieved.

henrymai (approved):

PR #821, opened 2022-10-31
Add CompositeHandler parameter to CF template

shivachinnam:
*Description of changes:*
We need to add an extra env variable for connection string manually to the lambda function for Athena to access any database. By allowing to use different handler we can use the Default connection string when we only have one postgres instance.


henrymai (approved):

henrymai:
Thanks for your contribution.

PR #824, opened 2022-11-04
Cut release for v2022.44.1

akuzin1:
Cut release for v2022.44.1

    Add CompositeHandler parameter to CF template (#821)
    Add support for optional LambdaRole parameter for AthenaSynapseConnector (#808)
    Added Schema name condition in COUNT_RECORDS_QUERY
    DDBPredicateUtils: Fix bug with predicate generation
    AWS Signer set empty content (#814)
    BlockUtils ClassCastException (#813)
    Improve error message with invalid spill bucket (#812)
    Don't catch just to rethrow
    ValuesGenerator: fix String:getBytes() issue


henrymai (approved):

henrymai (changes requested):
Changing status to block for now.

PR #828, opened 2022-11-11
MSK Connector changes

namratas-trianz:
*Issue #, if available:*

*Description of changes:*

Connector to pull data from Kafka Clusters


henrymai:
Hi Namratas,

Is it intentional that there are two pull requests for MSK? https://github.com/awslabs/aws-athena-query-federation/pull/818

If not, can you merge them together (or abandon the one that's not supposed to exist anymore)? Thanks.

namratas-trianz:
Hi @henrymai , Yes we will abandon https://github.com/awslabs/aws-athena-query-federation/pull/818. 
#828 need to be reviewed.

athena-msk/etc/test-config.json
```diff
@@ -0,0 +1,25 @@
+{
```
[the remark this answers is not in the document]
namratas-trianz: Hi @kparwal , I have updated README in word document with necessary screenshot, will share it across.

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/serde/MskJsonDeserializer.java
```diff
@@ -0,0 +1,95 @@
...
+                ));
+            });
+
+            close();
+            return topicResultSet;
+        }
+        catch (Exception e) {
+            throw new SerializationException("MskJsonDeserializer: Error when deserializing byte[] to TopicResultSet");
+        }
+    }
```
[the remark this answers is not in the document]
namratas-trianz: Hi @kparwal , I have handled the review comments and committed the changes.

athena-msk/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,37 @@
...
+        <Lambda name="Lambda">
+            <PatternLayout>
+                <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1}:%L - %m%n</pattern>
+            </PatternLayout>
+        </Lambda>
+    </Appenders>
+    <Loggers>
+        <Logger name="com.amazonaws.athena.connector.lambda" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-debug}" />
+        <Logger name="com.athena.connectors.msk" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-debug}" />
+        <Root level="debug">
```
henrymai, line 33: The root level should be "info" (see all of the other log4j2.xmls to see how the others are done).

athena-msk/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,37 @@
...
+    <Appenders>
+        <Lambda name="Lambda">
+            <PatternLayout>
+                <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1}:%L - %m%n</pattern>
+            </PatternLayout>
+        </Lambda>
+    </Appenders>
+    <Loggers>
+        <Logger name="com.amazonaws.athena.connector.lambda" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-debug}" />
+        <Logger name="com.athena.connectors.msk" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-debug}" />
```
henrymai, line 32: The default level should be warn not debug.

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/serde/MskCsvDeserializer.java
```diff
@@ -0,0 +1,100 @@
...
+                ));
+            });
+
+            close();
+            return topicResultSet;
+        }
+        catch (Exception e) {
+            LOGGER.error("MskCsvDeserializer: Error when deserializing byte[] to TopicResultSet");
+            e.printStackTrace();
+        }
```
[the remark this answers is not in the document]
henrymai: Spoke with Namratas this morning and she pointed out that this is because the interface they are inheriting from doesn't allow any exceptions to be thrown.

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/serde/MskJsonDeserializer.java
```diff
@@ -0,0 +1,99 @@
...
+     * @return Object
+     */
+    private Object castValue(org.apache.arrow.vector.types.pojo.Field field, String value)
+    {
+        Object object = null;
+        try {
+            object = cast(field, value);
+        }
+        catch (Exception ignored) {
+        }
```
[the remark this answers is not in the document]
henrymai: We shouldn't ignore the exception here even if the caller eventually swallows the exception. We should remove the try catch here and let the exception be logged by the calling code in this situation (since the calling code will log and eat the exception).

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/serde/MskCsvDeserializer.java
```diff
@@ -0,0 +1,100 @@
...
+     * @return Object
+     */
+    private Object castValue(org.apache.arrow.vector.types.pojo.Field field, String value)
+    {
+        Object object = null;
+        try {
+            object = cast(field, value);
+        }
+        catch (Exception ignored) {
+        }
```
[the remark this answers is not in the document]
henrymai: We shouldn't ignore the exception here even if the caller eventually swallows the exception. We should remove the try catch here and let the exception be logged by the calling code in this situation (since the calling code will log and eat the exception).

athena-msk/src/main/resources/log4j2.xml
namratas-trianz: Hi @henrymai , I have addressed the review comments.

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/serde/MskJsonDeserializer.java
```diff
@@ -0,0 +1,100 @@
...
+                catch (Exception e) {
+                    LOGGER.error("MskJsonDeserializer: Error in castValue : while converting raw value to typed value");
+                    e.printStackTrace();
+                }
+            });
+            return topicResultSet;
+        }
+        catch (Exception e) {
+            LOGGER.error("MskJsonDeserializer: Error when deserializing byte[] to TopicResultSet");
+            e.printStackTrace();
```
henrymai, line 81: nit:

Use the version of LOGGER.error where you can include the exception as the last parameter. This will automatically log the stack trace for you: https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Logger.html#error-java.lang.String-java.lang.Throwable-

I'm not sure if `e.printStackTrace()` will actually show up in our logs since that will be going to stdout instead of the logger.

Make sure to address this in all other locations that you do `e.printStackTrace()` as well.

athena-msk/athena-msk.yaml
```diff
@@ -0,0 +1,125 @@
+Transform: 'AWS::Serverless-2016-10-31'
```
henrymai, line 1: How did you guys test this with IAM authentication?

It looks like you will have to add something to this template in order for the user to specify a custom function execution role with the correct permissions otherwise it won't work, right?

See this: https://docs.aws.amazon.com/lambda/latest/dg/with-msk.html#msk-permissions-iam-policy

I think (not sure), that you'll have to do something like what was done in the athena-synapse connector with setting up a custom FunctionRole: https://github.com/awslabs/aws-athena-query-federation/blob/020df2a851250054b455d900bc6d0122b02e1b66/athena-synapse/athena-synapse.yaml#L87

Except in this case, you will have to add `Parameters` for the the various `arn:aws:kafka:___` resources that the user wants to specify to be a part of the policy document.

henrymai: Note that I could be mistaken about this btw and IAM just works with the template as is already.

If so, can you guys just provide a brief explanation as to why the function execution policy isn't necessary?

Thanks.

namratas-trianz: Hi @henrymai , For IAM, it is a custom role that need to be attached to the lambda execution role. User has to create a policy [Go to MSK cluster -> Properties ->Security Settings -> Configure IAM policy -> fetch the content and create a new policy]and attach that to lambda execution role.

henrymai: Ok, in that situation, can we do something like what was done with Synapse so that the user can specify the Function Execution Role up front as a parameter?

namratas-trianz: Sure Henry, We can do that, Basically if auth type is IAM, user must provide the custom lambda role. This will be part of readme and yaml.  If he does not provide custom lambda role, later on he has to add the custom policy to lambda execution role manually. Is that a fair understanding?

henrymai: Yes, that's fine for now I think.

namratas-trianz: Sure, will make this change available.

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/serde/MskCsvDeserializer.java
```diff
@@ -71,15 +71,13 @@ public TopicResultSet deserialize(String topic, byte[] data)
                     ));
                 }
                 catch (Exception e) {
-                    LOGGER.error("MskCsvDeserializer: Error in castValue : while converting raw value to typed value");
-                    e.printStackTrace();
+                    LOGGER.error("MskCsvDeserializer: Error in castValue : while converting raw value to typed value", e.getMessage(), e);
```
henrymai, line 74: I don't think this will actually work, you are ending up using the wrong version of .error() (you're using the variadac object version for string formatting).

All you need to do is this (notice how its just (String, Exception):
```
LOGGER.error("MskCsvDeserializer: Error in castValue : while converting raw value to typed value", e);
```

See this for reference:
https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Logger.html#error-java.lang.String-java.lang.Throwable-
```
void [error](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/src-html/org/apache/logging/log4j/Logger.html#line.893)([String](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html?is-external=true) message,
           [Throwable](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html?is-external=true) throwable)

Logs a message at the [ERROR](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Level.html#ERROR) level including the stack trace of the [Throwable](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html?is-external=true) throwable passed as parameter.

Parameters:
    message - the message object to log.
    throwable - the Throwable to log, including its stack trace.
```

athena-msk/pom.xml
```diff
@@ -0,0 +1,197 @@
...
+    <artifactId>athena-msk</artifactId>
+    <name>Athena MSK Connector</name>
+    <version>2022.46.1</version>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+        <slf4jVersion>1.7.30</slf4jVersion>
+        <log4j2Version>2.16.0</log4j2Version>
+        <aws.lambda-java-log4j2.version>1.5.0</aws.lambda-java-log4j2.version>
```
henrymai, line 21: Remove `slf4jVersion`

Remove `log4j2Version` 

Remove `aws.lambda-java-log4j2.version`

All of these have definitions in the root pom.xml already and can just be used.

athena-msk/pom.xml
```diff
@@ -0,0 +1,197 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
+            <version>3.3.1</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <version>2.13.4</version>
```
henrymai, line 32: Replace this line with:
```
<version>${fasterxml.jackson.version}</version>
```

This property is already defined in the root pom.xml.

This is especially important because I just had to deal with a customer request asking us to bump this version today. Making use of the root property will automatically address it here too.

athena-msk/pom.xml
```diff
@@ -0,0 +1,197 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>${slf4jVersion}</version>
```
henrymai, line 37: Replace with: `${slf4j-log4j.version}`

athena-msk/pom.xml
```diff
@@ -0,0 +1,197 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>31.0-jre</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4jVersion}</version>
```
henrymai, line 55: Replace with: `${slf4j-log4j.version}`

henrymai (commented):
Very close to being done. Just waiting on the changes for athena-msk.yaml in regards to allowing the user to pass in a custom execution role.

athena-msk/pom.xml
namratas-trianz: Hi @henrymai , one query, just LambdaRoleARN is enough right? Do we need to add PermissionBoundary ?

henrymai: Yes, I think the LambdaRoleARN should be enough for now. But we may get a request for PermissionBoundary in the future. We don't have to deal with it right now.

athena-msk/athena-msk.yaml
```diff
@@ -0,0 +1,149 @@
...
+            Effect: Allow
+            Resource: !Sub 'arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:*'
+          - Action:
+              - logs:CreateLogStream
+              - logs:PutLogEvents
+            Effect: Allow
+            Resource: !Sub 'arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/${LambdaFunctionName}:*'
+          - Action:
+              - athena:GetQueryExecution
+              - s3:ListAllMyBuckets
```
henrymai, line 137: I noticed that the following actions are now missing:
```
- kms:Decrypt
- glue:*
```


Is this intentional?

henrymai (approved):

Issue #399, opened 2021-04-28
[FEATURE] DynamoDB nullable nested struct


joshuanapoli:
In the DynamoDB connector, a nested struct property with value `null` is treated as an error. On the other hand, a missing property is mapped by the connector to `null`.

To be clear, I have a column with schema like this:
```
struct<id:string,nested:struct<prop:string>>
```

A DynamoDB value like this "works"; the nested struct property is absent. Through the connector, Athena gets `{id: "works", nested: null}`.
```
{
  id: "works"
}
```

A DynamoDB value like this fails; the connector throws an exception because nested is null, which does not match the `struct<prop:string>` type:
```
{
  id: "fails",
  nested: null
}
```

I'd prefer to have missing property treated the same as null value for the property; both should lead to null a value for the property in Athena.

We could add an if null return null check at https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/resolver/DynamoDBFieldResolver.java#L80

We could potentially change the way that our application persists records to DynamoDB to be compatible with the Connector's treatment of null and missing values for nested structs. But this is the tail wagging the dog; we expect the federated query system to be flexible to adapt to our application's existing schemas.

james-ni:
I got into same problem. 

When I have 2 items in DynamoDB table

```json
# item 1
{
  "id": "001",
  "location": {
    "suburb": {
      "name": "some_value"
    }
  }
}

# item 2
{
  "id": "002",
  "location": {
    "suburb": null
  }
}
```

The Athena query `select * from <DYNAMODB_TABLE>` failed and Lambda cloudwatch logs showed:

```
Error while processing field location: java.lang.RuntimeException
java.lang.RuntimeException: Error while processing field location
	at com.amazonaws.athena.connectors.dynamodb.DynamoDBRecordHandler.lambda$readWithConstraint$0(DynamoDBRecordHandler.java:193)
	at com.amazonaws.athena.connector.lambda.data.S3BlockSpiller.writeRows(S3BlockSpiller.java:174)
	at com.amazonaws.athena.connectors.dynamodb.DynamoDBRecordHandler.readWithConstraint(DynamoDBRecordHandler.java:149)
	at com.amazonaws.athena.connector.lambda.handlers.RecordHandler.doReadRecords(RecordHandler.java:192)
[...]

PR #839, opened 2022-11-23
Avoid errors on null values for dynamodb

pappacena:
*Issue #, if available:* https://github.com/awslabs/aws-athena-query-federation/issues/399

*Description of changes:*
As described in the issue, dynamodb connector breaks when there is `null` values in complex structures. There is a comment in the original issue mentioning that [this PR](https://github.com/awslabs/aws-athena-query-federation/pull/780) fixed the issue, but I'm still facing the same described error when my data structure contains `null`.

I've tested manually the connector with the following change, and it works fine.

I would like some feedback to know if this is an acceptable change, and if anything else would be needed for this PR to be accepted.

### DIFF
--- a/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/resolver/DynamoDBFieldResolver.java
+++ b/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/resolver/DynamoDBFieldResolver.java
@@ -82,6 +82,10 @@ public class DynamoDBFieldResolver
                 return DDBTypeUtils.coerceValueToExpectedType(fieldValue, field, fieldType, metadata);
         }
 
+	if (fieldValue == null) {
+            return null;
+	}
+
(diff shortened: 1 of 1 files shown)


athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/resolver/DynamoDBFieldResolver.java
```diff
@@ -82,6 +82,10 @@ public Object getFieldValue(Field field, Object originalValue)
                 return DDBTypeUtils.coerceValueToExpectedType(fieldValue, field, fieldType, metadata);
         }
 
+        if (fieldValue == null) {
+            return null;
+        }
+
```
henrymai, line 88: Can you move this check earlier? To like under line 70?

pappacena: Sure! Done!

henrymai (approved):

PR #843, opened 2022-11-28
feature/athena gcs

trianz-azam:
*Issue #, if available:*
None

*Description of changes:*
The Amazon Athena connector for **GCS (Google Cloud Storage)** enables Amazon Athena to run SQL queries on uploaded **PARQUET** or **CSV** files in the GCS buckets.

This connector uses a newer Apache Arrow Dataset API, a contribution of @henrymai to Athena. This specific version of Arrow Dataset API (v10.0.1)  is used to retrieve metadata and fetch results from either **PARQUET** or **CSV** files.


athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsCompositeHandler.java
```diff
@@ -0,0 +1,44 @@
...
+public class GcsCompositeHandler
+        extends CompositeHandler
+{
+    /**
+     * The default constructor that initializes metadata and record handlers for GCS
+     */
+    public GcsCompositeHandler() throws InvocationTargetException, NoSuchMethodException, InstantiationException, IllegalAccessException, IOException
+    {
+        super(new GcsMetadataHandler(), new GcsRecordHandler());
+        installCaCertificate();
```
henrymai, line 42: Shouldn't `installGoogleCredentialsJsonFile` also be called here so that you don't have to call it multiple times later on in the metadatahandler and then in the recordhandler.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsUtil.java
```diff
@@ -0,0 +1,114 @@
...
+        File src = new File(file.getAbsolutePath() + File.separator + "cacert.pem");
+        File dest = new File(Paths.get("/tmp").toAbsolutePath() + File.separator  + "cacert.pem");
+        if (!dest.exists()) {
+            Files.copy(src.toPath(), dest.toPath(), StandardCopyOption.REPLACE_EXISTING);
+        }
+    }
+
+    public static void installGoogleCredentialsJsonFile(String gcsCredentialJsonString) throws IOException
+    {
+        File dest = new File(Paths.get("/tmp").toAbsolutePath() + File.separator  + GOOGLE_SERVICE_ACCOUNT_JSON_TEMP_FILE_NAME);
```
henrymai, line 105: Just use the `GOOGLE_APPLICATION_CREDENTIALS` environment variable as dest rather than hardcoding it to `/tmp/service-account.json`

This way you don't have to worry as much about the user changing this environment variable. As long as they set it to a path that we have write permissions to, it will just work and we don't have to worry about it.

However, you should probably make sure to create all the parent directories for the file that the environment variable points to (if the user sets something with more than one level then we'll fail even though the code could have just automatically created the parent paths).

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsUtil.java
```diff
@@ -0,0 +1,114 @@
...
+        return field.getType() == null
+                || field.getType().equals(Types.MinorType.NULL.getType());
+    }
+
+    public static void installCaCertificate() throws IOException
+    {
+        ClassLoader classLoader = GcsRecordHandler.class.getClassLoader();
+        File file = new File(requireNonNull(classLoader.getResource("")).getFile());
+        File src = new File(file.getAbsolutePath() + File.separator + "cacert.pem");
+        File dest = new File(Paths.get("/tmp").toAbsolutePath() + File.separator  + "cacert.pem");
```
henrymai, line 97: Just use the environment variable `SSL_CERT_FILE` to set up the dest instead of hard coding this to `/tmp/cacert.pem`

This way you don't have to worry as much about the user changing this environment variable. As long as they set it to a path that we have write permissions to, it will just work and we don't have to worry about it.

However, you should probably make sure to create all the parent directories for the file that the environment variable points to (if the user sets something with more than one level then we'll fail even though the code could have just automatically created the parent paths).

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsRecordHandler.java
```diff
@@ -0,0 +1,191 @@
...
+        ) {
+            // We are loading records batch by batch until we reached at the end.
+            while (reader.loadNextBatch()) {
+                try (
+                        // Returns the vector schema root.
+                        // This will be loaded with new values on every call to loadNextBatch on the reader.
+                        VectorSchemaRoot root = reader.getVectorSchemaRoot()
+                ) {
+                    // We will loop on batch records and consider each records to write in spiller.
+                    for (int rowIndex = 0; rowIndex < root.getRowCount(); rowIndex++) {
```
henrymai, line 153: I think you can invert the loops on line 153 and 177.

Meaning in pseudo code:

    for each column:
        write all the row elements

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsSchemaUtils.java
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Builds the table schema based on the provided field by the retrieved instance of {@link StorageTable}
+     *
+     * @param datasource   An instance of {@link StorageMetadata}
+     * @param databaseName Name of the bucket in GCS
+     * @param tableName    Name of the storage object (file) from GCS
+     * @return An instance of {@link Schema}
+     */
+    public static Schema buildTableSchema(StorageMetadata datasource, String databaseName, String tableName) throws Exception
```
henrymai, line 70: Why do we need this? Can't we just use the Arrow Dataset API without reading the records in to read just the schema? This way you won't have to build anything, you can simply just return the Schema.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsSchemaUtils.java
```diff
@@ -0,0 +1,147 @@
...
+            Schema schema = schemaBuilder.build();
+            return schema;
+        }
+        else {
+            LOGGER.error("Table '{}' was not found under schema '{}'", tableName, databaseName);
+            throw new GcsConnectorException("Table '" + tableName + "' was not found under schema '" + databaseName + "'");
+        }
+    }
+
+    public static Field getCompatibleField(Field field)
```
henrymai, line 93: Same here, why are we doing anything with types here? Shouldn't the Dataset API just gives us the correct Arrow Schema Fields?

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/storage/AbstractStorageMetadata.java
```diff
@@ -0,0 +1,713 @@
...
+        for (Bucket bucket : buckets.iterateAll()) {
+            String bucketName = bucket.getName();
+            String validName = getValidEntityName(bucketName);
+            if (!databaseBuckets.containsKey(validName)) {
+                loadedEntitiesList.add(new LoadedEntities(validName));
+                databaseBuckets.put(validName, bucketName);
+            }
+        }
+        return ImmutableList.copyOf(databaseBuckets.keySet());
+    }
```
henrymai, line 162: I don't think this is correct.
Not all root buckets are going to be hive stores.
Additionally there could be databases nested under a non root bucket as well, right?

I haven't had time yet to come up with an alternative, but just wanted to call this out first.

athena-gcs/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,37 @@
...
+    <Appenders>
+        <Lambda name="Lambda">
+            <PatternLayout>
+                <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1}:%L - %m%n</pattern>
+            </PatternLayout>
+        </Lambda>
+    </Appenders>
+    <Loggers>
+        <Logger name="com.amazonaws.athena.connector.lambda" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-info}"/>
+        <Logger name="com.amazonaws.athena.connectors.gcs" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-info}"/>
```
henrymai, line 32: Default level should be warn for these. See the other log4j2.xml files.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsSchemaUtils.java
henrymai: I actually see that you're doing what I mention in AbstractStorageMetadata.java already:

line 138:
```
       BufferAllocator allocator = new RootAllocator();
        DatasetFactory factory = new FileSystemDatasetFactory(allocator,
                NativeMemoryPool.getDefault(), getFileFormat(), uri);
        // inspect schema
        return factory.inspect().getFields();
```


So is there a reason you aren't doing something similar here?

henrymai (commented):
Initial quick glance review so far. I still need to dive deeper and think about some other things, but I just wanted to call out some obvious things first.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/storage/AbstractStorageMetadata.java
henrymai: Another thought I just had is that it seems like you guys should consider maybe integrating with Glue for this.

Since this is exactly the reason why the Hive metadata store exists, so that people can express the location of their data and schema of their data.

Again, this is just me commenting on this from just glancing, so I could be missing something.

If you do go down the route of integrating with Glue, I just wanted to call out that you can coerce the existing data that is read in by the Dataset api into using a new Schema by just using this constructor:

VectorSchemaRoot(Schema schema, List<FieldVector> fieldVectors, int rowCount)

 - Where you pass in the new schema from Glue and then use the existing fieldVectors, and row count, from an existing VectorSchemaRoot.
 -  Also note that we already have a GlueMetadataHandler that you can extend that will already automatically read the Glue schema string and convert it to an Arrow Schema for you, see: `GlueMetadataHandler.java:doGetTable(BlockAllocator blockAllocator, GetTableRequest request, TableFilter filter)` line 422

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsSchemaUtils.java
trianz-azam: Yes, the [GcsSchemaUtils](https://github.com/awslabs/aws-athena-query-federation/pull/843/files/f4960a5a4e2e32000ac11914879bdfefed823ee2#diff-b8cb256ec5759f01b9431fa24e85b955758c8d00ab2c4369e48c454566f4e4b4) essentially used this method to get all the fields. It just iterates through the Field collection and adds an extra partition Field.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsSchemaUtils.java
trianz-azam: Sometimes it returns an incompatible type per Athena specification. For example, Time (NONOSECONDS) isn't compatible with Athena. We make such a field compatible with Athena and return in such a case.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsCompositeHandler.java
trianz-azam: Taken care

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/storage/AbstractStorageMetadata.java
trianz-azam: We'll integrate Glue as a future enhancement

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsRecordHandler.java
trianz-azam: This loop iterates through the rows in the Dataset reader, while the later loop is within a lambda for each field vector to spill. I think inverting won't work here. I might misunderstand. 

If I am wrong, please correct me.

henrymai: What you are currently doing is iterating over the row indices and then you iterate over each column vector for a given row index.

I'm explaining to iterate over the columns in the outer loop instead.
This is for exploiting cache locality since Arrow is a column oriented format, meaning row elements per column are going to be automatically prefetched into the cache when you iterate over it.

trianz-azam: Loop inverted

henrymai: I just looked at your change and this is not what I am suggesting. This is even worse because you hare making at least two copies now (first copy is when you read into `List<Object> columnValues` and then another copy when you do `transpose`).

Again, the entire point of Arrow is to avoid copies and that its column oriented.

Please revert your inversion commit and I will have to put work into this myself...

trianz-azam: Hi @henrymai, I have reverted the last commit; as you said

henrymai:
Please rebase your changes properly against master. This pull request is currently showing commits that are duplicated from master in your pull request.

trianz-azam:
>

trianz-azam:
Hi @henrymai, can we create a new PR?

henrymai:
Yes you can create a new PR if you need to.

trianz-azam:
Hi @henrymai , as you have confirmed, we have created a new PR here: https://github.com/awslabs/aws-athena-query-federation/pull/929

Please review it and give us your valuable feedback

Issue #845, opened 2022-12-01
[BUG] Querying bigquery from multiple projects with one service account


asa-bell:
I need to query bigquery from multiple projects via athena.

I have generated a service account in project_A and authorized this service account to access bigquery in project_B.
In lambda's environment variable, I specified gcp_project_id as project_B.
After the configuration in athena, I can see the datasets, databases, and tables in project_B in the athena console.
But when I query the data of project_B with SQL, I get the error: The specified dataset is not found in project_A.

In the "project history" of bigquery, I see that the query is in project_A. The project is not specified in SQL.

Any help is appreciated.

jason-pixl:
### Problem analysis

The problem should be in the aws-athena-query-federation repository.

The latest master branch(332ff5fb5ef72ef9f5847c0a7b15a11b80ec5b54) and v2022.47.1(tag), the problem should exist

The bug will occur while using the different configuration method when using aws-athena-query-federation: the same SecretNamePrefix corresponds to multiple GCPProjectIDs

### Code analysis

The following codes are analyzed and modified based on master branch

(git commit id: 332ff5fb5ef72ef9f5847c0a7b15a11b80ec5b54)

The code does not explicitly execute setProjectId when initializing BigQuery, but the projectId calculated by SecretNamePrefix

When initializing BigQuery, java-bigquery provides a method to set projectId, but it does not use

When querying datasets, databases, and tables, the projectId is explicitly set, so it can work normally.

However, when performing SQL query, the projectId is not set. After studying the source code of java-bigquery, when executing the SQL statement, the projectId will be obtained from BigQuery, and then the SQL statement will be executed.

```
public static BigQuery getBigQueryClient() throws IOException
{
    BigQueryOptions.Builder bigqueryBuilder = BigQueryOptions.newBuilder();
    String endpoint = System.getenv(BigQueryConstants.BIG_QUERY_ENDPOINT);
    if (StringUtils.isNotEmpty(endpoint)) {
        bigqueryBuilder.setHost(endpoint);
    }

    bigqueryBuilder.setCredentials(getCredentialsFromSecretsManager());
    return bigqueryBuilder.build().getService();
}
```

henrymai:
Hi @asa-bell 

I think I know what the problem is, can you confirm if your project names have mixed casing or if they are all lower case?

henrymai:
[...]

PR #862, opened 2022-12-06
[BUG] Querying bigquery from multiple projects with one service account

jason-pixl:
*Issue #, if available:*https://github.com/awslabs/aws-athena-query-federation/issues/845

*Description of changes:*https://github.com/awslabs/aws-athena-query-federation/issues/845#issuecomment-1338826803

### DIFF
--- a/athena-timestream/athena-timestream.yaml
+++ b/athena-timestream/athena-timestream.yaml
@@ -65,8 +65,9 @@ Resources:
                 - glue:GetDatabase
                 - athena:GetQueryExecution
                 - s3:ListAllMyBuckets
-                #TODO: make this more specific later
-                - timestream:*
+                - timestream:Describe*
+                - timestream:List*
+                - timestream:Select*
               Effect: Allow
               Resource: '*'
           Version: '2012-10-17'


henrymai:
Hi @jason-pixl 

Thank you for looking into this and working on this.

You are on the right trail here in regards to setting the `projectId` on the BigQuery client, because I do notice that we don't set the `projectId` anywhere in `BigQueryRecordHandler::readWithConstraint` whereas we do at least set it as part of the api invocations in the `BigQueryMetadataHandler` (where the issue author actually does mention they were able to retrieve project B's datasets, tables, etc correctly).

However, the environment variable is intended to be an override, where typically the `projectId` is instead derived from the Athena SQL query, which provides the user much more flexibility than having to set the environment variable each time before they make a query.

So instead of doing that, we should actually be setting the `projectId` on the BigQuery client after the code calls `BigQueryUtils.getProjectName()` (which does make use of the environment variable override if set, but if not, just gets the `projectId` from the Athena SQL query by default).

For reference see the code base for usages and the definition of:
```BigQueryUtils.getProjectName(listTablesRequest);```

In order to do what I mention above, we would need to refactor the code so that it doesn't just construct the BigQueryClient immediately in the constructor of (at least) the BigQueryRecordHandler.

I will push an update to your pull request with the refactor to show you what I mean.

jason-pixl:
> Hi @jason-pixl
> 
> Thank you for looking into this and working on this.
> 
> You are on the right trail here in regards to setting the `projectId` on the BigQuery client, because I do notice that we don't set the `projectId` anywhere in `BigQueryRecordHandler::readWithConstraint` whereas we do at least set it as part of the api invocations in the `BigQueryMetadataHandler` (where the issue author actually does mention they were able to retrieve project B's datasets, tables, etc correctly).
> 
> However, the environment variable is intended to be an override, where typically the `projectId` is instead derived from the Athena SQL query, which provides the user much more flexibility than having to set the environment variable each time before they make a query.
> 
> So instead of doing that, we should actually be setting the `projectId` on the BigQuery client after the code calls `BigQueryUtils.getProjectName()` (which does make use of the environment variable override if set, but if not, just gets the `projectId` from the Athena SQL query by default).
> 
> For reference see the code base for usages and the definition of: `BigQueryUtils.getProjectName(listTablesRequest);`
> 
> In order to do what I mention above, we would need to refactor the code so that it doesn't just construct the BigQueryClient immediately in the constructor of (at least) the BigQueryRecordHandler.
> 
> I will push an update to your pull request with the refactor to show you what I mean.

In the beginning, I studied how QueryJobConfiguration sets projectId, but unfortunately, although google-cloud-bigquery provides it, it is a private method that is not exposed to the outside world, but when executing SQL statements, they will go to bigQueryClient Read the corresponding projectId, if you refactor, this may help you

henrymai:
Right, the idea is to set it on the Builder before creating the client. Which is why we need to refactor the code so that we can do so. I will have my patch up shortly.

henrymai:
I didn't have permissions to push to your fork and branch, so I had to create a new pull request with my changes.

Here it is: https://github.com/awslabs/aws-athena-query-federation/pull/863/files

Took me longer than expected because of dealing with the tests and PowerMockito.

PR #866, opened 2022-12-08
Added Condition for Function Execution Policy

adithyapathipaka:
*Issue #, if available:*

Unable to Create DynamoDB Connectors if Custom Lambda Role is Used.
Screenshot 2022-12-08 at 17 42 29

*Description of changes:*

Added Condition Check for Function Policy to create only if Custom Function role is passed to SAM Resource


henrymai (approved):
Thanks for your contribution.

Issue #827, opened 2022-11-09
[FEATURE] Simplify the use of the default connector string with data sources


JGoutin:
**Is your feature request related to a problem? If yes, please describe.**

It seem to not be possible to use the default connection string of a connector with an Athena data source.

In the doc we have:

> `$catalog_connection_string` Required. A database instance connection string. Prefix the string with the name of the catalog used in Athena. For example, if the catalog registered with Athena is `mymysqlcatalog`, then the environment variable name is `mymysqlcatalog_connection_string`.
> 
> `default` Required. The default connection string. This string is used when the catalog is `lambda:${AWS_LAMBDA_FUNCTION_NAME}`.

But:
* It is not possible to name a data source as `lambda:${AWS_LAMBDA_FUNCTION_NAME}`.
* Naming a data source `${AWS_LAMBDA_FUNCTION_NAME}` don't work better.

Using `$catalog_connection_string` look like a dirty workaround for me because:
* The connector is managed using Serverless Application Repository/Cloudformation, but there is no way to configure `$catalog_connection_string` in the template (There is only `DefaultConnectionString` parameter). In addition, my infrastructure is managed with Terraform. => But for this connector I need to manually add an environment variable to the lambda, and this variable will likely disappear or become outdated if the configuration change and the lambda is redeployed or simply if I change the password.
* I have only a single data source, and I need to set my connection string twice (In `default` + in `$catalog_connection_string` variables) with the same value.
* This need an extra manual intervention (So reduce the configuration ease, make the data source not working by default, ...).

**Describe the solution you'd like**

Have a straightforward solution to configure a connector that use the default connection string with a data source.

Why simply not using the `default` value if there is no relevant `$catalog_connection_string` value ? In this case  `$catalog_connection_string`  is an override, and `default` a real default value. This is a more far simple and common behavior.

Or at least, ensure that a data source named `${AWS_LAMBDA_FUNCTION_NAME}` use the default connection string.

**Describe alternatives you've considered**
The only alternative is to manually set `$catalog_connection_string` on the lambda.

**Additional context**

See also, I am likely not the first one to lost some time on this :
#683 #215 #418 #517

henrymai:
Hi,

Do you think this patch will address your use case? https://github.com/awslabs/aws-athena-query-federation/pull/870

JGoutin:
Hi, 
Yes, this patch address my use case.

henrymai:
Hi @JGoutin 

[...]

Issue #683, opened 2022-03-17
[BUG] Default connector cannot be named as described by Redshift connector documentation


Limess:
**Describe the bug**
The redshift connector documentation suggests that if the Athena catalog is named `lambda:${AWS_LAMBDA_FUNCTION_NAME}` then the default connection string will be used.

However, this is an invalid Athena catalog name.

**Expected behavior**
It should be possible to create a connector which uses the default connection string, without then having to modify the created lambda function manually.

**Connector Details (please complete the following information):**
 - Version: 2022.10.1
 - Name Redshift

Limess:
Hmm, after some reading it seems that the idea is to _not_ add the connectors as Athena datasources, but just query at runtime as  `lambda:${AWS_LAMBDA_FUNCTION_NAME}`? But this does not allowing viewing catalog metadata/tables up front?

henrymai:
Hi @Limess 

In regards to:
> Expected behavior
> It should be possible to create a connector which uses the default connection string, without then having to modify the created lambda function manually.

You should be able to just use the default connection string now if you do not have any other connection strings defined after this patch: 65c29e4bad1d89634b30b3604abc90b4adfabc3e

Please upgrade to any version of the Redshift connector after v2023.01.1 to get this fix.

Thanks.

PR #870, opened 2022-12-10
Allow the user to choose the CompositeHandler

henrymai:
Allow the user to choose the CompositeHandler

This is useful when the user has only a single connection that and they do
not want to modify the environment variables on their lambda after deployment.

In this situation, with this patch, a user during deployment can:

  - Set the DefaultConnectionString to their only connection string.

  - Set the CompositeHandler to *CompositeHandler (instead of
    *MuxCompositeHandler)

Original idea and implementation comes from this pull request for the postgres connector: #821

This patch addresses issues: #827, #683


henrymai:
Replaced with an even simpler patch: https://github.com/awslabs/aws-athena-query-federation/pull/936

PR #891, opened 2022-12-13
Optimized offset logic

feroz-trianz:
*Issue #, if available:*

*Description of changes:*
We have optimized the offset logic for Kafka and added some debug log.

### DIFF
--- a/athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskMetadataHandler.java
+++ b/athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskMetadataHandler.java
@@ -173,13 +173,23 @@ public class AmazonMskMetadataHandler extends MetadataHandler
                 .map(it -> new TopicPartition(it.topic(), it.partition()))
                 .collect(Collectors.toList());
 
-        LOGGER.debug("[kafka][split] total partitions {} found for topic: {}", topicPartitions.size(), topic);
+        LOGGER.debug("[KafkaPartition] total partitions {} found for topic: {}", topicPartitions.size(), topic);
+
+        // Get start offset of each topic partitions from kafka server.
+        Map<TopicPartition, Long> startOffsets = kafkaConsumer.beginningOffsets(topicPartitions);
+        if (LOGGER.isDebugEnabled()) {
+            startOffsets.forEach((k, v) -> {
+                LOGGER.debug("[KafkaPartitionOffset] start offset info [topic: {}, partition: {}, start-offset: {}]",
+                        k.topic(), k.partition(), v
+                );
+            });
+        }
 
         // Get end offset of each topic partitions from kafka server.
         Map<TopicPartition, Long> endOffsets = kafkaConsumer.endOffsets(topicPartitions);
         if (LOGGER.isDebugEnabled()) {
             endOffsets.forEach((k, v) -> {
-                LOGGER.debug("[kafka][split] offset info [topic: {}, partition: {}, end-offset: {}]",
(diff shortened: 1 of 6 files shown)


athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskMetadataHandler.java
```diff
@@ -256,39 +273,43 @@ private Schema getSchema(String tableName) throws Exception
...
+            yOffset = xOffset + MAX_RECORDS_IN_SPLIT;
 
-            // The last endOffset of the last piece must not be greater than the endOffset
-            // of topic partition, it will be equal to endOffset of the topic partition.
-            if (endOffset < yOffset) {
+            // The last yOffset of the last piece must not be greater than the endOffset
+            // of the topic partition, it will be at least equal to endOffset of the topic partition.
+            if (yOffset > endOffset) {
                 yOffset = endOffset;
             }
```
henrymai, line 314: Replace lines 312 - 314 with:

```
yOffset = Math.min(yOffset, endOffset);
```

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskRecordHandler.java
```diff
@@ -176,23 +200,28 @@ private void consume(
...
      */
-    private void execute(
+    private boolean execute(
             BlockSpiller spiller,
             ReadRecordsRequest recordsRequest,
             QueryStatusChecker queryStatusChecker,
             SplitParameters splitParameters,
             ConsumerRecord<String, TopicResultSet> record)
     {
+        final boolean[] isExecuted = {false};
```
henrymai, line 212: Why is this an array? Only one element is ever written to or read from.

Also when you make this back into a normal boolean, you should probably remove the final modifier there's no reason for it to be final if you want to write to it.

feroz-trianz: Variable used in lambda expression should be final or effectively final - that's why we have used this approach.

Alternate way we can do the following:

        AtomicBoolean isExecuted = new AtomicBoolean(false);
        spiller.writeRows((Block block, int rowNum) -> {
            for (MSKField field : record.value().getFields()) {
                boolean isMatched = block.offerValue(field.getName(), rowNum, field.getValue());
                if (!isMatched) {
                    return 0;
                }
            }
            // For debug insight
            splitParameters.spilled += 1;
            isExecuted.set(true);
            return 1;
        });
        return isExecuted.get();


Let us know your thought.

henrymai: Since all you're using this boolean for is logging, can you just use the logger on line 222 and not use a boolean at all (and also of course just turn it back into `void execute(...)`).

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskMetadataHandler.java
feroz-trianz: Changed accordingly

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskRecordHandler.java
feroz-trianz: Changed accordingly

henrymai (approved):

PR #994, opened 2023-01-28
athena-gcs: some more cleanup

henrymai:
- Simplified the logic in checkPartitionWithConstraints/partitionConstraintsSatisfied to make it easier to reason about.
- Various other simplifications and cleanups.

### DIFF
--- a/athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
+++ b/athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
@@ -173,7 +173,7 @@ public class GcsMetadataHandler
             Schema schema = datasource.buildTableSchema(table, allocator);
             Map<String, String> columnNameMapping = getColumnNameMapping(table);
             List<Column> partitionKeys = table.getPartitionKeys() == null ? List.of() : table.getPartitionKeys();
-            Set<String> partitionCols = table.getPartitionKeys().stream()
+            Set<String> partitionCols = partitionKeys.stream()
                 .map(next -> columnNameMapping.getOrDefault(next.getName(), next.getName())).collect(Collectors.toSet());
             return new GetTableResponse(request.getCatalogName(), request.getTableName(), schema, partitionCols);
         }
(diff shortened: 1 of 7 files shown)


athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/storage/StorageMetadata.java
```diff
@@ -200,26 +192,22 @@ protected Schema getFileSchema(String bucketName, String path, FileFormat format
...
-            columnValueConstraintMap.getOrDefault(partitionEntry.getKey(), Optional.empty()).stream()
-                .anyMatch(valueConstraintSet -> {
-                    boolean result = valueConstraintSet.contains(partitionEntry.getValue());
-                    LOGGER.debug("Partition entry: {}, valueConstraintSet: {}, result: {}", partitionEntry, valueConstraintSet, result);
-                    return !result;
-                })
-        );
-        return !constraintViolationExists;
+        // For all constraints (column name -> set of valid values), validate that any partition column being
+        // constrained has a value that is within the constrained value set.
```
hackett123, line 203: nice rewrite

hackett123 (approved):
Approving, but we will be waiting until Trianz can test this before merging (this is a note for any other reviewers)

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
```diff
@@ -172,7 +171,6 @@ public GetTableResponse doGetTable(BlockAllocator blockAllocator, GetTableReques
             //fetch schema from dataset api
             Schema schema = datasource.buildTableSchema(table, allocator);
             Map<String, String> columnNameMapping = getColumnNameMapping(table);
-            List<Column> partitionKeys = table.getPartitionKeys() == null ? List.of() : table.getPartitionKeys();
             Set<String> partitionCols = table.getPartitionKeys().stream()
```
akuzin1, line 174: According to line 175 it is possible that `table.getPartitionKeys()` may return the value null, therefore we should handle that case in the following line where we're setting variable `partitionCols`.

henrymai: Great catch. This was a left over bug where `partitionKeys` should have been used instead of being deleted.

PR #1019, opened 2023-02-03
athena-gcs: eliminate reliance on external environment variables

henrymai:
Uses jna to set the environment for native calls and avoids any possibility of users setting the wrong values.


hackett123 (approved):
this is going to make things easier for users, nice find. let's contact Tim as soon as Trianz does testing.

PR #1032, opened 2023-02-06
athena-gcs: fix bug with trimming leading /

henrymai:
In StorageMetadata.java:172 prior to this patch, it was trying to trim the leading / even when it didn't exist.

This patch fixes that and also fixes all other places in the athena-gcs connector where it tries to trim leading slashes using the error prone `.substring(1)` method.


hackett123 (approved):
looks good. nice arrayindexoutofbounds catch

Issue #365, opened 2021-03-05
[FEATURE] Expose IAM customizations in SAM templates


davidaah:
**Is your feature request related to a problem? If yes, please describe.**

Currently the published Athena connectors do not support bringing your own role or providing an IAM permissions boundary ARN. This limits the ability to consume these connectors in a governed environment where application develops do not have access to create arbitrary IAM roles

**Describe the solution you'd like**

Each template should expose two additional parameters corresponding to the 'AWS::Serverless::Function' resource, specifically the `Role` and `PermissionsBoundary` attributes (see https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html for more details). 

**Describe alternatives you've considered**

The only workaround we have currently would be to fork the repo, make manual modifications to the configuration yaml and re-publish in to a private serverless application repository in our accounts.

**Additional context**

N/A

tkcmada:
👍

victorfeight:
Hello Team. Any updates on when a custom LambdaRole parameter would be available for usage in the PostgreSQL connector? https://github.com/awslabs/aws-athena-query-federation/tree/master/athena-postgresql

Otherwise, if there's a custom working SAM template with the Role included, I could try to deploy to Serverless Application Repository. I wasn't able to adapt the DynamoDB connector's LambdaRole parameter to a custom yaml template, facing:
Error: [InvalidTemplateException('Every Condition member must be a string.')] Every Condition member must be a string.

Attached is my go at it:
[template.yml.txt](https://github.com/awslabs/aws-athena-query-federation/files/10523339/template.yml.txt)

henrymai:
@victorfeight 

In the `template.yml.txt` that you attached, it looks like the issue is due to this:

```
Condition: !Not HasLambdaRole
```

Instead of doing it this way:

```
Condition: NotHasLambdaRole
```

I've seen that error myself before when I also tried to use the condition like `!Not HasLambdaRole`, which is why the condition is now `NotHasLambdaRole` instead.

victorfeight:
Thank you, that did help.

henrymai:
@victorfeight 

Glad to hear it works for you.

Also feel free to submit a pull request if you think your changes are general enough to be used by everyone else too (NOTE: I didn't do a diff with your example so I don't know if it is general or not).

victorfeight:
[...]

PR #1065, opened 2023-02-22
Add LambdaRoleARN parameter to sqlserver and mysql connectors

ned9:
Add `LambdaRoleARN` parameter to athena-sqlserver and athena-mysql connectors

This adds the ability to specify an ARN of the execution role for the deployed lambda connector. If none is supplied, then a role is generated with same policies as previous behaviour. 

These changes are based on the changes made to achieve the same result for the athena-postgresql connector in #995.

See also #365.


henrymai:
Hi @ned9 

Just wanted to confirm, you've tested that these changes work, right?

ned2:
Yep, deployed both connector templates to our account, and deployed stacks for each, testing both pathways: with and without an ARN provided for the new parameter. And we get the expected behaviour in each case.

henrymai (approved):

henrymai:
Just forcefully merged it since the build workflow seems to be broken. Thanks for the contribution.

PR #1090, opened 2023-03-02
Refactor code to not use System.getenv directly

henrymai:
Refactor code to not use System.getenv directly
 - This change is necessary for future features.

Get rid of environment variable mocking
 - This change is for general cleanup since we don't need this anymore.


hackett123 (approved):
This was unpleasant to review, so I can only imagine how unpleasant it was to write. Nice update

PR #1120, opened 2023-03-16
GCS connector fix partition column value issue

Trianz-Akshay:
fix for GCS partition column value issue when partition column values are not present in files.

added partition columns and respective values in the split and offered that from split to read with constraint.
need to pass the partition columns and split to execute the method to set the value in the spiller. 

Tested the fix for partition and nonpartitioned data sources with glue schema.

### DIFF
--- a/athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
+++ b/athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
@@ -259,6 +259,8 @@ public class GcsMetadataHandler
             // set partition column name and value in split
             for (FieldVector fieldVector : partitions.getFieldVectors()) {
                 fieldVector.getReader().setPosition(curPartition);
+                if (fieldVector.getName().equalsIgnoreCase(FILE_FORMAT) || fieldVector.getName().equalsIgnoreCase(STORAGE_SPLIT_JSON))
+                    throw new RuntimeException("column name is same as metadata");
                 splitBuilder.add(fieldVector.getName(), fieldVector.getReader().readObject().toString());
             }
             splits.add(splitBuilder.build());
--- a/athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
+++ b/athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
@@ -256,7 +256,9 @@ public class GcsMetadataHandler
                     .add(FILE_FORMAT, table.getParameters().get(CLASSIFICATION_GLUE_TABLE_PARAM))
                     .add(STORAGE_SPLIT_JSON, storageSplitJson);
 
+            // set partition column name and value in split
             for (FieldVector fieldVector : partitions.getFieldVectors()) {
+                fieldVector.getReader().setPosition(curPartition);
                 splitBuilder.add(fieldVector.getName(), fieldVector.getReader().readObject().toString());
             }
             splits.add(splitBuilder.build());
--- a/athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
+++ b/athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
@@ -259,8 +259,9 @@ public class GcsMetadataHandler
             // set partition column name and value in split
             for (FieldVector fieldVector : partitions.getFieldVectors()) {
                 fieldVector.getReader().setPosition(curPartition);
-                if (fieldVector.getName().equalsIgnoreCase(FILE_FORMAT) || fieldVector.getName().equalsIgnoreCase(STORAGE_SPLIT_JSON))
+                if (fieldVector.getName().equalsIgnoreCase(FILE_FORMAT) || fieldVector.getName().equalsIgnoreCase(STORAGE_SPLIT_JSON)) {
                     throw new RuntimeException("column name is same as metadata");
+                }
                 splitBuilder.add(fieldVector.getName(), fieldVector.getReader().readObject().toString());
             }
             splits.add(splitBuilder.build());
--- a/athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
+++ b/athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
@@ -48,6 +48,7 @@ import com.amazonaws.services.secretsmanager.AWSSecretsManager;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.util.VisibleForTesting;
+import org.apache.arrow.vector.FieldVector;
 import org.apache.arrow.vector.types.pojo.Schema;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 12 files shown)


athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
```diff
@@ -251,8 +253,12 @@ public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest
             LOGGER.info("MetadataHandler=GcsMetadataHandler|Method=doGetSplits|Message=StorageSplit JSON\n{}",
                     storageSplitJson);
             Split.Builder splitBuilder = Split.newBuilder(spillLocation, makeEncryptionKey())
-                    .add(CLASSIFICATION_GLUE_TABLE_PARAM, table.getParameters().get(CLASSIFICATION_GLUE_TABLE_PARAM))
+                    .add(FILE_FORMAT, table.getParameters().get(CLASSIFICATION_GLUE_TABLE_PARAM))
                     .add(STORAGE_SPLIT_JSON, storageSplitJson);
+
+            for (FieldVector fieldVector : partitions.getFieldVectors()) {
+                splitBuilder.add(fieldVector.getName(), fieldVector.getReader().readObject().toString());
+            }
```
henrymai, line 261: It looks you are putting a mapping of partition column name -> partition value as metadata onto the split, right?

However, it looks like you are just adding only the first partition value but its possible to have more than one partition value for each partition column, right?

So how does this work in that case?

hackett123:
Nit - in the future, if your commits are all accomplishing the same task, you can squash them to keep the master branch organized.

hackett123 (dismissed):

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
hackett123: getPartitionsFolderLocationUri in src/main/java/com/amazonaws/athena/connectors/gcs/common/PartitionUtil.java uses this same pattern. Is it always the case that the partition has one value? If not, then we will want to revisit that logic as well.

hackett123: Actually Henry noticed that in PartitionUtil.java you're setting the reader position, so the logic in that class is fine

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsRecordHandler.java
```diff
@@ -167,17 +169,23 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
             BlockSpiller spiller,
-            List<FieldVector> gcsFieldVectors, int rowIndex)
+            List<FieldVector> gcsFieldVectors, int rowIndex, List<Field> partitionColumns, Split split)
     {
         spiller.writeRows((Block block, int rowNum) -> {
             boolean isMatched = true;
+            // offer value for partition column
+            for (Field field : partitionColumns) {
+                isMatched &= block.offerValue(field.getName().toLowerCase(), rowNum, split.getProperty(field.getName().toLowerCase()));
+            }
```
henrymai, line 187: Are there already tests checking to make sure that the partition column value is showing up in the read result?

If not, can you add tests for this.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
Trianz-Akshay: The logic is correct. Earlier we relied on getPartitionsFolderLocationUri to setPosition() but set the explicit position in the same block of code now.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsRecordHandler.java
Trianz-Akshay: updated the test with assertion to check the no. of split and value.

athena-gcs/src/test/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandlerTest.java
```diff
@@ -319,5 +319,7 @@ public void testDoGetSplits() throws Exception
         when(table.getPartitionKeys()).thenReturn(columns);
         GetSplitsResponse response = gcsMetadataHandler.doGetSplits(blockAllocator, request);
         assertNotNull(response);
+        assertEquals(2, response.getSplits().size());
+        assertNotNull(response.getSplits().stream().findFirst().get().getProperty("year"));
```
henrymai, line 323: Why are you asserting only that its not null and only asserting the first value?

athena-gcs/src/test/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandlerTest.java
```diff
@@ -318,6 +316,46 @@ public void testDoGetSplits() throws Exception
         );
         when(table.getPartitionKeys()).thenReturn(columns);
         GetSplitsResponse response = gcsMetadataHandler.doGetSplits(blockAllocator, request);
-        assertNotNull(response);
+        assertEquals(2, response.getSplits().size());
+        assertEquals(2, response.getSplits().stream().map(split -> split.getProperties().get("year")).count());
+    }
```
henrymai, line 321: You are still only checking for the __count__ here, you are not validating that the values of "year" (which are "2000" and "2001").

athena-gcs/src/test/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandlerTest.java
```diff
@@ -318,6 +316,46 @@ public void testDoGetSplits() throws Exception
...
+        when(awsGlue.getTable(any())).thenReturn(getTableResult);
+        when(getTableResult.getTable()).thenReturn(table);
+        List<Column> columns = com.google.common.collect.ImmutableList.of(
+                createColumn("yearCol", "varchar"),
+                createColumn("monthCol", "varchar")
+        );
+        when(table.getPartitionKeys()).thenReturn(columns);
+        GetSplitsResponse response = gcsMetadataHandler.doGetSplits(blockAllocator, request);
+        assertEquals(10, response.getSplits().size());
+        assertEquals(10, response.getSplits().stream().map(split -> split.getProperties().get("year")).count());
```
henrymai, line 359: Same here, you are only still checking for the count here.
You need to validate that the values are what we expect.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
```diff
@@ -251,8 +253,14 @@ public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest
...
                     storageSplitJson);
             Split.Builder splitBuilder = Split.newBuilder(spillLocation, makeEncryptionKey())
-                    .add(CLASSIFICATION_GLUE_TABLE_PARAM, table.getParameters().get(CLASSIFICATION_GLUE_TABLE_PARAM))
+                    .add(FILE_FORMAT, table.getParameters().get(CLASSIFICATION_GLUE_TABLE_PARAM))
                     .add(STORAGE_SPLIT_JSON, storageSplitJson);
+
+            // set partition column name and value in split
+            for (FieldVector fieldVector : partitions.getFieldVectors()) {
+                fieldVector.getReader().setPosition(curPartition);
+                splitBuilder.add(fieldVector.getName(), fieldVector.getReader().readObject().toString());
```
henrymai, line 262: I would also add a check here to make sure that none of the `fieldVector.getName()`s are the same as `FILE_FORMAT`, since they are both just top level keys.

If they are the same, just throw a RuntimeException to stop us from continuing.

Trianz-Akshay: updated the code.

Issue #1047, opened 2023-02-14
Athena Opensearch ListTables fails for opensearch on VPC


pratikshetty94:
# Have set the security group and subnets same as the opensearch. But getting this: Failed to get tables names from lambda function due to com.amazonaws.services.lambda.invoke.LambdaFunctionException: Error processing request to list indices: Unable to calculate a request signature: null

# Lambda env variables
auto_discover_endpoint | true
disable_spill_encryption | false
query_scroll_timeout | 60
query_timeout_cluster | 10
query_timeout_search | 720
spill_bucket | bucket-name
spill_prefix | athena-spill

# Error:
Error processing request to list indices: Unable to calculate a request signature: null: java.lang.RuntimeException
java.lang.RuntimeException: Error processing request to list indices: Unable to calculate a request signature: null
	at com.amazonaws.athena.connectors.elasticsearch.ElasticsearchMetadataHandler.doListTables(ElasticsearchMetadataHandler.java:211)
	at com.amazonaws.athena.connector.lambda.handlers.MetadataHandler.doHandleRequest(MetadataHandler.java:250)
	at com.amazonaws.athena.connector.lambda.handlers.CompositeHandler.handleRequest(CompositeHandler.java:135)
	at com.amazonaws.athena.connector.lambda.handlers.CompositeHandler.handleRequest(CompositeHandler.java:103)
Caused by: java.lang.RuntimeException: Unable to calculate a request signature: null
	at org.elasticsearch.client.RestClient.extractAndWrapCause(RestClient.java:938)
	at org.elasticsearch.client.RestClient.performRequest(RestClient.java:300)
	at org.elasticsearch.client.RestClient.performRequest(RestClient.java:288)
	at org.elasticsearch.client.RestHighLevelClient.internalPerformRequest(RestHighLevelClient.java:1632)
	at org.elasticsearch.client.RestHighLevelClient.performRequest(RestHighLevelClient.java:1602)
	at org.elasticsearch.client.RestHighLevelClient.performRequestAndParseEntity(RestHighLevelClient.java:1572)
	at org.elasticsearch.client.IndicesClient.getAlias(IndicesClient.java:1315)
	at com.amazonaws.athena.connectors.elasticsearch.AwsRestHighLevelClient.getAliases(AwsRestHighLevelClient.java:82)
[...]

PR #1157, opened 2023-03-29
Fixed accessing domain endpoints that are in a VPC when using auto_discover_endpoint

akuzin1:
*Issue #, if available:*
[1047](https://github.com/awslabs/aws-athena-query-federation/issues/1047)
*Description of changes:*
getEndpoint() returns null if domain is in a VPC. Need to use getEndpoints() instead, which returns a Map containing key "vpc" that stores value of domain endpoint. This resolves issue of accessing and listing Tables of a domain located inside a VPC when `auto_discover_endpoint` is set to true and connector does not use `domain_mapping` environment variable.


henrymai (approved):

henrymai:
Bypassed the checks because github actions is broken right now: https://www.githubstatus.com/

Issue #292, opened 2020-11-11
[FEATURE] [QUESTION] DynamoDB connector - targeting a cross-account Glue catalog


tudor-a:
I have a question that's related to a new feature. It is related to one of the DynamoDB Connector features mentioned in the Readme file, that says:

`glue_catalog - (Optional) Can be used to target a cross-account Glue catalog. By default the connector will attempt to get metadata from its own Glue account.`

I've searched through the entire dynamodb repo for more clues, I looked into the commits and I've found no further mentions, or instructions on  how to use this functionality. Is this achieved separately from the deployment of the connector, by adding necessary IAM permissions (e.g. for Glue), or are there any other AWS services that need to be included to achieve this?
Thank you

avirtuos:
@tudor-a The value you place in that Lambda env variable will be used to determine which Glue catalog the connector uses. Can you help me understand your specific question? what are you trying to do and what part are you stuck on? I agree that the documentation could be clearer. We'll work on that.

tudor-a:
Hello @avirtuos, thanks for the feedback!
What I don't understand is, do I need to deploy other resources except DynamoDB connector to enable  access to a cross-account Glue catalog?
I currently only deploy the standard DynamoDB Connector, and I tested adding cross-account roles that would enable Athena access to a dynamodb connector lambda function from a different account. (My guess is that I've taken a wrong approach to this).

After researching other repos here I found this: https://github.com/awslabs/amazon-athena-cross-account-catalog. Do I need to deploy the Amazon Athena Cross Account Data Catalog as well to enable what I need?

> @tudor-a **The value you place in that Lambda env variable will be used to determine which Glue catalog the connector uses.** Can you help me understand your specific question? what are you trying to do and what part are you stuck on? I agree that the documentation could be clearer. We'll work on that.

I'm trying to enable access to a Glue catalog in account B by running an Athena query with a DynamoDB Connector in account A.
1. Can you explain what do you mean by "the value in that Lambda env variable will be used to determine which Glue catalog the connector uses"? What type of value should this be?
2. Also - if I specify a different Glue catalog (the one from another account) does it mean that this DynamoDB Connector can be used only to target the Glue catalog from the other account?
3. Regarding IAM policies, my guess is that IAM policies are necessary to access the cross-account glue data catalog and the target S3 bucket?

Any feedback on **the steps necessary to establish targeting a cross-account Glue catalog using the DynamoDB Connector** would be very welcome!
[...]

PR #1179, opened 2023-04-03
[DynamoDB] Add Cross-Account Functionality

hackett123:
Prior to this change, we supported using cross-account Glue data catalogs to fetch metadata for DynamoDB tables. However, we did not have any support for allowing customers to query a cross-account DynamoDB table, so this provided very little value.

With this change, customers can now query DynamoDB tables across accounts, both with and without the use of a cross account Glue data catalog. To do so, the customer adds an environment variable to their connector, `cross_account_role_arn`. They also have to perform standard IAM configurations for cross-account resource access. Specifically, the account containing the DynamoDB table (account A) adds a role which the calling account (account B) with the connector has permissions to assume, containing a policy against the table and its indices to Scan, Query, and DescribeTable. An additional ListTables permission against the account is needed as well. Account B then adds a policy to the function execution role to assume the role in account A and their queries will now work. I provide the sample setup below which will be easier to digest.

*Issue #, if available:* https://github.com/awslabs/aws-athena-query-federation/issues/292

*Description of changes:*

[Testing]
- Tested with two AWS accounts. I am providing the relevant IAM policies below with obscured table names and account ids for the DynamoDB changes. I am not providing the policies for the cross data catalog because that functionality already exists today.

### IAM policies:
For each of the policies below, "Account A" will refer to the account containing the DynamoDB table, and "Account B" will refer to the account containing the Athena DynamoDB Connector.

#### Account A - new IAM policy and role

Policy:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "dynamodb:Scan",
                "dynamodb:Query",
                "dynamodb:DescribeTable"
            ],
            "Resource": [
                "arn:aws:dynamodb:*:<ACCOUNT_A_ID>:table/<TABLE_NAME>/index/*",
                "arn:aws:dynamodb:*:<ACCOUNT_A_ID>:table/<TABLE_NAME>"
            ]
        },
        {
            "Effect": "Allow",
            "Action": [
                "dynamodb:ListTables"
            ],
            "Resource": [
                "arn:aws:dynamodb:*:<ACCOUNT_A_ID>:table/*"
            ]
        }
    ]
}
```
Role `CrossAccountQueryAndScanDdbRole` has above policy. Here is the trusted entities for the role - note that we can use the AWS account ID even though this will be called from a lambda in account B.
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::<ACCOUNT_B_ID>:root"
            },
            "Action": "sts:AssumeRole",
            "Condition": {}
        }
    ]
}
```

#### Account B - new policy attached to lambda function role
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": "sts:AssumeRole",
            "Resource": "arn:aws:iam::<ACCOUNT_A_ID>:role/CrossAccountQueryAndScanDdbRole"
        }
    ]
}
```

### Queries ran:

```
// List schemas if cross-account glue is or is not configured
show schemas in lambda:ddb

// List tables if cross-account glue is configured
show tables in lambda:ddb.databaseforcrossaccountddb

// Describe table if cross-account glue is configured
describe `lambda:ddb`.`databaseforcrossaccountddb`.`<OBSCURED>`

// Describe table if cross-account glue is not configured
describe `lambda:ddb`.`default`.`<OBSCURED>`

// Select - if cross-account glue is configured
select * from "lambda:ddb"."databaseforcrossaccountddb"."<OBSCURED>" limit 1;

// Select - if cross-account glue is not configured
select * from "lambda:ddb"."default"."<OBSCURED>" limit 1;

```

### DIFF
--- a/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
+++ b/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
@@ -19,6 +19,7 @@
  */
 package com.amazonaws.athena.connectors.dynamodb;
 
+import com.amazonaws.athena.connector.credentials.CrossAccountCredentialsProvider;
 import com.amazonaws.athena.connector.lambda.QueryStatusChecker;
 import com.amazonaws.athena.connector.lambda.ThrottlingInvoker;
 import com.amazonaws.athena.connector.lambda.data.Block;
@@ -135,7 +136,9 @@ public class DynamoDBMetadataHandler
     public DynamoDBMetadataHandler(java.util.Map<String, String> configOptions)
     {
         super(SOURCE_TYPE, configOptions);
-        this.ddbClient = AmazonDynamoDBClientBuilder.standard().build();
+        this.ddbClient = AmazonDynamoDBClientBuilder.standard()
+            .withCredentials(CrossAccountCredentialsProvider.getCrossAccountCredentialsIfPresent(configOptions, "DynamoDBMetadataHandler_CrossAccountRoleSession"))
+            .build();
         this.glueClient = getAwsGlue();
         this.invoker = ThrottlingInvoker.newDefaultBuilder(EXCEPTION_FILTER, configOptions).build();
         this.tableResolver = new DynamoDBTableResolver(invoker, ddbClient);
--- a/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBRecordHandler.java
+++ b/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBRecordHandler.java
@@ -20,6 +20,7 @@
 package com.amazonaws.athena.connectors.dynamodb;
 
 import com.amazonaws.AmazonWebServiceRequest;
+import com.amazonaws.athena.connector.credentials.CrossAccountCredentialsProvider;
 import com.amazonaws.athena.connector.lambda.QueryStatusChecker;
 import com.amazonaws.athena.connector.lambda.ThrottlingInvoker;
 import com.amazonaws.athena.connector.lambda.data.Block;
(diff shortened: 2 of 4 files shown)


athena-dynamodb/pom.xml
```diff
@@ -33,6 +33,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-java-sdk-sts</artifactId>
+            <version>1.12.440</version>
```
henrymai, line 39: For the version, use this instead:
`<version>${aws-sdk.version}</version>`

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBClientBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+                .withRoleSessionName("DynamoDBCrossAccountQuerySession");
+            AssumeRoleResult assumeRoleResult = stsClient.assumeRole(assumeRoleRequest);
+            Credentials credentials = assumeRoleResult.getCredentials();
+            BasicSessionCredentials basicSessionCredentials = new BasicSessionCredentials(credentials.getAccessKeyId(), credentials.getSecretAccessKey(), credentials.getSessionToken());
+            AWSCredentialsProvider credentialsProvider = new AWSStaticCredentialsProvider(basicSessionCredentials);
+            return AmazonDynamoDBClientBuilder
+                .standard()
+                .withCredentials(credentialsProvider)
+                .build();
+        }
```
henrymai, line 60: It looks like this code is fairly generic, I think you might be able to structure this in a way where you can just have like a generic "CrossAccountCredentialProvider" in the SDK (which is basically this method that takes in a configOptions map and instead it only returns an AwsCredentialsProvider).

Then the environment variable this generic "CrossAccountCredentialProvider" would be a generic "CrossAccountRoleARN" environment variable.

In the case where that environment variable isn't defined, you can just return: `DefaultAWSCredentialsProviderChain.getInstance()`

Then all the connectors would just call this method blindly and use whatever AwsCredentialsProvider they get back blindly.

hackett123: That's a good idea, let me abstract just the credentials provider out. Callers will just provide the config options map and a name for the role session.

athena-dynamodb/pom.xml
hackett123: ack

henrymai (approved):

Issue #941, opened 2023-01-04
[BUG] PostgreSQL Monetary types are not queryable


danielcmessias:
**Describe the bug**

If you have a [Monetary](https://www.postgresql.org/docs/current/datatype-money.html) column in Postgres and try to query in Athena it will fail with the exception

```
GENERIC_USER_ERROR: Encountered an exception[java.lang.RuntimeException] from your LambdaFunction[...] executed in context[...] with message[org.postgresql.util.PSQLException: Bad value for type double : 25,000.00]
```

**To Reproduce**

1. Create a table in Postgres with a column of data type 'money'
2. Add a row with some value such as '$25,000.00'
3. Attempt to query in Athena

**Expected behavior**

You receive the column data, casted to an appropriate Athena type (double?)
 

**Screenshots / Exceptions / Errors**
If applicable, add screenshots, exception stack traces, or error messages to help explain your problem.  Stack traces for exceptions thrown by your Connector can be found in CloudWatch Logs under the 'aws/lambda/<Lambda_function_name>' log group.

**Connector Details (please complete the following information):**

serverlessrepo:applicationId | arn:aws:serverlessrepo:us-east-1:292517598671:applications/AthenaPostgreSQLConnector
serverlessrepo:semanticVersion | 2022.39.1

**Additional context**

akuzin1:
Thank you for reaching out! The bug fix should be included in latest release, so please upgrade to version v2023.14.1 for PostgreSQL connector. Feel free to reopen issue if anything else is encountered.

danielcmessias:
@akuzin1 Thanks!!

PR #1183, opened 2023-04-06
Added Money type support for Postgres Connector

akuzin1:
Fixes parsing Money Type in Postgres.
Issue #941

### DIFF
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
@@ -200,7 +200,6 @@ public abstract class JdbcRecordHandler
     protected Extractor makeExtractor(Field field, ResultSet resultSet, Map<String, String> partitionValues)
     {
         Types.MinorType fieldType = Types.getMinorTypeForArrowType(field.getType());
-
         final String fieldName = field.getName();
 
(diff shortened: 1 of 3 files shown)


henrymai (approved):

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
```diff
@@ -252,7 +251,14 @@ protected Extractor makeExtractor(Field field, ResultSet resultSet, Map<String,
             case FLOAT8:
                 return (Float8Extractor) (Object context, NullableFloat8Holder dst) ->
                 {
-                    dst.value = resultSet.getDouble(fieldName);
+                    try {
+                        dst.value = resultSet.getDouble(fieldName);
+                    }
+                    catch (java.sql.SQLException ex) {
```
hackett123, line 257: Is the exception thrown the generic SqlException? or can we make this more specific?

henrymai: The actual exception thrown is from Postgres and it extends from SqlException, but since we are in athena-jdbc, we don't want to introduce dependencies in postgres, so we catch SqlException instead.

hackett123: I'm fine with this then, thanks for clarifying.

PR #1188, opened 2023-04-08
big query return athena catalog name instead of project name from gcp

nauy2697:
big query return athena catalog name instead of project name from big query for GetTableResponse.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
+++ b/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
@@ -163,7 +163,7 @@ public class BigQueryMetadataHandler
         final Schema tableSchema = getSchema(projectName, getTableRequest.getTableName().getSchemaName(),
                 getTableRequest.getTableName().getTableName());
         // TODO: Do we actually need to lowercase here?
-        return new GetTableResponse(projectName.toLowerCase(), getTableRequest.getTableName(), tableSchema);
+        return new GetTableResponse(getTableRequest.getCatalogName().toLowerCase(), getTableRequest.getTableName(), tableSchema);
     }
 
     /**


athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
```diff
@@ -163,7 +163,7 @@ public GetTableResponse doGetTable(BlockAllocator blockAllocator, GetTableReques
         final Schema tableSchema = getSchema(projectName, getTableRequest.getTableName().getSchemaName(),
                 getTableRequest.getTableName().getTableName());
         // TODO: Do we actually need to lowercase here?
-        return new GetTableResponse(projectName.toLowerCase(), getTableRequest.getTableName(), tableSchema);
+        return new GetTableResponse(getTableRequest.getCatalogName().toLowerCase(), getTableRequest.getTableName(), tableSchema);
```
henrymai, line 166: projectName is already supposed to be the `getTableRequest.getCatalogName()` unless its overridden by the environment variable. See the implementation of `BigQueryUtils.getProjectName`: https://github.com/awslabs/aws-athena-query-federation/blob/5bd4c7935ac5fb9c7fcb520e176225059de53964/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryUtils.java#L104

nauy2697: Tested with big query, select works. All other response such as `ListTablesResponse` and `ListSchemasResponse` used athena catalog name instead. Not sure why only specific GetTableResponse used projectID. In addition, the project id set in the environment variable are used to retrieve data from GCP. 

But i will reach out to Trianz to double check.

henrymai: > projectName always take GCP_PROJECT_ID that we set in environment variable first

That's what I said already: 
"projectName is already supposed to be the getTableRequest.getCatalogName() unless its overridden by the environment variable."

PR #1200, opened 2023-04-12
[fix]athena-google-bigquery: return athena catalog name instead of project name from gbq.

Jithendar12:
INC00001110

*Description of changes:*

### DIFF
--- a/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryUtils.java
+++ b/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryUtils.java
@@ -97,13 +97,12 @@ public class BigQueryUtils
      * The Lambda environment variables are first inspected and if it does not exist, then we take it from the catalog
      * name in the request.
      *
-     * @param catalogName The catalogName from the request that is passed in from the Athena Connector framework.
      * @param configOptions The configOptions contains the fallback value for the project id.
      * @return The project name.
      */
-    static String getProjectName(String catalogName, java.util.Map<String, String> configOptions)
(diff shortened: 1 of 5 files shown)


athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryUtils.java
```diff
@@ -97,13 +97,12 @@ public static String getEnvBigQueryCredsSmId(java.util.Map<String, String> confi
...
      *
-     * @param catalogName The catalogName from the request that is passed in from the Athena Connector framework.
      * @param configOptions The configOptions contains the fallback value for the project id.
      * @return The project name.
      */
-    static String getProjectName(String catalogName, java.util.Map<String, String> configOptions)
+    static String getProjectName(java.util.Map<String, String> configOptions)
     {
-        return configOptions.getOrDefault(BigQueryConstants.GCP_PROJECT_ID, catalogName);
+        return configOptions.get(BigQueryConstants.GCP_PROJECT_ID);
```
henrymai, line 105: It seems like this method can just be entirely deleted right?

Because every time we're using this, we're just passing the projectName into BigQueryUtils.getBigQueryClient(..., configOptions). So `BigQueryUtils.getBigQueryClient()` can just grab the projectName itself from the configOptions.

So this method + all the lines that call this method can just be deleted.

I didn't comprehensively check this, but I think this is the case, right?

Jithendar12: Hi henry, yes we can delete getProjectName() method completely. Commited the change

nauy2697 (approved):

Issue #792, opened 2022-10-09
[BUG] Using upper key database or table names  with connector Timestream fails


andasv:
**Describe the bug**
When using upper key database or table names  with connector Timestream fails:

GENERIC_USER_ERROR: Encountered an exception[com.amazonaws.services.timestreamquery.model.ValidationException] from your LambdaFunction[arn:aws:lambda:us-east-1:978572884825:function:timestreamathenaconnector] executed in context[retrieving meta-data] with message[The table 'iotmulti' does not exist in database 'testdb' at line 1:1 (Service: AmazonTimestreamQuery; Status Code: 400; Error Code: ValidationException; Request ID: BTC575JOI5ABUT6PB7PEPKVKY4; Proxy: null)]
This query ran against the "testdb" database, unless qualified by the query. 

**To Reproduce**
1. Create sample query for IoT Data in Amazon Timestream
2. Add Athena connector
3. Run query
SELECT * FROM "testdb"."IoTMulti" limit 10;

**Expected behavior**
Athena shall show the content of IoTMulti table

**Connector Details (please complete the following information):**
 - Version: 2022.39.1
 - Name timestream
 - Athena Query IDs BTC575JOI5ABUT6PB7PEPKVKY4

hackett123:
Hi [andasv](https://github.com/andasv), thanks for being patient here. I'm taking a look at this issue.

Can you confirm for us whether or not you are using Glue for your table metadata?

hackett123:
I got a chance to root cause this issue. The problem is that the Athena query engine, Trino, follows Hive naming semantics and will lower-case the schema name and table name in a query. (AWS Glue also follows this convention, so my earlier question is not important.)

What this means is that if you run a query such as `SELECT * FROM MySchema.MyTable`, when Trino receives the query, it converts your datasource name to `myschema.mytable`. When we then go to invoke your connector, we pass in `myschema.mytable`, which does not match the Timestream db/table name.

This can lead to confusing behavior, because when you run a DDL query such as `DESCRIBE MySchema.MyTable`, we will just directly invoke your lambda and pass in `MySchema.MyTable`, which will succeed, leading to confusion.

**The simplest workaround here is for users to make their timestream db and table names lowercase to conform with the current Hive-compliant semantics of the query engine.**

hackett123:
[...]

PR #1201, opened 2023-04-14
[Timestream] Add case-insensitive lookup to allow queries against cased dbs/tables to succeed

hackett123:
*Issue #, if available:* [#792](https://github.com/awslabs/aws-athena-query-federation/issues/792)

*Description of changes:*

Because Trino follows Hive naming conventions, where schema names and table names are lower-cased and are case-insensitive, DML queries against our connectors will receive lower-cased versions of schema and table names. This will lead to query failures if the Timestream schema/table are not both in lowercase format.

Example: Database name is "MixedCaseDb" and table name is "tableWithMixedCase".

`SELECT * FROM "lambda:timestream"."MixedCaseDb"."tableWithMixedCase"` will be sent down to Timestream with the schema name `mixedcasedb` and table name `tablewithmixedcase`, which will then fail to resolve because Timestream is case-sensitive.

This PR adds logic to do a lookup of a case-insensitive match for the schema and table name against the Timestream datasource if the initial attempt to fetch the table definition fails.

We also pull out the functionality to make the actual Timestream API calls so that we can parse the results one page at a time instead of after reading all pages.

Testing was done against databases with mixed cases, tables with mixed cases, and db/table combos that were all lowercase.

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/util/PaginatedRequestIterator.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/util/PaginatedRequestIterator.java
@@ -1,0 +1,73 @@
+/*-
+ * #%L
+ * Amazon Athena Query Federation SDK
+ * %%
+ * Copyright (C) 2019 - 2023 Amazon Web Services
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


hackett123:
@henrymai has asked that we block this until we can refactor the searches for a case-insensitive schema/table to halt upon hitting a match, rather than needing to possible pull in N pages of schema names or table names.

So I'll leave the PR up but will send out a new revision during working hours next week.

hackett123:
Updated PR to reflect asks from @henrymai . Refactor is based on athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskMetadataHandler.java

Issue #511, opened 2021-10-28
[BUG] Elasticsearch connector unable to list/query datastreams


npow:
**Describe the bug**
1. Show tables does not return any datastreams: ```show tables in `lambda:elasticsearch`.es```
2. When trying to query datastream directly, runtime exception is thrown: ```describe `lambda:elasticsearch`.es.pages```

**To Reproduce**
1. run `show tables` on an ES instance with datastreams
2. try to query the datastream directly

**Expected behavior**
Expect datastreams to show up and behave like regular tables

**Screenshots / Exceptions / Errors**
```
java.lang.RuntimeException: com.amazonaws.services.lambda.invoke.LambdaFunctionException: Error processing request to map index (pages): null
```

**Connector Details (please complete the following information):**
 - Version: [e.g. 2020.04.01 - 2021.42.1
 - Name [e.g. redis, hbase, jdbc]: elasticsearch
 - Athena Query IDs [if applicable]:  d38b8bfe-81eb-4fd0-9883-2381941360f5

**Additional context**
Elasticsearch datastreams: https://www.elastic.co/guide/en/elasticsearch/reference/master/data-streams.html

chngpe:
Hi,

We have added support for data stream on ES connectors, please use our latest version >=2023.18.1

PR #1240, opened 2023-05-01
add data stream support for elasticsearch

chngpe:
Add support for Data Stream for elastic search.

Data Stream contains multiple indices, see following notes for more information.

https://opensearch.org/docs/1.1/opensearch/data-streams/

https://opster.com/guides/opensearch/opensearch-machine-learning/opensearch-data-streams/#How-to-create-data-streams-in-OpenSearch

*Issue #, if available:*
https://github.com/awslabs/aws-athena-query-federation/issues/511

### DIFF
--- a/athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
+++ b/athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
@@ -25,7 +25,6 @@ import com.amazonaws.athena.connector.lambda.data.BlockWriter;
 import com.amazonaws.athena.connector.lambda.data.SchemaBuilder;
 import com.amazonaws.athena.connector.lambda.domain.Split;
 import com.amazonaws.athena.connector.lambda.domain.TableName;
-import com.amazonaws.athena.connector.lambda.domain.spill.SpillLocation;
 import com.amazonaws.athena.connector.lambda.handlers.GlueMetadataHandler;
 import com.amazonaws.athena.connector.lambda.metadata.GetSplitsRequest;
 import com.amazonaws.athena.connector.lambda.metadata.GetSplitsResponse;
@@ -45,16 +44,21 @@ import com.google.common.collect.ImmutableMap;
 import org.apache.arrow.util.VisibleForTesting;
 import org.apache.arrow.vector.types.pojo.Field;
 import org.apache.arrow.vector.types.pojo.Schema;
+import org.elasticsearch.client.RequestOptions;
+import org.elasticsearch.client.indices.GetDataStreamRequest;
+import org.elasticsearch.client.indices.GetIndexRequest;
+import org.elasticsearch.client.indices.GetIndexResponse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * This class is responsible for providing Athena with metadata about the domain (aka databases), indices, contained
@@ -96,6 +100,8 @@ public class ElasticsearchMetadataHandler
      * from the Elasticsearch instance (e.g. "_shards:5" - retrieve shard number 5).
      */
     private static final String SHARD_VALUE = "_shards:";
+
+    protected static final String INDEX_KEY = "index";
 
     private final AWSGlue awsGlue;
     private final AwsRestHighLevelClientFactory clientFactory;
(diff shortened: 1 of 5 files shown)


athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -187,6 +195,14 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
 
                     indices.add(new TableName(request.getSchemaName(), index));
                 }
+
```
henrymai, line 198: github won't let me select this whole function, but you can rewrite and simplify this entire method:

```
    @Override
    public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesRequest request)
            throws RuntimeException
    {
        logger.debug("doListTables: enter - " + request);
        String endpoint = getDomainEndpoint(request.getSchemaName());
        AwsRestHighLevelClient client = clientFactory.getOrCreateClient(endpoint);
        try {
            Stream<String> indices = client.getAliases().stream().filter(index -> !index.startsWith("."));
            // handle data stream, 1 data stream can contains multiple indices which start with ".ds-xxxxxxxxxx"
            Stream<String> dataStreams = client.indices()
                .getDataStream(new GetDataStreamRequest("*"), RequestOptions.DEFAULT)
                .getDataStreams()
                .stream()
                .map(dataStream -> dataStream.getName());
            List<TableName> tableNames = Stream.concat(indices, dataStreams)
                .map(tableName -> new TableName(request.getSchemaName(), tableName))
                .collect(Collectors.toList());
            return new ListTablesResponse(request.getCatalogName(), tableNames, null);
        }
        catch (IOException error) {
            throw new RuntimeException("Error retrieving indices: " + error.getMessage(), error);
        }
    }
```

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -292,30 +308,35 @@ public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest
...
-                            .of(domain, endpoint, SHARD_KEY, SHARD_VALUE + shardId.toString())));
+                for (String index : indices) {
+                    Set<Integer> shardIds = client.getShardIds(index, queryTimeout);
+                    for (Integer shardId : shardIds) {
+                        // Every split must have a unique location if we wish to spill to avoid failures
+                        SpillLocation spillLocation = makeSpillLocation(request);
+                        // Create a new split (added to the splits set) that includes the domain and endpoint.
+                        // In addition, we retain shard and index information (to be used later by the Record Handler)
+                        splits.add(new Split(spillLocation, makeEncryptionKey(), ImmutableMap
+                                .of(domain, endpoint, SHARD_KEY, SHARD_VALUE + shardId.toString(), INDEX_KEY, index)));
```
henrymai, line 329: Rewrite this function using Streams. It will be simpler and avoids the ITM anti-pattern.

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -172,21 +178,25 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
     {
         logger.debug("doListTables: enter - " + request);
 
-        List<TableName> indices = new ArrayList<>();
-
         try {
             String endpoint = getDomainEndpoint(request.getSchemaName());
             AwsRestHighLevelClient client = clientFactory.getOrCreateClient(endpoint);
             try {
```
henrymai, line 184: Why do you have the nested trys here?

If you're worried about logging, just log before something critical is done and then just let the exception bubble up as it does without catching it (other than checked exceptions that have to be rethrown).

This way, if you think about it, when the exception happens and it errors out, we have to look at the logs anyway, and then we'll just see in the logs the last log of what we were trying to do before the exception happened.

Additionally, even without that log message, we know exactly where to look because the exception will contain the appropriate stack trace already.

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -288,37 +296,41 @@ public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest
...
 
-        return new GetSplitsResponse(request.getCatalogName(), splits);
+    private Set<Integer> getShardsIDsFromES(AwsRestHighLevelClient client, String catalogName, String index)
+    {
+        try {
+            return client.getShardIds(index, queryTimeout);
+        }
+        catch (RuntimeException | IOException error) {
+            throw new RuntimeException(String.format("Error trying to generate splits for catalog: %s, index: %s, error message: %s", catalogName, index, error.getMessage()), error);
+        }
```
henrymai, line 333: Why are you try catching and rethrowing here when the caller is again going to try catch?

You can just declare that this method throws IOException and then just let the caller worry about it.

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -288,37 +296,41 @@ public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest
...
+                    .collect(Collectors.toSet());
+
+            return new GetSplitsResponse(request.getCatalogName(), splits);
         }
-        catch (RuntimeException error) {
+        catch (RuntimeException | IOException error) {
             throw new RuntimeException("Error trying to generate splits for index (" +
-                    index + "): " + error.getMessage(), error);
+                    request.getCatalogName() + "): " + error.getMessage(), error);
         }
```
henrymai, line 323: Same comment here about the exception handling.

henrymai (approved):

PR #1241, opened 2023-05-02
Even distribution of records across splits in Athena Snowflake Connector

ankita-yadav12:
*Issue # INC00001108, Receiving uneven split distribution when using Athena Snowflake connector.

*Description of changes:*
The modified partition logic in Snowflake Connector will have even record distribution across each partition.
It will have maximum 50 partitions and number of records in each partition is decided by dividing total number of records by 50.


athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -187,44 +182,25 @@ public void getPartitions(BlockWriter blockWriter, GetTableLayoutRequest getTabl
                     totalRecordCount = rs.getLong(1);
                 }
                 if (totalRecordCount > 0) {
-                    // if number of partitions are more than defined limit "MAX_PARTITION_COUNT"
-                    // it will do maximum 50 partitions,49 partitions will have 500000 records each and last partition will have the remaining number of records.
-                    double limitValue = totalRecordCount / PARTITION_RECORD_COUNT;
+                    long partitionRecordCount = (long) (Math.ceil(totalRecordCount / MAX_PARTITION_COUNT));
+                    LOGGER.info(" Total Page  Count" +  partitionRecordCount);
+                    double limitValue = totalRecordCount / partitionRecordCount;
                     double limit = (int) Math.ceil(limitValue);
```
akuzin1, line 188: What is the purpose of `limit` and `limitValue`? Wouldn't limit always be `MAX_PARTITION_COUNT(or 50)`

ankita-yadav12: Yes Akuzin,you are right.For most of the scenarios its 50 i.e. MAX_PARTITION_COUNT.
But for some cases,like when the records are less than 50,limit will not be equal to MAX_PARTITION_COUNT.
We discussed this point internally also and found that its not good to create unnecessary splits even for smaller number of records.
We were thinking to update the logic,if the number of records are less than or equal to 10000,it should be returned in a single split.
We need your opinion on this.

long partitionRecordCount = (long) (Math.ceil(totalRecordCount / MAX_PARTITION_COUNT));
double limitValue = (partitionRecordCount <= 200) ? (long) totalRecordCount : partitionRecordCount;
double limit = (int) Math.ceil(totalRecordCount / limitValue);

akuzin1: Yes I agree, I think that would be the correct approach. Thank you

ankita-yadav12:
Hi Akuzin, I have modified the logic to return single split in case of records less than 10000 and 50 even splits for large number of records. Please review it and let me know if any changes required. Thanks

ankita-yadav12:
Hi Akuzin, I have modified the logic to return single split in case of records less than 10000 and 50 even splits for large number of records. Please review it and let me know if any changes required.
Thanks

akuzin1 (approved):

PR #1242, opened 2023-05-02
adding equals, toString, and hashCode to GetDataSourceCapabilitiesReq…

burhan94:
Need to add these methods so that the object can get cached in the engine.

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/GetDataSourceCapabilitiesRequest.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/GetDataSourceCapabilitiesRequest.java
@@ -22,6 +22,7 @@ package com.amazonaws.athena.connector.lambda.metadata;
 import com.amazonaws.athena.connector.lambda.security.FederatedIdentity;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Objects;
 
 public class GetDataSourceCapabilitiesRequest
         extends MetadataRequest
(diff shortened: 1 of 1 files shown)


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/GetDataSourceCapabilitiesRequest.java
```diff
@@ -46,4 +47,38 @@ public void close()
...
+    public boolean equals(Object o)
+    {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        GetDataSourceCapabilitiesRequest that = (GetDataSourceCapabilitiesRequest) o;
```
henrymai, line 71: nit:
Just rename the parameter to the equals method from `o` to `that` and then you can get rid of this line and all the other lines using `o` also automatically improve.

henrymai: Here's an example:

```
jshell> class A {
   ...>     public String asdf;
   ...>
   ...>     public boolean equals(A that) {
   ...>         return this.asdf == that.asdf;
   ...>     }
   ...>
   ...> }
|  created class A


jshell> var asdf1 = new A(); asdf1.asdf = "aaaa";
asdf1 ==> A@306a30c7
$4 ==> "aaaa"

jshell> var asdf2 = new A(); asdf2.asdf = "bbbb";
asdf2 ==> A@255316f2
$6 ==> "bbbb"

jshell> asdf1.equals(asdf2);
$7 ==> false
```

henrymai: Sorry actually, this fails:

```
jshell> var m = Map.of(asdf1, "1", asdf2, "2", asdf3, "3");
m ==> {A@61=3, A@62=2, A@61=1}
```

Because Map ends up using the Object version of equals.

So your version is fine the way it is.

PR #1258, opened 2023-05-08
gracefully exit get data stream name for ES version that does not support data stream

chngpe:
gracefully exit for AWS ES.
AWS's version of ES doesn't support data stream, it is a feature only available in the Elasticsearch distributed by Elastic itself, licensed under the Elastic License.

See for more info. https://stackoverflow.com/questions/68006918/aws-elasticsearch-service-and-datastream

### DIFF
--- a/athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
+++ b/athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
@@ -186,10 +186,19 @@ public class ElasticsearchMetadataHandler
                 .filter(index -> !index.startsWith("."));
 
         // get all data streams from ES, 1 data stream can contains multiple indices which start with ".ds-xxxxxxxxxx"
-        Stream<String> dataStreams = client.indices().getDataStream(new GetDataStreamRequest("*"), RequestOptions.DEFAULT).getDataStreams()
-                .stream()
-                .map(dataStream -> dataStream.getName());
-
+        Stream<String> dataStreams;
+        try {
+            dataStreams = client.indices().getDataStream(new GetDataStreamRequest("*"), RequestOptions.DEFAULT).getDataStreams()
+                    .stream()
+                    .map(dataStream -> dataStream.getName());
+        }
+        // gracefully exit for AWS ES.
+        // AWS's version of ES doesn't support data stream, it is a feature only available in the Elasticsearch distributed by Elastic itself, licensed under the Elastic License.
+        // https://stackoverflow.com/questions/68006918/aws-elasticsearch-service-and-datastream
+        catch (Exception ex) {
+            logger.warn(String.format("Unable to retrieve datastream or cluster version not support data stream, ignore datastream, message: %s", ex.getMessage()), ex);
+            dataStreams = Stream.empty();
+        }
         //combine two different data sources and create tables
         List<TableName> tableNames = Stream.concat(indicesStream, dataStreams)
                 .map(tableName -> new TableName(request.getSchemaName(), tableName))


athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -186,10 +186,19 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
...
+        try {
+            dataStreams = client.indices().getDataStream(new GetDataStreamRequest("*"), RequestOptions.DEFAULT).getDataStreams()
+                    .stream()
+                    .map(dataStream -> dataStream.getName());
+        }
+        // gracefully exit for AWS ES.
+        // AWS's version of ES doesn't support data stream, it is a feature only available in the Elasticsearch distributed by Elastic itself, licensed under the Elastic License.
+        // https://stackoverflow.com/questions/68006918/aws-elasticsearch-service-and-datastream
+        catch (Exception ex) {
+            logger.warn(String.format("Unable to retrieve datastream or cluster version not support data stream, ignore datastream, message: %s", ex.getMessage()), ex);
```
hackett123, line 199: You should pass in strings to the logger the way we do it everywhere else:
```
logger.warn("Exception retrieving data stream. Message: {}", ex.getMessage());
```

We can use the {} as a placeholder and then just pass in the args after. You could also just do:
```
logger.warn("Exception retrieving data stream: {}", ex);
```

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -186,10 +186,19 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
...
+            dataStreams = Stream.empty();
```
hackett123, line 200: wondering if we should just init dataStreams to `Stream.empty()` on line 189 so you don't have to do this in the catch block. It would be best if we could just do this all in one line but I don't think there's a good way to do that. Only alternative would be to put all this data stream logic in a function and then just grab it on line 203 with getDataStreams() and let it handle returning either an empty stream or non-empty stream. In the interest of minimizing "how much" our functions do, I'm leaning towards that.

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -296,6 +294,29 @@ public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest
         return new GetSplitsResponse(request.getCatalogName(), splits);
     }
 
+    /**
+     * Get DataStream names from ES/OS cluster if supported.
+     * return empty if not supported.
+     * Notes: AWS's version of ES doesn't support data stream, it is a feature only available in the Elasticsearch distributed by Elastic itself, licensed under the Elastic License.
+     * @param client
+     * @return
```
hackett123, line 302: You have an empty return annotation here. Can just delete lines 301 and 302 because you already described the method purpose on 298.

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -296,6 +294,29 @@ public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest
...
+    private Stream<String> getDataStreamNamesFromClient(AwsRestHighLevelClient client)
+    {
+        try {
+            return client.indices().getDataStream(new GetDataStreamRequest("*"), RequestOptions.DEFAULT).getDataStreams()
+                    .stream()
+                    .map(dataStream -> dataStream.getName());
+        }
+        // gracefully exit for AWS ES.
+        // AWS's version of ES doesn't support data stream, it is a feature only available in the Elasticsearch distributed by Elastic itself, licensed under the Elastic License.
+        // https://stackoverflow.com/questions/68006918/aws-elasticsearch-service-and-datastream
```
hackett123, line 313: you don't need these comments because you already explained this in your method header.

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -296,6 +294,29 @@ public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest
...
     }
 
+    /**
+     * Get DataStream names from ES/OS cluster if supported.
+     * return empty if not supported.
+     * Notes: AWS's version of ES doesn't support data stream, it is a feature only available in the Elasticsearch distributed by Elastic itself, licensed under the Elastic License.
+     * @param client
+     * @return
+     */
+    private Stream<String> getDataStreamNamesFromClient(AwsRestHighLevelClient client)
```
hackett123, line 304: nit: I would just call this `getDataStreamNames`

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -186,9 +186,7 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
                 .filter(index -> !index.startsWith("."));
 
         // get all data streams from ES, 1 data stream can contains multiple indices which start with ".ds-xxxxxxxxxx"
-        Stream<String> dataStreams = client.indices().getDataStream(new GetDataStreamRequest("*"), RequestOptions.DEFAULT).getDataStreams()
-                .stream()
-                .map(dataStream -> dataStream.getName());
+        Stream<String> dataStreams = getDataStreamNamesFromClient(client);
 
         //combine two different data sources and create tables
         List<TableName> tableNames = Stream.concat(indicesStream, dataStreams)
```
hackett123, line 192: If you aren't using it in more than this one line, omit 189 and then just replace this line with `List<TableName> tableNames = Stream.concat(indicesStream, getDataStreamNames(client))`

athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchMetadataHandler.java
```diff
@@ -296,6 +291,25 @@ public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest
...
+    private Stream<String> getDataStreamNames(AwsRestHighLevelClient client)
+    {
+        try {
+            return client.indices().getDataStream(new GetDataStreamRequest("*"), RequestOptions.DEFAULT).getDataStreams()
+                    .stream()
+                    .map(dataStream -> dataStream.getName());
+        }
+        // gracefully exit for exception or non-support data stream.
+        catch (Exception ex) {
+            logger.warn("getDataStreamNamesFromClient: Unable to retrieve datastream or cluster version not support data stream, ignore datastream, message: {}.", ex.getMessage(), ex);
```
hackett123, line 308: If you want to pass in the exception object itself, then it's not worthwhile to pass in the message separately. when slf4j logs the warning, it will log the exception's message and stack trace. So you should just write this:

Notice that I also updated the method name you write (which. you updated in the method header but not here)
```
logger.warn("getDataStreamNames: Unable to receive data stream, or the cluster version does not support data stream.", ex);
```

Also, what is the exact exception that's thrown? Can we make this catch block more specific?

hackett123 (approved):

PR #1262, opened 2023-05-13
LocalKeyFactory: Don't block forever

henrymai:
When we are using SecureRandom.getInstanceStrong(), it may block forever when reading from /dev/random on some platforms where it doesn't have enough entropy.

In those situations, instead of blocking forever, we will timeout now after 1 second and throw an exception.

Note that this might still happen even if we just use `new SecureRandom()` because it may read from /dev/random initially for as the `seedFile` before it switches to using /dev/urandom.

See: https://github.com/openjdk/jdk/blob/master/src/java.base/unix/classes/sun/security/provider/NativePRNG.java#L157


chngpe (approved):

henrymai:
Note that if `/dev/random` blocks forever, the thread running the future will also block forever since the underling implementation of `SecureRandom` uses a `FileInputStream` to read.

However, this is fine, since we should not expect the application to continue to function after we timeout and throw the RuntimeException in the main thread anyway.

Issue #61, opened 2019-11-22
[FEATURE] JDBC connector only works with lowercase table names.


abhijeetvg:
Implement something similar to `DynamoDBTableResolver` to resolve case sensitive table names and also column names. We may be able to handle column names but using schema meta-data so that we dont really need a table name cache like we do with other connectors.

Olliepop:
This issue also affects database names with uppercase chars. 

For example, querying database "ExecutionerDb" we get the error:

> GENERIC_USER_ERROR: Encountered an exception[java.lang.RuntimeException] from your LambdaFunction[arn:aws:lambda:********:**************:function:athena-jdbc-connector] executed in context[retrieving meta-data] with message[Could not find table in executionerdb]

When we create a new database with a lowercase name this is not an error.

Likely affected by the case insensitive search argument: https://github.com/awslabs/aws-athena-query-federation/blob/20bfdd4b2cb3dcef4337dee469909faec10bf41b/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/resolver/DynamoDBTableResolver.java#L153

milesgranger:
Also suffering from this, complete show stopper for us without significant rework of existing tables/workflows. 
ref https://github.com/awslabs/aws-athena-query-federation/issues/281

jbgh2:
Has this made it on to the roadmap?

codek:
We're also facing this problem.  Our database is built and managed by Outsystems (low code app development tool) and by default the tables are init cap'd.  I'd imagine it's a relatively common thing?  (We're also using RDS)

bdstein33:
This is a huge blocker for us as well.  Super bummed about this.

dbjellan:
This is a really big blocker for us, please prioritize getting the fix merged

macohen:
@aimethed can you link to the individual connector issues where we are addressing casing here, please? We'll track them there since this connector is no longer used on its own.

aimethed:
Here are the PRs for the JDBC connectors we have implemented some form of case handling:
* https://github.com/awslabs/aws-athena-query-federation/pull/2487
* https://github.com/awslabs/aws-athena-query-federation/pull/2387

If there are any other connectors that should have additional casing handling (ex: case-insensitive search), please open a new issue for those and we will address them.

PR #1289, opened 2023-05-24
Case insensitive table name search in JDBC connectors

dbadaya1:
*Issue #, if available:* https://github.com/awslabs/aws-athena-query-federation/issues/61

*Description of changes:* In doGetTable method, if a table is not found in the database, then do a lookup for all tables in the database, and then do a local case-insensitive search for table name. 
This is similar to how DynamoDB connector does case-insensitive search https://github.com/awslabs/aws-athena-query-federation/blob/b99bba84302a9299f5d53e3a4da1cb0d9f67ab2e/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/resolver/DynamoDBTableResolver.java#L148-L172

Note that if database contains a table with all lowercase letters, then that table will always be returned even if there are other tables with uppercase letters. This is because lowercase table name will be returned in the first call to getTable, so code flow will never reach where we lookup all tables in the database. This is existing behavior even without this change.  For example, if the database contains users, Users and USers, then table *users* will be returned for all possible case variations of *users* in the query. 

If the query is for table Users, but database only contains tables uSers, then uSers will be returned as we don't know the actual case of the table specified in the query. 

### DIFF
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
@@ -238,7 +238,7 @@ public abstract class JdbcMetadataHandler
         }
     }
 
-    private TableName caseInsensitiveTableSearch(Connection connection, final String databaseName,
+    protected TableName caseInsensitiveTableSearch(Connection connection, final String databaseName,
                                                      final String tableName) throws Exception
     {
         List<TableName> tables = listTables(connection, databaseName)
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
@@ -219,45 +219,29 @@ public abstract class JdbcMetadataHandler
     {
         try (Connection connection = jdbcConnectionFactory.getConnection(getCredentialProvider())) {
             Schema partitionSchema = getPartitionSchema(getTableRequest.getCatalogName());
-            Optional<Schema> schema = getSchema(connection, getTableRequest.getTableName(), partitionSchema);
-            if (schema.isPresent()) {
-                return new GetTableResponse(getTableRequest.getCatalogName(), getTableRequest.getTableName(), schema.get(),
+            TableName caseInsensitiveTableMatch = caseInsensitiveTableSearch(connection, getTableRequest.getTableName().getSchemaName(),
+                    getTableRequest.getTableName().getTableName());
+            Schema caseInsensitiveSchemaMatch = getSchema(connection, caseInsensitiveTableMatch, partitionSchema);
+            return new GetTableResponse(getTableRequest.getCatalogName(), caseInsensitiveTableMatch, caseInsensitiveSchemaMatch,
                         partitionSchema.getFields().stream().map(Field::getName).collect(Collectors.toSet()));
-            }
-            else {
-                LOGGER.info("Table {} not found.  Falling back to case insensitive search.", getTableRequest.getTableName());
-
-                TableName caseInsensitiveTableMatch = caseInsensitiveTableSearch(connection, getTableRequest.getTableName().getSchemaName(),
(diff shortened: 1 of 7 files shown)


dbadaya1:
Just realized that JdbcMetadataHandler::doGetTable method is overridden by some of the implementations, so those will need to be updated as well. At-least Mysql, Postgres and Redshift connectors don't override this method, so the change will work for them. Will discuss with the team if we need to prioritize all other connectors right away or if it can wait.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
```diff
@@ -218,13 +219,46 @@ public GetTableResponse doGetTable(final BlockAllocator blockAllocator, final Ge
...
+            }
         }
     }
 
-    private Schema getSchema(Connection jdbcConnection, TableName tableName, Schema partitionSchema)
-            throws SQLException
+    private TableName caseInsensitiveTableSearch(Connection connection, final String databaseName,
+                                                     final String tableName) throws Exception
+    {
+        List<TableName> tables = listTables(connection, databaseName)
```
hackett123, line 244: For other readers, we can't benefit from pagination here because we can't put LIMIT on the tables we return from the jdbc query, so this is actually the right way to do this

hackett123:
So DJ and I chatted offline about potentially resolving the table name in another way - we can do something like this instead of fetching all the tables, so we can just push down the resolution to the underlying db. Here's how we could do it for mysql:

```
SELECT table_name, lower(table_name)
FROM information_schema.tables
WHERE lower(table_name) = 'inputfromathena'
```
and then `table_name` is the resolved table name.

this would help our runtime a lot

henrymai:
> Note that this will only improve the performance for case when the table name in the database contains upper case letters, i.e. first call to database getTable returns empty result, so we do a case-insensitive search for getTables (which can be improved using what I described above)

Something that I mentioned to Michael is that we can get better performance and resolve this issue at the same time by just sending a single query as follows:

```
SELECT table_name, lower(table_name)
FROM information_schema.tables
WHERE table_name = 'inputfromathena' or lower(table_name) = 'inputfromathena'
LIMIT 1
```

Where the first clause will short circuit the query if the table already exists as all lowercase and then return the first case insensitive match for the table name. This eliminates the need for two queries or any searching for the table name within the connector.

dbadaya1:
@henrymai Yes, Micheal discussed that as well with me but I missed mentioning it here. The things is that `doGetTable()` returns the schema of the table as well. Currently, for the happy case (i.e. no upper-case letters in table name in database), we do a JDBC::getColumns call to get the schema as well as implicitly check if the table exists. So, unless we make changes to build the table schema from the output of `information_schema`, we will end up doing two network calls for the happy case - first to resolve the table name, and then to build the schema. 

But in future, if we make changes to build the schema from information_schema, then we can short-circuit using the approach that you mentioned. I can open an issue to track this.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
```diff
@@ -218,13 +219,22 @@ public GetTableResponse doGetTable(final BlockAllocator blockAllocator, final Ge
...
-                    partitionSchema.getFields().stream().map(Field::getName).collect(Collectors.toSet()));
+            TableName caseInsensitiveTableMatch = caseInsensitiveTableSearch(connection, getTableRequest.getTableName().getSchemaName(),
+                    getTableRequest.getTableName().getTableName());
+            Schema caseInsensitiveSchemaMatch = getSchema(connection, caseInsensitiveTableMatch, partitionSchema);
+            return new GetTableResponse(getTableRequest.getCatalogName(), caseInsensitiveTableMatch, caseInsensitiveSchemaMatch,
+                        partitionSchema.getFields().stream().map(Field::getName).collect(Collectors.toSet()));
         }
     }
 
+    protected TableName caseInsensitiveTableSearch(Connection connection, final String databaseName,
```
hackett123, line 230: last nit, let's have a comment explaining that subclasses supporting this search will override it with functionality and that it's a no-op by default. as is, it "looks" like a bug

hackett123 (approved):

PR #1303, opened 2023-05-29
Remove powermockito dependency

namratas-trianz:
*Issue #, if available:*
Remove powermockito usage
*Description of changes:*


athena-gcs/src/test/java/com/amazonaws/athena/connectors/gcs/GenericGcsTest.java
```diff
@@ -0,0 +1,95 @@
...
+    protected void closeMockedObjects() {
+        mockedS3Builder.close();
+        mockedSecretManagerBuilder.close();
+        mockedAthenaClientBuilder.close();
+        mockedGoogleCredentials.close();
+        mockedGcsUtil.close();
+        mockedServiceAccountCredentials.close();
+    }
+
+    protected static void setInternalState(Object target, String field, Object value) {
```
henrymai, line 61: Just wondering, why is this method necessary vs just using `FieldUtils.writeField`?

Additionally add a comment here why as well if it really is necessary, so that future readers will know the reasoning as well.

henrymai (commented):
Just one minor comment. Thanks a lot for doing this.

athena-gcs/src/test/java/com/amazonaws/athena/connectors/gcs/GenericGcsTest.java
trianz-azam: Hi @henrymai, I've removed setInternalState in class GenericGcsTest and used FieldUtils.writeField as suggested

henrymai (approved):

PR #1307, opened 2023-06-02
Update SqlServerMetadataHandler.java - remove filter issqluser

ssenathi:
The code make assumption that all databases are created by an user who is also issqluser , which is not the case and usually all the databases principle id  points to dbo who is not an issqluser

*Issue #, if available:*

*Description of changes:*


AbdulR3hman:
this has already been part of our changes; closing this PR.

PR #1308, opened 2023-06-02
Federation Release Tests Support

hackett123:
- Contains CDK package for infra spinup, ETL, teardown
- Contains automated test workflow to ensure APIs can be hit without issue and pushdowns work
  - Right now, we test just ddb and mysql
- Create GitHub workflow

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/run_release_tests.yml
+++ b/.github/workflows/run_release_tests.yml
@@ -1,0 +1,47 @@
+name: Run release tests
+
+on:
+  workflow_dispatch:
+
+
+
+jobs:
+  run_release_tests:
+    name: run release tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: install dependencies. note aws is pre-installed and we use a specific action for node.
+        run: |
+          sudo apt install -y tar gzip unzip wget less groff vim git python3 python3-pip unixodbc-dev curl jq;
+          pip3 install boto3;
+      - name: install node and npm
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - name: Set up JDK 11
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'corretto'
+          java-version: '11'
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Build jars
+        run: mvn clean package -T 1C -DskipTests
+      - name: env vars
+        run: echo "REPOSITORY_ROOT=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+      - name: Run tests
+        run: |
+            cd $GITHUB_WORKSPACE/validation_testing;
+            python3 main.py;
+        env:
+          # env vars that are secrets should always be set in the env field, so they aren't visible in $GITHUB_ENV file
+          AWS_DEFAULT_REGION: 'us-east-1'
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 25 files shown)


.github/workflows/run_release_tests.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: env vars
+        run: echo "REPOSITORY_ROOT=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+      - name: Run tests
+        run: |
+            cd $GITHUB_WORKSPACE/validation_testing;
+            python3 main.py;
+        env:
+          # env vars that are secrets should always be set in the env field, so they aren't visible in $GITHUB_ENV file
+          AWS_DEFAULT_REGION: 'us-east-1'
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
hackett123, line 41: these will be temporary credentials to a reduced scope IAM role. our internal release process will have their own scripts to automatically set these and then delete them from the repo afterwards.

GH secrets are encrypted and cannot be accessed, even be committers. They are passed into workflows but are not logged.

validation_testing/scripts/glue_job_synchronous_execution.py
```diff
@@ -0,0 +1,35 @@
...
+    job_run_id = start_job_response['JobRunId']
+    for polls in range(max_polls):
+        status = glue.get_job_run(JobName=job_name, RunId=job_run_id)['JobRun']['JobRunState']
+        if status == 'SUCCEEDED':
+            break 
+        elif status in ['FAILED', 'STOPPED', 'TIMEOUT']:
+            print('Job failed')
+            break
+        elif polls >= max_polls:
+            print('Exceeded max polls')
```
AbdulR3hman, line 21: would we have to manually check the job if it timed out? if so maybe we should add a warning in the logs.

validation_testing/scripts/glue_job_synchronous_execution.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def start_job_and_wait_for_completion(job_name):
+    start_job_response = glue.start_job_run(JobName=job_name)
+    job_run_id = start_job_response['JobRunId']
+    for polls in range(max_polls):
+        status = glue.get_job_run(JobName=job_name, RunId=job_run_id)['JobRun']['JobRunState']
+        if status == 'SUCCEEDED':
+            break 
+        elif status in ['FAILED', 'STOPPED', 'TIMEOUT']:
```
AbdulR3hman, line 17: if this fails, should we provide a warning for manual cleanup?

validation_testing/scripts/glue_job_synchronous_execution.py
hackett123: Addressing both of your comments at once - if there is an issue with the glue job, either bc it didn't finish within our 20 min polling limit, or the job itself timed out, or the job itself failed or stopped, we will fail the full release tests. Then in the logs we would see that `echo "FINISHED RUNNING GLUE JOBS FOR ${CONNECTOR_NAME}."` didn't run and the last echo'd line was "Starting glue jobs...", so we would already know the issue was with the glue job.

In terms of next steps, yes we would then have to go check the glue job itself. But I don't think that's a problem, we will need to go investigate in the console regardless of if we knew it was a timeout or failure. And the reason for fast-failing here, instead of allowing the infra destruction to happen, is so we don't have to deal with the glue job being torn down before we can investigate.

So that was my reasoning behind this set up, let me know if you disagree with any part of it

AbdulR3hman (approved):

PR #1335, opened 2023-06-15
added new connector for OpenSearch data source

akuzin1:
*Description of changes:*
This pull request introduces a new OpenSearch connector that implemented the Athena JDBC interface, leveraging the the OpenSearch driver, to enable metadata and data retrieval for the Athena query engine.


athena-opensearch/src/main/java/com/amazonaws/athena/connectors/opensearch/OpensearchMetadataHandler.java
```diff
@@ -0,0 +1,426 @@
...
+                String columnName = resultSet.getString("COLUMN_NAME");
+                if (columnType != null && SupportedTypes.isSupported(columnType)) {
+                    if (columnType instanceof ArrowType.List) {
+                        schemaBuilder.addListField(columnName, getArrayArrowTypeFromTypeName(
+                                resultSet.getString("TYPE_NAME"),
+                                resultSet.getInt("COLUMN_SIZE"),
+                                resultSet.getInt("DECIMAL_DIGITS")));
+                    }
+                    else if (columnType instanceof ArrowType.Struct) {
+                        continue;
```
hackett123, line 309: i remember you said that structs are flattened into single fields containing ".", but then what circumstance will return a column of type Struct where you need this continue case?

athena-opensearch/src/main/java/com/amazonaws/athena/connectors/opensearch/OpensearchMetadataHandler.java
```diff
@@ -0,0 +1,426 @@
...
+                    }
+                    else {
+                        if (columnName.contains(".")) {
```
hackett123, line 312: are column names that are not flattened structs allowed to contain "." ?

athena-opensearch/src/main/java/com/amazonaws/athena/connectors/opensearch/OpensearchMetadataHandler.java
```diff
@@ -0,0 +1,426 @@
...
+
+    private List<TableName> listTables(final Connection jdbcConnection, final String databaseName)
+            throws SQLException
+    {
+        try (ResultSet resultSet = getTables(jdbcConnection, databaseName)) {
+            ImmutableList.Builder<TableName> list = ImmutableList.builder();
+            while (resultSet.next()) {
+                // checks if table name is internal table ex: .kibana
+                String tableName = resultSet.getString("TABLE_NAME");
+                if (tableName.startsWith(".")) {
```
hackett123, line 376: nit, just combine these two conditionals with &&. it would make sense to keep them separate if we had alternate behavior in the inner conditional, but we do not.

athena-opensearch/src/main/java/com/amazonaws/athena/connectors/opensearch/OpensearchMetadataHandler.java
```diff
@@ -0,0 +1,426 @@
...
+                    if (!tableName.matches(DATA_STREAM_REGEX)) {
+                        continue;
+                    }
+                }
+                // update schema from null because Opensearch doesn't have schema but change later
+                // resultSet.updateString("TABLE_SCHEM", databaseName);
+                
+                // Temporary in schemaName 
+                list.add(getSchemaTableName(resultSet, databaseName));
```
hackett123, line 385: having a hard time understanding the comments here

athena-opensearch/src/main/java/com/amazonaws/athena/connectors/opensearch/OpensearchMetadataHandler.java
```diff
@@ -0,0 +1,426 @@
...
+    private TableName getSchemaTableName(final ResultSet resultSet, String schemaName)
+            throws SQLException
+    {
+        String tableName = resultSet.getString("TABLE_NAME");
+        String tableSchema = resultSet.getString("TABLE_SCHEM");
+        Pattern pattern = Pattern.compile(DATA_STREAM_REGEX);
+        Matcher matcher = pattern.matcher(tableName);
+        if (matcher.find()) {
+            tableName = matcher.group(2);
+        }
```
hackett123, line 400: can you explain this? I thought you mentioned data streams are not yet supported. Also, let's avoid magic numbers or add an example in the comment for what this looks like to explain why we match on group 2

athena-opensearch/src/main/java/com/amazonaws/athena/connectors/opensearch/OpensearchRecordHandler.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Data handler, user must have necessary permissions to read from necessary tables.
+ */
+public class OpensearchRecordHandler
+        extends JdbcRecordHandler
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(OpensearchRecordHandler.class);
+
+    private static final String OPENSEARCH_QUOTE_CHARACTER = "`";
```
hackett123, line 62: you already have this defined in OpensearchConstants

athena-opensearch/src/main/java/com/amazonaws/athena/connectors/opensearch/OpensearchRecordHandler.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public PreparedStatement buildSplitSql(Connection jdbcConnection, String catalogName, TableName tableName, Schema schema, Constraints constraints, Split split)
+            throws SQLException
+    {
+        PreparedStatement preparedStatement = jdbcSplitQueryBuilder.buildSql(jdbcConnection, null, null, tableName.getTableName(), schema, constraints, split);
+
+        // Disable fetching all rows.
+        preparedStatement.setFetchSize(10000);
```
hackett123, line 102: let's put this in a constant.

what happens if the input query has over 10k results?

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
```diff
@@ -135,14 +142,21 @@ protected JdbcCredentialProvider getCredentialProvider()
...
     @Override
     public void readWithConstraint(BlockSpiller blockSpiller, ReadRecordsRequest readRecordsRequest, QueryStatusChecker queryStatusChecker)
             throws Exception
     {
-        LOGGER.info("{}: Catalog: {}, table {}, splits {}", readRecordsRequest.getQueryId(), readRecordsRequest.getCatalogName(), readRecordsRequest.getTableName(),
+        LOGGER.info("Read Record Request {}: Catalog: {}, table {}, splits {}", readRecordsRequest.getQueryId(), readRecordsRequest.getCatalogName(), readRecordsRequest.getTableName(),
                 readRecordsRequest.getSplit().getProperties());
         try (Connection connection = this.jdbcConnectionFactory.getConnection(getCredentialProvider())) {
-            connection.setAutoCommit(false); // For consistency. This is needed to be false to enable streaming for some database types.
+            if (this.getAutoCommit().isPresent()) {
```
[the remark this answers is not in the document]
hackett123: oh i see, in the opensearch record handler you have it return Optional.empty(). But is there a reason to even have that as an optional wrapping around a boolean, instead of it just being a boolean?

athena-opensearch/src/main/java/com/amazonaws/athena/connectors/opensearch/OpensearchUtils.java
```diff
@@ -0,0 +1,62 @@
...
+import com.amazonaws.athena.connector.lambda.data.FieldBuilder;
+import org.apache.arrow.vector.types.Types;
+import org.apache.arrow.vector.types.pojo.ArrowType;
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.arrow.vector.types.pojo.FieldType;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class OpensearchUtils
```
hackett123, line 31: if you don't expect this class's functionality to increase, you can just rename it more specific to OpensearchNestedStructBuilder so that it's better tied to its purpose.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/FieldBuilder.java
```diff
@@ -42,6 +42,7 @@ public class FieldBuilder
     private final boolean nullable;
     //Using LinkedHashMap because Apache Arrow makes field order important so honoring that contract here
     private final Map<String, Field> children = new LinkedHashMap<>();
+    private final Map<String, FieldBuilder> nestedChildren = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
hackett123: after reading through the rest of the PR, let me know if my understanding is correct.

if i have a field like:

```
"topObj": {
  "innerObj": {
    "intField": 1,
    "strField": "asdf"
  },
  "int": 5
}
```

then opensearch will send back each of these fields as their own columns, and our result set will have this:
`topObj.int`
`topObj.innerObj.intField`
`topObj.innerObj.strField`

so the reason for this new nested children field and it's corresponding logic with `OpensearchUtils` class is so you can say `createNestedStruct` andd pass in `topObj.innerObj.intField` (for example), have it find the struct `topObj`, and continue to build to the struct definition for topObj?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/BlockUtils.java
```diff
@@ -849,6 +849,12 @@ else if (value instanceof Integer) {
                     if (value == null) {
                         float4Writer.writeNull();
                     }
+                    else if (value instanceof java.lang.Integer) {
```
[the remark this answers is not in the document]
akuzin1: They should not. I used similar structure to other data types in `BlockUtils::writeSimpleValue`, checking if value is of specific instance helps with writing value to block, doesn't change core behavior.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/FieldBuilder.java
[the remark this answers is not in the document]
akuzin1: Yes that's correct. It makes it more intuitive to make complex nested fields in my opinion.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/SchemaBuilder.java
```diff
@@ -89,6 +89,18 @@ public SchemaBuilder addStructField(String fieldName)
         return this;
     }
 
+    /**
+     * Adds a new Nested STRUCT Field to the Schema as a top-level Field.
```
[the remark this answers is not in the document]
akuzin1: No, I don't believe that existing connectors should be impacted, its independent. We do support nested structs in other connectors. The added functions were kind of tailored to OpenSearch use case and I thought it made sense to expand the SchemaBuilder functionality.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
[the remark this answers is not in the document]
akuzin1: Yes, so the reason it's an Optional is because, the presence of a value does not signify `true`, we want to maintain the original behavior.
I can go back and do additional testing, however, we did not want to setAutoCommit for the OpenSearch Connector.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
```diff
@@ -168,8 +185,9 @@ public void readWithConstraint(BlockSpiller blockSpiller, ReadRecordsRequest rea
                     rowsReturnedFromDatabase++;
                 }
                 LOGGER.info("{} rows returned by database.", rowsReturnedFromDatabase);
-
-                connection.commit();
+                if (this.getAutoCommit().isPresent()) {
```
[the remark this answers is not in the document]
akuzin1: We only need to commit connection if we setAutoCommit, and we don't need to do that for the OpenSearch driver.

athena-opensearch/src/main/java/com/amazonaws/athena/connectors/opensearch/OpensearchMetadataHandler.java
akuzin1: Left over comments. Will get rid of.

athena-opensearch/src/main/java/com/amazonaws/athena/connectors/opensearch/OpensearchRecordHandler.java
akuzin1: Setting fetch size `Gives the JDBC driver a hint as to the number of rows that should be fetched from the database when more rows are needed for ResultSet objects generated by this Statement`.
So what this really controls is the number of network calls, as well as enforces pagination so that SQL Plugin uses v2 instead of legacy engine. Otherwise legacy version only returns 200 rows per table.

macohen:
What’s the sql/jdbc feature that’s needed from opensearch to bring this closer to merging? Also, are there any performance comparisons done between sql/jdbc vs using a Java or another client?

akuzin1:
Hi, @macohen! We were waiting for the implementation of shard partitioning, which would allow to parallelize the fetching of data with the JDBC driver. Currently the connector implements a single partition solution, which resulted in performance regression compared to current connector version. 
I had opened an feature request and was working with the OpenSearch team to track item, however, due to shifting priorities the item got backlogged.

akuzin1:
@macohen linking issue [here](https://github.com/opensearch-project/sql/issues/1478) from` opensearch-project/sql `repository

macohen:
We are going to need to take another look at the OpenSearch connector altogether. Closing this stale PR.

PR #1365, opened 2023-07-07
Pushing down 'not equal' to predicate

akuzin1:
**Description of changes:**
Modified statement that is responsible for adding `IS NOT NULL` clause to generate sql for JDBC connectors.
Prior to change, when evaluating a `not equal to` predicate it would short circuit and not push down all predicates to sql statement.
For Example:
`SELECT * FROM item WHERE i_manufact_id <> 138`
Would be translated to:
`SELECT * FROM item WHERE (i_manufact_id IS NOT NULL)`

This is the result of calling `getSpan()` on `SortedRangeSet`, which returns a Range that is a summary of the lowest lower bound and the highest upper bound that represents this ValueSet.

Therefore, the following ValueSet from statement with `i_manufact_id <> 138` predicate:

```
        i_manufact_id=SortedRangeSet{
            type=Int(64, true), nullAllowed=false, lowIndexedRanges={
                Marker{
                   valueBlock=Int(64, true), nullValue=true, valueBlock=true, bound=ABOVE
                   }=Range{
                        low=Marker{
                            valueBlock=Int(64, true), nullValue=true, valueBlock=true, bound=ABOVE
                           }, 
                        high=Marker{
                            valueBlock=Int(64, true), nullValue=false, valueBlock=138, bound=BELOW
                        }
                   }, 
                Marker{
                   valueBlock=Int(64, true), nullValue=false, valueBlock=138, bound=ABOVE
                   }=Range{
                        low=Marker{
                            valueBlock=Int(64, true), nullValue=false, valueBlock=138, bound=ABOVE
                        }, 
                        high=Marker{
                            valueBlock=Int(64, true), nullValue=true, valueBlock=true, bound=BELOW
                        }
                   }
               }
           }
       }
```

and  the following ValueSet from statement with `i_manufact_id IS NOT NULL` predicate:

```
i_manufact_id=SortedRangeSet{
            type=Int(64, true), nullAllowed=false, lowIndexedRanges={
                Marker{
                    valueBlock=Int(64, true), nullValue=true, valueBlock=true, bound=ABOVE
                }
                =Range{
                    low=Marker{
                        valueBlock=Int(64, true), nullValue=true, valueBlock=true, bound=ABOVE
                    }, 
                    high=Marker{
                        valueBlock=Int(64, true), nullValue=true, valueBlock=true, bound=BELOW
                    }
                }
           }
        }
```

evaluate to the same `Range` when called `getSpan` on and they exit at the same condition.

According to ANSI SQL standards, when assessing a predicate comparison involving a `NULL` value, it does not yield true or false outcomes and is consequently excluded from the result set. This means that if there is more than one comparison operator performed on a single column, it will automatically drop `NULL` values, so `IS NOT NULL` only needs to be pushed down if it that's the only one.

So for example, in the following statement:
`SELECT * FROM item WHERE i_manufact_id <> 138 AND i_manufact_id IS NOT NULL`
the latter part will be inherently be achieved by pushing down first part.

The change results in 'IS NOT NULL' predicate only being pushed down if there's one Range present in ValueSet.

Generated SQL After Change: 
`SELECT * FROM item WHERE ((i_manufact_id < 138) OR (i_manufact_id > 138))`

***Testing:***
Because `JdbcSplitQueryBuilder` is an abstract class and can't be initialized, I added unit tests to the `MySqlQueryStringBuilder` to verify proper behavior for generated SQL string, when predicates `!=` and `IS NOT NULL` are being used.

### DIFF
--- a/athena-mysql/src/test/java/com/amazonaws/athena/connectors/mysql/MySqlQueryStringBuilderTest.java
+++ b/athena-mysql/src/test/java/com/amazonaws/athena/connectors/mysql/MySqlQueryStringBuilderTest.java
@@ -1,0 +1,108 @@
+/*-
+ * #%L
+ * athena-mysql
+ * %%
+ * Copyright (C) 2019 Amazon Web Services
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package com.amazonaws.athena.connectors.mysql;
+
+import com.amazonaws.athena.connector.lambda.QueryStatusChecker;
+import com.amazonaws.athena.connector.lambda.data.BlockAllocator;
+import com.amazonaws.athena.connector.lambda.data.BlockAllocatorImpl;
+import com.amazonaws.athena.connector.lambda.data.BlockWriter;
+import com.amazonaws.athena.connector.lambda.data.FieldBuilder;
+import com.amazonaws.athena.connector.lambda.data.SchemaBuilder;
+import com.amazonaws.athena.connector.lambda.domain.Split;
+import com.amazonaws.athena.connector.lambda.domain.TableName;
+import com.amazonaws.athena.connector.lambda.domain.predicate.Constraints;
+import com.amazonaws.athena.connector.lambda.domain.predicate.Range;
+import com.amazonaws.athena.connector.lambda.domain.predicate.SortedRangeSet;
+import com.amazonaws.athena.connector.lambda.metadata.GetSplitsRequest;
+import com.amazonaws.athena.connector.lambda.metadata.GetSplitsResponse;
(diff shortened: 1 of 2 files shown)


chngpe (approved):

burhan94:
can we add some test cases for this pls?

athena-mysql/src/test/java/com/amazonaws/athena/connectors/mysql/MySqlQueryStringBuilderTest.java
```diff
@@ -0,0 +1,110 @@
...
+        .addField(FieldBuilder.newBuilder("testCol3", org.apache.arrow.vector.types.Types.MinorType.VARCHAR.getType()).build())
+        .addField(FieldBuilder.newBuilder("testCol4", org.apache.arrow.vector.types.Types.MinorType.FLOAT8.getType()).build())
+        .addField(FieldBuilder.newBuilder("partition_name", org.apache.arrow.vector.types.Types.MinorType.VARCHAR.getType()).build())
+        .build();
+
+    @Test
+    public void generateSqlIsNotNull() throws Exception
+    {
+        Map<String, ValueSet> constraintsMap = new HashMap<>();
+        constraintsMap.put("testCol2", SortedRangeSet.of(false, Range.all(allocator, org.apache.arrow.vector.types.Types.MinorType.INT.getType())));
```
hackett123, line 81: nit, you can init and add the map values in one line using ImmutableMap.of(key, value, key2, value2, ...).

It's generally cleaner to avoid initing a collection and later adding or modifying it (called the init-then-modify antipattern). This here isn't maybe only a violation of it from a purists perspective, but you can imagine how it can lead to cleaner code. so this comment applies to both tests in this file

PR #1373, opened 2023-07-12
limit clause pushdown fix for certain data sources

namratas-trianz:
*Issue #, if available:*

*Description of changes:*
Certain data sources does not support limit clause, when for these data sources , a limit query is executed in Athena, it throws exception, as currently limit clause is pushed down.  The fix is to solve the error raised due to limit clause pushdown.


chngpe (approved):

hackett123:
Can you please add a test to validate that if Constraints come in with a LIMIT clause that we don't generate the LIMIT clause in the sql builder?

namratas-trianz:
> Can you please add a test to validate that if Constraints come in with a LIMIT clause that we don't generate the LIMIT clause in the sql builder?

Hi @hackett123 , added test cases.

Issue #1073, opened 2023-02-25
[FEATURE] Allow multiplexing handler connection strings to be configured via SAM templates


ned9:
Connectors which provide a multiplexing handlers to route queries from data source catalogues to specific target databases rely on configured lambda environment variables to route incoming queries to the appropriate target database. The environment variable names take the form `$<catalog>_connection_string`, where `catalog` corresponds to the name of the Athena data source from which queries will be routed to the location specified by the connection string in the value of the variable.

Currently, there's no way to configure these environment variables from the SAM templates of the connectors, which means that, unless I'm missing something, none of the multiplexing connectors can be configured for automated deployment through eg the CDK.

The simplest workaround to this is to simply not use multiplexing handlers, and have separate connectors for each target database, however this has the downside of potentially greatly increasing the surface area of the deployed stack when there are many different databases, is is the case for our use-case.

See these related issues: #143, #517

This limitation affects the following connectors, which all provide multiplexing handlers:
* athena-cloudera-hive
* athena-cloudera-impala
* athena-datalakegen2
* athena-db2
* athena-hortonworks-hive
* athena-mysql
* athena-oracle
* athena-postgresql
* athena-redshift
* athena-saphana
* athena-snowflake
* athena-sqlserver
* athena-synapse
* athena-teradata

henrymai:
Hi @ned9 

It looks like you can use CDK to update environment variables after deployment: https://stackoverflow.com/questions/70825543/aws-cdk-update-existing-lambda-environment-variables/70832790#70832790

Have you tried that yet?

henrymai:
Closing out this response since there's been no response for 3 weeks. Please re-open if the link above doesn't help.

ned9:
Apologies for the slow reply!

Thanks for digging up that candidate strategy. It's possible that could do what we need, but it wasn't clear to us that it would work in the context of a single stack hydration, where the existing resource being targeted was only just created in the same deployment. Given there's some degree of uncertainty, we decided not to risk any more delays and went with the safe option of just spinning up a separate connector lambda for each database that is added.

PR #1394, opened 2023-07-24
Add missing env variables to template.

dtaskai:
*Description of changes:* Add missing environment variables to Athena DynamoDB connector template, according to documentation: https://docs.aws.amazon.com/athena/latest/ug/connectors-dynamodb.html


dtaskai:
We dearly miss these few environment variables from the template, as without them being present in the template we have no way of setting them programmatically in CDK, only manually on the Lambda itself. 😞

athena-dynamodb/athena-dynamodb.yaml
```diff
@@ -24,6 +24,20 @@ Parameters:
     Description: 'The prefix within SpillBucket where this function can spill data.'
     Type: String
     Default: athena-spill
+  SpillPutRequestHeaders:
+    Description: '(Optional) A JSON encoded map of request headers and values for the Amazon S3 putObject request that is used for spilling. (for example, {"x-amz-server-side-encryption" : "aws:kms"}).'
+    Type: String
+  DisableGlue:
+    Description: "(Optional) If present and set to 'true', the connector does not attempt to retrieve supplemental metadata from AWS Glue."
+    Default: 'false'
+    Type: String
```
chngpe, line 33: Add allowedValues to provided validation on input 
    AllowedValues:
      - true
      - false

athena-dynamodb/athena-dynamodb.yaml
```diff
@@ -24,6 +24,20 @@ Parameters:
...
+  GlueCatalog:
+    Description: '(Optional) Use this option to specify the account id of a different account for a cross-account AWS Glue catalog. By default, the connector attempts to get metadata from its own AWS Glue account.'
+    Type: String
+  DisableProjectionAndCasing:
+    Description: "(Optional) Disables projection and casing. Use if you want to query DynamoDB tables that have casing in their column names and you do not want to specify a columnMapping property on your AWS Glue table. Can be set to 'auto' or 'always', set to 'auto' by default. See documentation for more information."
+    Default: 'auto'
+    Type: String
```
chngpe, line 40: Add allowedValues to provided validation on input
    AllowedValues:
      - auto
      - always

athena-dynamodb/athena-dynamodb.yaml
```diff
@@ -51,6 +65,8 @@ Parameters:
 
 Conditions:
   HasKMSKeyId: !Not [!Equals [!Ref KMSKeyId, ""]]
+  HasSpillPutRequestHeaders: !Not [!Equals [!Ref HasSpillPutRequestHeaders, ""]]
```
chngpe, line 68: This `HasSpillPutRequestHeaders` should be `SpillPutRequestHeaders`

athena-dynamodb/athena-dynamodb.yaml
```diff
@@ -24,6 +24,20 @@ Parameters:
...
+  SpillPutRequestHeaders:
+    Description: '(Optional) A JSON encoded map of request headers and values for the Amazon S3 putObject request that is used for spilling. (for example, {"x-amz-server-side-encryption" : "aws:kms"}).'
+    Type: String
+  DisableGlue:
+    Description: "(Optional) If present and set to 'true', the connector does not attempt to retrieve supplemental metadata from AWS Glue."
+    Default: 'false'
+    Type: String
+  GlueCatalog:
+    Description: '(Optional) Use this option to specify the account id of a different account for a cross-account AWS Glue catalog. By default, the connector attempts to get metadata from its own AWS Glue account.'
+    Type: String
```
chngpe, line 36: allowed Values as `'^\d{12}$'`

chngpe:
Hi!

Thank you for submitting the PR for adding the missing environment variables into the template. 
I left some comments, mostly on restrict the user input to only the allowed values. 

-P

henrymai:
This PR shouldn't actually be necessary if all you need to do is set environment variables after deployment with CDK.

See my other comment here about the same topic: https://github.com/awslabs/aws-athena-query-federation/issues/1073#issuecomment-1445026825

dtaskai:
Thanks for the input and the proposed solution @henrymai! though I feel like the expected behavior would be that all of the documented environment variables are available in the template, not just select ones (also makes for a much leaner solution in CDK), that is why we were surprised that our deployment was failing in the first place.

atipapp:
@henrymai seems like Dominik addressed your comment and I agree with him. I think specifying it through a template parameter would be leaner.

 @chngpe can you please re-review?

henrymai:
@atipapp I don't work on this team anymore. I just added my comment as a way for you guys to resolve your problems faster without waiting for another release.

chngpe:
Hi! 
Sorry for late response.

After discussion this internally with team. The parameters was intentionally left out in the template. We want the templates to be extremely light weight and easier for customer to get started. Unfortunately, due to the same reason team decide not to add more parameters to the template. If needed, please follow the suggestion mentioned above

```
https://github.com/awslabs/aws-athena-query-federation/pull/1394#issuecomment-1653241263
https://github.com/awslabs/aws-athena-query-federation/issues/1073#issuecomment-1445026825
```

PR #1453, opened 2023-08-31
Fix dynamic filter date on timezone issue.

chngpe:
*Issue #, if available:*

*Description of changes:*


athena-mysql/src/test/java/com/amazonaws/athena/connectors/mysql/MySqlQueryStringBuilderTest.java
```diff
@@ -105,4 +113,28 @@ public void generateSqlIsNotEqual() throws Exception
         assertEquals(expectedPreparedStatement, preparedStatement);
 
     }
+
+    @Test
+    public void testDatePredicate() throws Exception
```
burhan94, line 118: We need a test that validates that the long value of the date gets converted to the human readable date correctly. EG: 10440 converts to aug 2nd 1998

burhan94 (changes requested):

henrymai (approved):

Issue #506, opened 2021-10-21
[BUG] not authorized on <database> to execute command { find: "system.views" } with connector athena-docdb


theFlyRoper:
**Describe the bug**
DocumentDB fails when connecting to MongoDB if there are any collections in the current database that the connector account lacks access to. 

**To Reproduce**
1. Deploy athena-docdb connector with connection string for mongodb Atlas (https://www.mongodb.com/atlas/database)
2. Load the data source in the AWS console Athena query editor 
3. If there is a collection which the connecting account lacks permission to read, the docdb connector fails entirely.

**Expected behavior**
If the account lacks permission to read the collection, the connector should move to the next collection and not show the forbidden collection in the list of athena database tables.  Or it should show some indicator that the connector lacks read access to that collection.

**Screenshots / Exceptions / Errors**
![Athena_DocDB_BugReport_2021-10-21](https://user-images.githubusercontent.com/16727819/140528864-3cdf0951-e044-4474-8737-07b405fa52f5.png)
Here is more of the error text:
Failed to get metadata for table (system.views) from lambda function due to java.lang.RuntimeException: com.amazonaws.services.lambda.invoke.LambdaFunctionException: Query failed with error code 13 and error message 'not authorized on <database> to execute command { find: "system.views", limit: 10, batchSize: 10,, $clusterTime: { clusterTime: Timestamp(1634825858, 1), signature: { hash: BinData(0, <hash>), keyId: <key> } }, lsid: { id: UUID("00002000-efef-4444-b425-a7ae91a587b5") } }' on server <mongodb server>:27017

Trace:
Query failed with error code 13 and error message 'not 
authorized on <database> to execute command { find: "system.views", limit: 10, batchSize: 10, $db: "<database>", $clusterTime: { clusterTime: Timestamp(1634825858, 1), signature: { hash: BinData(0, 4F26FA56443940708650AF6D9B303D29A05723A7), keyId: 6964759163792523267 } }, lsid: { id: UUID("be502000-e864-4f18-b425-a7ae91a587b5") } }' on server <database>:27017: com.mongodb.MongoQueryException
[...]

PR #1491, opened 2023-09-20
DocDB skip collections that user has no permission

chngpe:
DocDB skip collections that user has no permission to access by using mongoDB command line to retrieve collection names

*Issue #, if available:*
https://github.com/awslabs/aws-athena-query-federation/issues/506
*Description of changes:*

### DIFF
--- a/athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandler.java
+++ b/athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandler.java
@@ -42,11 +42,15 @@ import com.amazonaws.services.athena.AmazonAthena;
 import com.amazonaws.services.glue.AWSGlue;
 import com.amazonaws.services.glue.model.Table;
 import com.amazonaws.services.secretsmanager.AWSSecretsManager;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
+import com.mongodb.MongoCommandException;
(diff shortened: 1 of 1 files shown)


athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandler.java
```diff
@@ -160,15 +168,58 @@ public ListSchemasResponse doListSchemaNames(BlockAllocator blockAllocator, List
...
             while (itr.hasNext()) {
                 tables.add(new TableName(request.getSchemaName(), itr.next()));
             }
 
             return new ListTablesResponse(request.getCatalogName(), tables, null);
         }
+        catch (MongoCommandException mongoCommandException) {
+            //do this in failed case instead of replace method in case API changes on doc db.
+            logger.warn("Exception on listCollectionNames on Mongo JAVA client, trying with mongo command line.", mongoCommandException);
+            List<TableName> tableNames = doListTablesWithCommand(client, request);
```
AbdulR3hman, line 183: is there a way to detect the underlying data-source, and perform this proactively? instead of falling back on an exception?

chngpe: db.hostInfo return this information but this required an Admin permission on DB, which not all data user will have. Therefore, doing this mechanism to not break or introduce any additional permission

AbdulR3hman (approved):

PR #1692, opened 2024-01-09
Remove direct dependency of Netty

AbdulR3hman:
### *Description of changes:*
Netty has been directly chosen as a direct dependency for Federation Repo whereas its never used directly by any code-base (except one util class that got replaces accordingly).

Most importantly; the depend-bot has been upgrading Netty; which is core to `arrow` functionality, and recently we have upgraded to latest version which caused our SDK to fail.

This change removes the burden of deciding which Netty to choose; and let the transitive dependency be resolve by the direct dependency (in this case Arrow). 

I've ran unit-tests locally; and everything looks fine; however I recommend that we run integration tests; I'll run them on my fork afterwards. 

### DIFF
--- a/athena-federation-sdk/pom.xml
+++ b/athena-federation-sdk/pom.xml
@@ -192,26 +192,6 @@
             <groupId>org.apache.arrow</groupId>
             <artifactId>arrow-memory-netty</artifactId>
             <version>${apache.arrow.version}</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>io.netty</groupId>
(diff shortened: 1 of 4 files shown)


AbdulR3hman:
running an integration test to make sure all is good: https://github.com/AbdulR3hman/aws-athena-query-federation/actions/runs/7476910397/job/20348393176 

(Redshift has been excluded due to some bug we are waiting on a fix or a workaround).

AbdulR3hman:
integration have passed.

akuzin1 (approved):

PR #1717, opened 2024-01-16
Adding extra cleanup procedures before executing release tests 

AbdulR3hman:
These changes will ensure that we clean up anything that left as a residue from previous runs; we have been experiencing some resources that are not detaching or deleting in timely manner that would results in the entire stack failing.

Given we switched to daily run for our tests; this will give plenty of time by the next run for the resources to detach correctly; then clean up those resources then the stack. Before starting a new release tests. 

I've tested locally;

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ node_modules
 package-lock.json
 cdk.out
 .env
+**/*.dylib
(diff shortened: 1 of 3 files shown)


AbdulR3hman:
I've tested the changes using github action; https://github.com/awslabs/aws-athena-query-federation/actions/runs/7547230152

akuzin1 (approved):

PR #1769, opened 2024-02-21
SDK-Side: Query Passthrough Implementation

AbdulR3hman:
### Query Passthrough (QPT)
These changes will allow the connector to extend query passthrough functionality, and provide an interface to define QPT signature per connector.

### DIFF
--- a/athena-federation-integ-test/pom.xml
+++ b/athena-federation-integ-test/pom.xml
@@ -198,7 +198,8 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
-            <version>3.14.0</version>
+             <!--- to meet engine version 3.12.0-->
+            <version>3.12.0</version>
         </dependency>
     </dependencies>
     <build>
--- a/athena-federation-sdk/pom.xml
+++ b/athena-federation-sdk/pom.xml
@@ -287,7 +287,8 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
-            <version>3.14.0</version>
+             <!--- to meet engine version 3.12.0 -->
+            <version>3.12.0</version>
         </dependency>
         <dependency>
             <groupId>net.jqwik</groupId>
@@ -457,7 +458,25 @@
                     </execution>
                 </executions>
             </plugin>
-
+            <plugin>
+                 <!--- Allow to Add *-source.jar;  -->
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-source-plugin</artifactId>
+                <version>${mvn.source.plugin.version}</version>
+                <executions>
+                    <execution>
+                        <id>attach-sources</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>jar-no-fork</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <attach>true</attach>
+                    <skipSource>false</skipSource>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 </project>
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
@@ -25,7 +25,6 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Objects;
 
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -49,14 +48,9 @@ public class Constraints
     private List<FederationExpression> expression;
     private final List<OrderByField> orderByClause;
     private long limit;
+    private Map<String, String> queryPassthroughArguments;
 
     @Deprecated
-    public Constraints(Map<String, ValueSet> summary) 
-    {
-        this(summary, Collections.emptyList(), Collections.emptyList(), DEFAULT_NO_LIMIT);
-    }
-
-    @JsonCreator
     public Constraints(@JsonProperty("summary") Map<String, ValueSet> summary,
                        @JsonProperty("expression") List<FederationExpression> expression,
                        @JsonProperty("orderByClause") List<OrderByField> orderByClause,
@@ -66,6 +60,28 @@ public class Constraints
         this.expression = expression;
         this.orderByClause = orderByClause;
         this.limit = limit;
+    }
+
+    /**
+     *
+     * @param summary
+     * @param expression
+     * @param orderByClause
+     * @param limit
+     * @param queryPassthroughArguments
+     */
+    @JsonCreator
+    public Constraints(@JsonProperty("summary") Map<String, ValueSet> summary,
+                       @JsonProperty("expression") List<FederationExpression> expression,
+                       @JsonProperty("orderByClause") List<OrderByField> orderByClause,
+                       @JsonProperty("limit") long limit,
+                       @JsonProperty("queryPassthroughArguments") Map<String, String> queryPassthroughArguments)
+    {
+        this.summary = summary;
+        this.expression = expression;
+        this.orderByClause = orderByClause;
+        this.limit = limit;
+        this.queryPassthroughArguments = queryPassthroughArguments;
     }
 
     /**
@@ -103,6 +119,16 @@ public class Constraints
         return this.orderByClause != null && this.orderByClause.size() > 0;
     }
 
+    public Map<String, String> getQueryPassthroughArguments()
+    {
+        return this.queryPassthroughArguments;
+    }
+
+    public boolean isQueryPassThrough()
+    {
+        return this.queryPassthroughArguments != null && !this.queryPassthroughArguments.isEmpty();
+    }
+
     @Override
     public boolean equals(Object o)
     {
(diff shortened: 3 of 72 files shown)


athena-federation-integ-test/pom.xml
```diff
@@ -198,7 +198,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
-            <version>3.14.0</version>
+            <version>3.12.0</version>
```
chngpe, line 201: why are we downgrading?

AbdulR3hman: It conflicts with internal Engine version; it has to be the same version or lower.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/GetTableRequest.java
```diff
@@ -55,6 +58,29 @@ public GetTableRequest(@JsonProperty("identity") FederatedIdentity identity,
...
+    @JsonCreator
+    public GetTableRequest(@JsonProperty("identity") FederatedIdentity identity,
+            @JsonProperty("queryId") String queryId,
+            @JsonProperty("catalogName") String catalogName,
+            @JsonProperty("tableName") TableName tableName,
+            @JsonProperty("queryPassthroughArguments") Map<String, String> queryPassthroughArguments)
+    {
+        super(identity, MetadataRequestType.GET_TABLE, queryId, catalogName);
+        requireNonNull(tableName, "tableName is null");
+        requireNonNull(queryPassthroughArguments, "queryPassthroughArguments is null");
```
chngpe, line 78: why are we deprecated above?

Are we promoting using the new GetTableRequest? If yes, then why is queryPassthroughArguments a required field? I assume customer should able to getTableRequest

athena-federation-sdk/pom.xml
```diff
@@ -287,7 +287,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
-            <version>3.14.0</version>
+            <version>3.12.0</version>
```
AbdulR3hman, line 290: Same here; we need to make sure that we aren't using another higher than the dependency used by the engine.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/optimizations/querypassthrough/QueryPassthroughSignature.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Adds the query passthrough implementation, if user has not disabled it, to the connector's capabilities
+     * @param capabilities
+     * @param configOptions
+     */
+    default void addQueryPassthroughCapabilityIfEnabled(ImmutableMap.Builder<String, List<OptimizationSubType>> capabilities, Map<String, String> configOptions)
+    {
+        if (allowQueryPassthrough(configOptions)) {
```
chngpe, line 141: What is the purpose of disabling this?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/GetTableRequest.java
AbdulR3hman: There are few things that are happening here;

* The previously `@Deprecated` constructor that was cleaned up; this was used in very old version of the SDK; and should have been cleaned up before; but since I had to deprecate the current constructor as well (see reasons below); I ended up doing the clean up as well. the previously deprecated constructor was only used by unit tests; thus made the clean up easier — Same changes took place at the backend (engine).

* Secondly; the reason why its now tagged deprecated; SerDe can only support one version @JsonCreator; (aka one constructor); and our SerDe doesn't have the flexibility of optional field (short coming of the SerDe); therefore an empty map is minimal; 

* Regarding why Passthrough map is required; it is not; users can pass empty map; as shown in the test cases. It needs to be part of the Signature due to the fact above; SerDe we are using can only support one constructor.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/optimizations/querypassthrough/QueryPassthroughSignature.java
AbdulR3hman: To allow customer to opt out from enable QPT; customer can add `enable_query_passthrough` env variable to their lambda environment and disable QPT.

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneMetadataHandlerTest.java
```diff
@@ -53,6 +53,7 @@
 
 import static org.junit.Assert.*;
 
+import static com.amazonaws.athena.connector.lambda.domain.predicate.Constraints.DEFAULT_NO_LIMIT;
```
chngpe, line 56: nit since this is test, this import not used

AbdulR3hman: I'll clean it up.

AbdulR3hman: done.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
```diff
@@ -49,23 +48,40 @@ public class Constraints
     private List<FederationExpression> expression;
     private final List<OrderByField> orderByClause;
     private long limit;
+    private Map<String, String> queryPassthroughArguments;
 
     @Deprecated
-    public Constraints(Map<String, ValueSet> summary) 
+    public Constraints(@JsonProperty("summary") Map<String, ValueSet> summary,
```
chngpe, line 54: why are we adding more stuff to a constructor that is deprecated?

chngpe: nvm... diff issue if you compare raw file

chngpe (approved):

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/GetTableRequest.java
```diff
@@ -44,7 +47,7 @@ public class GetTableRequest
      * @param catalogName The catalog name that the table belongs to.
      * @param tableName The name of the table metadata is being requested for.
      */
-    @JsonCreator
+    @Deprecated
```
burhan94, line 50: so i don't think this can be deprecated directly since this is a top level Request object. If I remember correctly, this needs to always have 1 single constructor and we need to modify older request/response serDe's to take this into account.

burhan94: a good example is to take a look at the way we modified the ListTableRequest API where we added pagination token + page size.

burhan94: also, there should really be no reason to ever call the deprecated API anymore. I think we should delete.

pom.xml
```diff
@@ -13,18 +13,21 @@
...
+        <!--- to meet engine version 2.0.7-->
+        <slf4j-log4j.version>2.0.7</slf4j-log4j.version>
         <mockito.version>4.11.0</mockito.version>
         <junit.version>4.13.2</junit.version>
         <jqwik.version>1.8.2</jqwik.version>
         <assertj.version>3.25.3</assertj.version>
         <testng.version>7.9.0</testng.version>
-        <fasterxml.jackson.version>2.16.1</fasterxml.jackson.version>
+         <!--- to meet engine version 2.12.6 -->
+        <fasterxml.jackson.version>2.12.6</fasterxml.jackson.version>
```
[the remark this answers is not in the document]
burhan94: I'm ok with moving forward as-is, but in the future, we should be more careful to not do things like this.

burhan94 (changes requested):

pom.xml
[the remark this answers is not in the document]
AbdulR3hman: I agree; we can talk offline; but the back-end are know enforcing Maven's `upper bound` restriction; and we have to maintain a version that is accepted;

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/GetTableRequest.java
```diff
@@ -35,26 +37,54 @@ public class GetTableRequest
...
+     * @param queryPassthroughArguments
      */
     @JsonCreator
     public GetTableRequest(@JsonProperty("identity") FederatedIdentity identity,
-            @JsonProperty("queryId") String queryId,
-            @JsonProperty("catalogName") String catalogName,
-            @JsonProperty("tableName") TableName tableName)
+                           @JsonProperty("queryId") String queryId,
+                           @JsonProperty("catalogName") String catalogName,
+                           @JsonProperty("tableName") TableName tableName, Map<String, String> queryPassthroughArguments)
```
burhan94, line 55: nit: fix spacing pls

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/GetTableRequest.java
```diff
@@ -35,26 +37,54 @@ public class GetTableRequest
...
+//    /**
+//     * Constructs a new GetTableRequest object.
+//     * @param identity
+//     * @param queryId
+//     * @param catalogName
+//     * @param tableName
+//     * @param queryPassthroughArguments
+//     */
+//    @JsonCreator
+//    public GetTableRequest(@JsonProperty("identity") FederatedIdentity identity,
```
burhan94, line 74: is this supposed to be commented out

burhan94 (changes requested):

AbdulR3hman:
rebased through CLI; it will re-run the checks.

burhan94 (approved):

PR #1770, opened 2024-02-22
initial Athena Clickhouse connector commit related to issue 1754

bishrtabbaa:
*Issue #, if available:*
1754

*Description of changes:*
Athena Clickhouse connector reusing Athena MySQL code and Clickhouse driver


athena-clickhouse/src/main/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseUtil.java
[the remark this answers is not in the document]
bishrtabbaa: Good question ... its a quirky artifact of Clickhouse supporting the MySQL wire protocol.

https://github.com/ClickHouse/ClickHouse/discussions/33548
https://github.com/ClickHouse/dbt-clickhouse/issues/74
https://github.com/ClickHouse/ClickHouse/issues/9336
https://dev.mysql.com/doc/refman/8.0/en/identifier-case-sensitivity.html

The goal was to reuse the maximum amount of Athena MySQL connector code and also maximize compatibility using the Clickhouse native driver.

athena-clickhouse/.aws-sam/build.toml
chngpe, line 1: Please remove file generated by SAM  CLI

athena-clickhouse/athena-clickhouse.yaml
```diff
@@ -0,0 +1,153 @@
...
+  LambdaRoleARN:
+    Description: "(Optional) A custom role to be used by the Connector lambda"
+    Type: String
+    Default: ""
+  DisableSpillEncryption:
+    Description: 'If set to ''false'' data spilled to S3 is encrypted with AES GCM'
+    Default: 'false'
+    Type: String
+  SecurityGroupIds:
+    Description: 'One or more SecurityGroup IDs corresponding to the SecurityGroup that should be applied to the Lambda function. (e.g. sg1,sg2,sg3)'
```
chngpe, line 50: are VPCs required for clickhouse DB?

chngpe (changes requested):
Majority of code are basically copy over from MySQL connectors. Can we just extends from MySQL connector instead of JDBC?

For example, [ClickHouseMetadataHandler.java](https://github.com/awslabs/aws-athena-query-federation/pull/1770/files#diff-9f10ed91eb325ec30478ee1c48889e85b12333f1119d988f6c12b985f4e4deab) extends from MySqlMetadataHandler so we can avoid duplicate code and leave the part that is necessary. 

For example, https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-redshift/src/main/java/com/amazonaws/athena/connectors/redshift/RedshiftMetadataHandler.java which extends from `PostGreSqlMetadataHandler` just override jdbc driver class and default port in constructor. (along with some necessary method for list paginated tables query)

athena-clickhouse/athena-clickhouse.yaml
bishrtabbaa: Yes, if self-managed.

athena-clickhouse/.aws-sam/build.toml
bishrtabbaa: Done.

athena-clickhouse/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+            <version>2022.47.1</version>
+        </dependency>
+        <!--
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
+            <version>${protobuf3.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.mysql</groupId>
```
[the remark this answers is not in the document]
bishrtabbaa: Done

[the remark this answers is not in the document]
bishrtabbaa: [ClickHouseMeta](https://github.com/awslabs/aws-athena-query-federation/pull/1770/files#diff-9f10ed91eb325ec30478ee1c48889e85b12333f1119d988f6c12b985f4e4deab) now extends `MySqlMetadataHandler`.

athena-clickhouse/pom.xml
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="utf-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>aws-athena-query-federation</artifactId>
+        <groupId>com.amazonaws</groupId>
+        <version>2022.47.1</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>athena-clickhouse</artifactId>
+    <version>2024.10.1</version>
```
aimethed, line 10: The version number should be 2022.47.1 to be in line with all other connectors

aimethed: The same goes for the dependencies on other connectors. ex: [athena-cloudera-impala](https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-cloudera-impala/pom.xml)

athena-clickhouse/src/main/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseMetadataHandler.java
```diff
@@ -0,0 +1,253 @@
...
+
+import static com.amazonaws.athena.connectors.jdbc.manager.JdbcMetadataHandler.TABLES_AND_VIEWS;
+
+/**
+ * Class handles metadata for ClickHouse and reuses code from Athena MySQL module for compatibility. User must have access to `schemata`, `tables`, `columns`, `partitions` tables in INFORMATION_SCHEMA.
+ */
+public class ClickHouseMetadataHandler
+        extends MySqlMetadataHandler
+{
+    static final Map<String, String> JDBC_PROPERTIES = ImmutableMap.of("databaseTerm", "SCHEMA");
```
aimethed, line 71: Is there a reason to not be using the ClickHouseConstants class here for this?

chngpe (dismissed):

athena-clickhouse/src/main/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseRecordHandler.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public PreparedStatement buildSplitSql(Connection jdbcConnection, String catalogName, TableName tableName, Schema schema, Constraints constraints, Split split)
+            throws SQLException
+    {
+        PreparedStatement preparedStatement = jdbcSplitQueryBuilder.buildSql(jdbcConnection, null, tableName.getSchemaName(), tableName.getTableName(), schema, constraints, split);
+
+        // Disable fetching all rows.
+        preparedStatement.setFetchSize(Integer.MIN_VALUE);
```
chngpe, line 99: Why are we disabling fetch all rows?

chngpe: Do we want to retrieve it row by row?  per Doc:https://dev.mysql.com/doc/connector-j/en/connector-j-reference-implementation-notes.html

```
To enable this functionality, create a Statement instance in the following manner:

stmt = conn.createStatement(java.sql.ResultSet.TYPE_FORWARD_ONLY,
              java.sql.ResultSet.CONCUR_READ_ONLY);
stmt.setFetchSize(Integer.MIN_VALUE);
The combination of a forward-only, read-only result set, with a fetch size of Integer.MIN_VALUE serves as a signal to the driver to stream result sets row-by-row. After this, any result sets created with the statement will be retrieved row-by-row.
```

bishrtabbaa: Correct.  I modeled the code on what I observed in the MySQL connector itself as well as other similar relational ones (e.g. PostgreSQL, RedShift, Oracle, etc).

athena-clickhouse/src/main/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseMetadataHandler.java
bishrtabbaa: Corrected now.  Thanks.

athena-clickhouse/pom.xml
bishrtabbaa: Corrected now.  Thanks.

athena-clickhouse/athena-clickhouse.yaml
```diff
@@ -71,7 +77,9 @@ Resources:
           default: !Ref DefaultConnectionString
       FunctionName: !Ref LambdaFunctionName
       Handler: "com.amazonaws.athena.connectors.clickhouse.ClickHouseMuxCompositeHandler"
-      CodeUri: "./target/athena-clickhouse-2022.47.1.jar"
+      CodeUri:
```
chngpe, line 80: We don't need to do that as SAR can handle this.

aimethed (approved):

PR #1819, opened 2024-03-15
Extended QPT to athena-dynamodb

AbdulR3hman:
*Description of changes:*
Will allow customer to execute QPT on DynamoDB using PartiQL; Only SELECT statements are supported.

### DIFF
--- a/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
+++ b/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
@@ -105,6 +105,7 @@ import static com.amazonaws.athena.connectors.dynamodb.constants.DynamoDBConstan
 import static com.amazonaws.athena.connectors.dynamodb.constants.DynamoDBConstants.SEGMENT_ID_PROPERTY;
 import static com.amazonaws.athena.connectors.dynamodb.constants.DynamoDBConstants.TABLE_METADATA;
 import static com.amazonaws.athena.connectors.dynamodb.throttling.DynamoDBExceptionFilter.EXCEPTION_FILTER;
+import static com.amazonaws.athena.connectors.dynamodb.util.DDBTableUtils.SCHEMA_INFERENCE_NUM_RECORDS;
 
 /**
  * Handles metadata requests for the Athena DynamoDB Connector.
@@ -262,7 +263,7 @@ public class DynamoDBMetadataHandler
         ExecuteStatementRequest executeStatementRequest =
                 ExecuteStatementRequest.builder()
                         .statement(partiQLStatement)
-                        .limit(4)
+                        .limit(SCHEMA_INFERENCE_NUM_RECORDS)
                         .build();
         //PartiQL on DynamoDB Doesn't allow a dry run; therefore, we look "Peek" over the first few records
         ExecuteStatementResponse response = ddbClient.executeStatement(executeStatementRequest);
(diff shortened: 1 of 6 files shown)


athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
```diff
@@ -230,6 +250,27 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
...
+        if (!request.isQueryPassthrough()) {
+            throw new IllegalArgumentException("No Query passed through [{}]" + request);
+        }
+
+        queryPassthrough.verify(request.getQueryPassthroughArguments());
+        String partiQLStatement = request.getQueryPassthroughArguments().get(DDBQueryPassthrough.QUERY);
+        ExecuteStatementRequest executeStatementRequest =
+                ExecuteStatementRequest.builder()
+                        .statement(partiQLStatement)
+                        .limit(4)
```
chngpe, line 265: use `SCHEMA_INFERENCE_NUM_RECORDS` from DDBTableUtils instead?

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/qpt/DDBQueryPassthrough.java
```diff
@@ -0,0 +1,93 @@
...
+        if (!upperCaseStatement.startsWith("SELECT")) {
+            throw new UnsupportedOperationException("Statement does not start with SELECT.");
+        }
+
+        // List of disallowed keywords
+        Set<String> disallowedKeywords = ImmutableSet.of("INSERT", "UPDATE", "DELETE", "CREATE", "DROP", "ALTER");
+
+        // Check if the statement contains any disallowed keywords
+        for (String keyword : disallowedKeywords) {
+            if (upperCaseStatement.contains(" " + keyword + " ") || upperCaseStatement.startsWith(keyword + " ")) {
```
chngpe, line 88: can we have just contains here? or are we assuming there will be column name as `insert`  here? Those are reserved word in DDB, so i think should be okay. https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ReservedWords.html

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTableUtils.java
```diff
@@ -187,6 +175,24 @@ public static Schema peekTableForSchema(String tableName, ThrottlingInvoker invo
         return schemaBuilder.build();
     }
 
+    public static SchemaBuilder buildSchemaFromItems(List<Map<String, AttributeValue>> items)
+    {
+        SchemaBuilder schemaBuilder = new SchemaBuilder();
+        Set<String> discoveredColumns = new HashSet<>();
+        for (Map<String, AttributeValue> item : items) {
+            for (Map.Entry<String, AttributeValue> column : item.entrySet()) {
+                if (!discoveredColumns.contains(column.getKey())) {
```
chngpe, line 184: can we improve it to not have this many layer?

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
AbdulR3hman: good catch; yes will do.

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/qpt/DDBQueryPassthrough.java
AbdulR3hman: will make the fix. and yes I don't think insert is allowed as its reserved keyword.

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBTableUtils.java
AbdulR3hman: Yes, I'll refactor it.

chngpe (approved):

PR #1821, opened 2024-03-15
Introducing CFN Templates for all Connectors, that are configured to use Glue Connections

akuzin1:
*Description of changes:*
Have modified CloudFormation templates to take in Glue Connection as new parameter, while removing other environment variables that are not required for configuring the execution role of connector.


ejeffrli:
We have templates for glue connections merged already. Example: https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-mysql/athena-mysql-connection.yaml

PR #1822, opened 2024-03-15
Extended QPT To Datalakegen2

AbdulR3hman:
*Issue #, if available:*

*Description of changes:*
Will allow customer to execute QPT on top of Datalakegen2


athena-datalakegen2/src/main/java/com/amazonaws/athena/connectors/datalakegen2/DataLakeGen2MetadataHandler.java
```diff
@@ -197,6 +204,68 @@ public GetTableResponse doGetTable(final BlockAllocator blockAllocator, final Ge
...
+
+                ArrowType columnType = JdbcArrowTypeConverter.toArrowType(
+                        metadata.getColumnType(columnIndex),
+                        precision,
+                        scale,
+                        configOptions);
+                String dataType = metadata.getColumnTypeName(columnIndex);
+
+                if (dataType != null && DataLakeGen2DataType.isSupported(dataType)) {
+                    columnType = DataLakeGen2DataType.fromType(dataType);
```
chngpe, line 242: The rest of the code is the same as JDBCMetadatahandler only this is different. Can we abstract this method? and override in child so we can avoid duplicate code

chngpe (changes requested):

chngpe (approved):

PR #1849, opened 2024-03-19
Enable secret_name variable to be passed as environment variable

ejeffrli:
*Issue #, if available:*

*Description of changes:*
Embed SECRET_NAME value into connection string if passed as environment variable before resolving secret. That way secrets are still utilizing our cache.


athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBRecordHandler.java
```diff
@@ -106,14 +108,22 @@ protected DocDBRecordHandler(AmazonS3 amazonS3, AWSSecretsManager secretsManager
...
+        if (configOptions.containsKey(SECRET_NAME) && !hasEmbeddedSecret(connStr)) {
+            connStr = connStr.substring(0, 10) + "${" + configOptions.get(SECRET_NAME) + "}@" + connStr.substring(10);
+        }
+        String endpoint = resolveSecrets(connStr);
         return connectionFactory.getOrCreateConn(endpoint);
     }
 
+    private boolean hasEmbeddedSecret(String connStr) 
+    {
+        return connStr.charAt(11) == '$';
```
AbdulR3hman, line 124: are we always sure its going to be at the 11th character?

PR #1853, opened 2024-03-21
Qpt vertica changes

Trianz-Akshay:
*Issue #, if available:*

*Description of changes:*
Query passthrough changes for vertica connector


athena-vertica/pom.xml
```diff
@@ -66,6 +66,12 @@
             <artifactId>ST4</artifactId>
             <version>${antlr.st4.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>athena-jdbc</artifactId>
+            <version>2022.47.1</version>
```
AbdulR3hman, line 72: please don't explicitly chose the version. It should be done by maven.

athena-vertica/src/main/java/com/amazonaws/athena/connectors/vertica/VerticaMetadataHandler.java
```diff
@@ -174,6 +203,79 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
...
+    {
+        ImmutableMap.Builder<String, List<OptimizationSubType>> capabilities = ImmutableMap.builder();
+        queryPassthrough.addQueryPassthroughCapabilityIfEnabled(capabilities, configOptions);
+
+        return new GetDataSourceCapabilitiesResponse(request.getCatalogName(), capabilities.build());
+    }
+
+
+    @Override
+    public GetTableResponse doGetQueryPassthroughSchema(final BlockAllocator blockAllocator, final GetTableRequest getTableRequest)
```
AbdulR3hman, line 222: The implementation of this method brings up few issues;

1. We are using JDBC Implementation here; (almost copy/paste from the JDBC Metadatahandler) whereas `doGetTable` uses `VerticaSchemaUtils` class. 
2. This brings up an important question; why are we not using the JDBC metadatahandler to begin with? If there is a good reason for it, then we shouldn't be using the JDBC implementation here either.
3. If it seems that there in fact we don't need to have different implementation between JDBC metadata handler and Vertica's, we need to use this opportunity to unify the code.

AbdulR3hman (dismissed):
please look over the comments and let me know.

athena-vertica/src/main/java/com/amazonaws/athena/connectors/vertica/VerticaMetadataHandler.java
Trianz-Akshay: incorporated the changes for vertica datatype to arrow in dogetQuerypassthroughSchema.

athena-vertica/src/main/java/com/amazonaws/athena/connectors/vertica/VerticaMetadataHandler.java
```diff
@@ -243,20 +316,29 @@ public void getPartitions(BlockWriter blockWriter, GetTableLayoutRequest request
...
         VerticaExportQueryBuilder queryBuilder = queryFactory.createVerticaExportQueryBuilder();
+        String preparedSQLStmt;
 
-        String preparedSQLStmt = queryBuilder.withS3ExportBucket(s3ExportBucket)
-                .withQueryID(queryID)
-                .withColumns(definition, schemaName)
-                .fromTable(tableName.getSchemaName(), tableName.getTableName())
-                .withConstraints(constraints, schemaName)
-                .build();
+        if (!request.getTableName().getQualifiedTableName().equalsIgnoreCase(queryPassthrough.getFunctionSignature())) {
```
aimethed, line 326: I think a better way to check if a request is a query passthrough request is using: `request.getConstraints().isQueryPassThrough()`

aimethed (commented):
Some minor changes/suggestions I noticed.

athena-vertica/src/main/java/com/amazonaws/athena/connectors/vertica/VerticaMetadataHandler.java
```diff
@@ -174,6 +202,51 @@ public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesReque
...
+    @Override
+    public GetTableResponse doGetQueryPassthroughSchema(final BlockAllocator blockAllocator, final GetTableRequest getTableRequest)
+            throws Exception
+    {
+        if (!getTableRequest.isQueryPassthrough()) {
+            throw new IllegalArgumentException("No Query passed through [{}]" + getTableRequest);
+        }
+
+        queryPassthrough.verify(getTableRequest.getQueryPassthroughArguments());
+        String customerPassedQuery = getTableRequest.getQueryPassthroughArguments().get(JdbcQueryPassthrough.QUERY);
```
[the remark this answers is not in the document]
Trianz-Akshay: changed to VerticaQueryPassthrough.QUERY

aimethed (approved):

PR #1860, opened 2024-03-22
Glue Connection Connector Integration 

akuzin1:
*Description of changes:*
Integrating connectors to be able to supplement connection properties from glue if connection is provided as environment variable `glue_connection`.
```
Example of how it may look in connector: 
glue_connection: "datasource-connection"
```
Introduced util class, reaches out to Glue, parses connection properties into a config map and supplements it to the connectors handlers.


ejeffrli:
We support glue connections using https://github.com/awslabs/aws-athena-query-federation/blob/0a795c356e8a117a800f303393002c0a3a0668a3/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/connection/EnvironmentProperties.java#L50 instead of this PR

PR #1877, opened 2024-03-28
Updating Teradata connector to accept both 'partitioncount' and 'partition_count' from Config

akuzin1:
*Description of changes:*
This change adds backwards compatibility, once connector is integrated to supplement properties from glue connection. Connection properties follow a snakecase naming convention, therefore, to follow that `partitioncount` was renamed to `partition_count`. As well as added default value.


ejeffrli:
https://github.com/awslabs/aws-athena-query-federation/blob/0a795c356e8a117a800f303393002c0a3a0668a3/athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataMetadataHandler.java#L247

We have already merged and squashed this commit.

PR #1920, opened 2024-04-16
Query Passthrough for Redis Connector

aimethed:
*Description of changes:*
Query Passthrough changes for the Redis connector. Works by taking in 3 arguments, the script, keys, and argv. The script is a lua script passed as a string. The keys and argv arguments are strings representing arrays of additional arguments for the script. These three arguments are then passed to the Lettuce client eval_ro command. See the [redis Lua scripting documentation](https://redis.io/docs/latest/develop/interact/programmability/eval-intro/) for more details.

Because query passthrough works by running Lua scripts, query passthrough may not work for clustered instances due to the limitations around a Redis Lua script accessing data from multiple hashes.

Additionally adds 4 additional lambda environment variables for storing QPT connection information since QPT does not access any glue table so cannot retrieve glue table metadata.

Examples:
```
SELECT * FROM TABLE(system.script(
    script => 'return redis.call("GET", ARGV[1])',
    keys => '[]',
    argv => '[testKey]'
));
```
```
SELECT * FROM TABLE(system.script(
    'return {redis.call(''ZSCAN'', KEYS[1], ARGV[1]), redis.call(''ZSCAN'', KEYS[2], ARGV[1])}',
    '[zset:numbers, zset:letters]',
    '[20]'
));
```

### DIFF
--- a/athena-redis/athena-redis.yaml
+++ b/athena-redis/athena-redis.yaml
@@ -49,6 +49,22 @@ Parameters:
     Description: "(Optional) An IAM policy ARN to use as the PermissionsBoundary for the created Lambda function's execution role"
     Default: ''
     Type: String
+  QPTConnectionEndpoint:
+    Description: "(Optional) The hostname:port:password of the Redis server that contains data for this table optionally using SecretsManager (e.g. ${secret_name}). Used for Query Pass Through queries only."
+    Default: ''
+    Type: String
+  QPTConnectionSSL:
+    Description: "(Optional) When True, creates a Redis connection that uses SSL/TLS. Used for Query Pass Through queries only."
+    Default: 'false'
+    Type: String
+  QPTConnectionCluster:
+    Description: "(Optional) When True, enables support for clustered Redis instances. Used for Query Pass Through queries only."
+    Default: 'false'
+    Type: String
+  QPTConnectionDBNumber:
+    Description: "(Optional) Set this number (for example 1, 2, or 3) to read from a non-default Redis database. Used for Query Pass Through queries only."
+    Default: 0
+    Type: Number
 Conditions:
   HasPermissionsBoundary: !Not [ !Equals [ !Ref PermissionsBoundaryARN, "" ] ]
 Resources:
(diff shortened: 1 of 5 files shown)


athena-redis/athena-redis.yaml
```diff
@@ -49,6 +49,22 @@ Parameters:
...
+    Default: 'false'
+    Type: String
+  QPTConnectionCluster:
+    Description: "(Optional) When True, enables support for clustered Redis instances. Used for Query Pass Through queries only."
+    Default: 'false'
+    Type: String
+  QPTConnectionDBNumber:
+    Description: "(Optional) Set this number (for example 1, 2, or 3) to read from a non-default Redis database. Used for Query Pass Through queries only."
+    Default: 0
+    Type: Number
```
akuzin1, line 67: Not sure if this has been discussed or not, but do these properties need to be added to Glue Connection object? How will these properties be incorporated with the revamped console experience?

aimethed: I think this is just a small part of the larger discussion that needs to be had about how redis will work with glue connections, since non-qpt redis uses the glue table metadata for connection information instead of lambda environment variables.

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/RedisRecordHandler.java
```diff
@@ -131,20 +136,79 @@ private RedisConnectionWrapper<String, String> getOrCreateClient(String rawEndpo
...
+
+        RedisCommandsWrapper<String, String> syncCommands = getSyncCommands(recordsRequest);
+
+        String script = queryPassthroughArgs.get(RedisQueryPassthrough.SCRIPT);
+        byte[] scriptBytes = script.getBytes();
+
+        String keys = queryPassthroughArgs.get(RedisQueryPassthrough.KEYS);
+        String[] keysArray = new String[0];
+        if (!keys.isEmpty()) {
+            keysArray = keys.substring(1, keys.length() - 1).split(",\\s*");
```
chngpe, line 185: is it used to remove "?

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/RedisRecordHandler.java
```diff
@@ -205,6 +269,38 @@ private ScanCursor loadKeys(RedisCommandsWrapper<String, String> syncCommands, S
...
+        });
+    }
+
+    /**
+     * Redis eval calls return an object of type T, which is either a String or List<T>.
+     * This flattens it into a String using a StringBuilder.
+     */
+    private void flattenRow(Object value, StringBuilder builder)
+    {
+        if (value instanceof String) {
```
chngpe, line 293: safe guard in case value is null here since recursive

athena-redis/src/main/java/com/amazonaws/athena/connectors/redis/RedisRecordHandler.java
chngpe: can you please comment the usage for this. thanks!

burhan94 (approved):

PR #2019, opened 2024-06-13
Hbase uppercase table

aimethed:
*Description of changes:*
Previously, hbase would fail when trying to access tables with uppercase characters because athena does not support uppercase tables. This change allows querying of hbase uppercase tables by handling uppercase names the same way the dynamodb connector handles this: by getting all tables in the hbase namespace and mapping the normal case name to its lowercase name. This ensures that if multiple hbase tables would have the same lowercase name (ex: Table and tAble would both be table), an error is thrown.


athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseTableNameUtils.java
```diff
@@ -0,0 +1,152 @@
...
+     * (since athena does not support them). If an hbase table name is found with the athena table name, it is returned.
+     * Otherwise, tryCaseInsensitiveSearch is used to find the corresponding hbase table.
+     *
+     * @param tableName the case insensitive table name
+     * @return the hbase table name
+     */
+    public static org.apache.hadoop.hbase.TableName getHbaseTableName(HBaseConnection conn, TableName athTableName)
+            throws IOException
+    {
+        if (conn.tableExists(getQualifiedTable(athTableName))) {
```
chngpe, line 109: Double check, we lower case the athTableName here. If Hbase has Uppercase, conn.getTableExists will able to find that table?

aimethed: This is a check to see if the table we are trying to get is already lowercase. If it is (tableExists), then we can skip the rest of the logic. Otherwise, we run the caseInsensitiveSearch.

athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseMetadataHandler.java
```diff
@@ -197,7 +198,7 @@ public ListTablesResponse doListTables(BlockAllocator blockAllocator, ListTables
         for (int i = 0; i < tables.length; i++) {
             TableName tableName = tables[i];
             tableNames.add(new com.amazonaws.athena.connector.lambda.domain.TableName(request.getSchemaName(),
-                    tableName.getNameAsString().replace(request.getSchemaName() + ":", "")));
+                    tableName.getNameAsString().toLowerCase(Locale.ENGLISH).replace(request.getSchemaName() + ":", "")));
```
chngpe, line 201: can we not do this? so metadata can reflect correctly

athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseTableNameUtils.java
```diff
@@ -0,0 +1,152 @@
...
+            return getQualifiedTable(athTableName);
+        }
+        Optional<org.apache.hadoop.hbase.TableName> caseInsensitiveMatch = tryCaseInsensitiveSearch(conn, athTableName);
+        if (caseInsensitiveMatch.isPresent()) {
+            return caseInsensitiveMatch.get();
+        }
+        else {
+            return null;
```
AbdulR3hman, line 117: I think its a better practice to change this to Optional.Empty();

athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseMetadataHandler.java
aimethed: If the tableName has uppercase characters, not lower casing here would cause an error as Athena TableNames don't support upper case characters.

athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseTableNameUtils.java
```diff
@@ -0,0 +1,152 @@
...
+     * Performs a case insensitive table search by listing all table names in the schema (namespace), mapping them
+     * to their lowercase transformation, and then mapping the given tableName back to a unique table. To prevent ambiguity,
+     * an IllegalStateException is thrown if multiple tables map to the given tableName.
+     * @param conn the HBaseConnection used to retrieve the tables
+     * @param tableName The Athena TableName to find the mapping to
+     * @return The HBase TableName containing the found HBase table and the Athena Schema (namespace)
+     * @throws IOException 
+     */
+    @VisibleForTesting
+    protected static Optional<org.apache.hadoop.hbase.TableName> tryCaseInsensitiveSearch(HBaseConnection conn, TableName tableName)
```
AbdulR3hman, line 131: I would add for the reader of this function is that HBase actually supports case sensitive tables; meaning two tables with the same name can co-exist. And we are attempting to resolve case insensitive `only and only if` there is one table per that string.

athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseMetadataHandler.java
aimethed: Just called, trino will lowercase it, so it is not needed here. Changing

athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseTableNameUtils.java
```diff
@@ -0,0 +1,152 @@
...
+            throws IOException
+    {
+        logger.info("Table {} not found.  Falling back to case insensitive search.", tableName.getTableName());
+        Multimap<String, String> lowerCaseNameMapping = ArrayListMultimap.create();
+        org.apache.hadoop.hbase.TableName[] tableNames = conn.listTableNamesByNamespace(tableName.getSchemaName());
+        for (org.apache.hadoop.hbase.TableName nextTableName : tableNames) {
+            lowerCaseNameMapping.put(nextTableName.getQualifierAsString().toLowerCase(Locale.ENGLISH), nextTableName.getNameAsString());
+        }
+        Collection<String> mappedNames = lowerCaseNameMapping.get(tableName.getTableName());
+        if (mappedNames.size() > 1) {
```
AbdulR3hman, line 141: this needs to be tested; I am worried that an exception will bubble up to the console when loading the schema for the database which contain two tables with the same string (different capitalization)

AbdulR3hman:
I believe we should preserve the original behavior; and put this change behind a feature flag as it will change what existing customers are expecting the default connector behavior.

athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseTableNameUtils.java
aimethed: I think that is covered with "To prevent ambiguity, an IllegalStateException is thrown if multiple tables map to the given tableName."

athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseTableNameUtils.java
aimethed: This exception does cause the query to fail (it was tested). This is the intended behavior.

athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseTableNameUtils.java
aimethed: Just updated the code so no longer has a null return. Now throws an error for 0 matches to mimic docdb behavior

aimethed:
Summary of changes:

- Now requires the following environment variable to be added to enable case insensitive search: `enable_case_insensitive_match` = `true`
- Updated doGetTable to return the name from the case insensitive search instead of the one from the request
  - If it's disabled, just returns the name
- Removed the check for if the table exists. Since this feature is now opt-in, the check causes unwanted behavior.
  - If there are two tables, `table` and `Table`, querying for `Table` before would succeed but match with `table`. Now errors due to multiple possible matches.
- Added a check for if the given name isn't lowercase. Addresses an error where doListTables was called with correctly-cased table name.

ejeffrli (approved):

mschoeni1 (approved):

PR #2022, opened 2024-06-14
unused dependency remove bigquery connector

Trianz-Akshay:
*Issue #, if available:*
 
*Description of changes:*
unused dependency removed from bigquery connector and tested.

### DIFF
--- a/athena-google-bigquery/src/test/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryFederationExpressionParserTest.java
+++ b/athena-google-bigquery/src/test/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryFederationExpressionParserTest.java
@@ -265,7 +265,7 @@ public class BigQueryFederationExpressionParserTest
                 ImmutableList.of(colOne, dateConstantExpression));
 
         String s = bigQueryExpressionParser.parseFunctionCallExpression((FunctionCallExpression) gtFunctionCall);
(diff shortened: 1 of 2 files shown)


athena-google-bigquery/src/test/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryFederationExpressionParserTest.java
```diff
@@ -265,7 +265,7 @@ public void testComplexExpressions_Simple_With_Date()
                 ImmutableList.of(colOne, dateConstantExpression));
 
         String s = bigQueryExpressionParser.parseFunctionCallExpression((FunctionCallExpression) gtFunctionCall);
-        assertEquals("(colOne > '2020-01-05')", bigQueryExpressionParser.parseFunctionCallExpression((FunctionCallExpression) gtFunctionCall));
+        assertEquals("(colOne > '2020-01-04')", bigQueryExpressionParser.parseFunctionCallExpression((FunctionCallExpression) gtFunctionCall));
```
aimethed, line 268: What is the reason for this? Pulled the PR locally and this caused the test to fail. I assume its a mistake?

Trianz-Akshay: It was faiing in my local, so i changed. but let me revert this in this case.

PR #2051, opened 2024-06-26
V2 migration secrets

aimethed:
*Description of changes:*
The base changes for the migration to AWS SDK v2 for java and then AWS Secrets Manager.


aimethed:
Linking to Secrets Manager migration issue https://github.com/awslabs/aws-athena-query-federation/issues/2078

athena-cloudera-impala/src/main/java/com/amazonaws/athena/connectors/cloudera/ImpalaFederationExpressionParser.java
```diff
@@ -17,7 +17,7 @@
  * limitations under the License.
  * #L%
  */
-package com.amazonaws.athena.connectors.hortonworks;
+package com.amazonaws.athena.connectors.cloudera;
```
chngpe, line 20: why did it change package?

aimethed: For some reason, it was in the hortonworks package even though it is for cloudera. I assume it was someone's mistake when copy pasting.

aimethed: There are a few files in the impala connector where it had to be imported because the package name was wrong. Ex: athena-cloudera-impala/src/main/java/com/amazonaws/athena/connectors/cloudera/ImpalaRecordHandler.java

chngpe (approved):

PR #2068, opened 2024-07-03
BigQuery empty table and projectId casing issues fix

VenkatasivareddyTR:
[BigQueryIssue.docx](https://github.com/user-attachments/files/16085413/BigQueryIssue.docx)
*Issue #, if available:*
Issue 1: Select count(*) on empty table returns a java exception
Details - They ran a query - SELECT COUNT(*) FROM SOME_EMPTY_TABLE_IN_BIG_QUERY caused a Java ILLEGALSTATEEXCEPTION

Issue 2: Connector does not update the case of projectId

*Description of changes:*
If table doesn't have any data, then we are throwing IllegalStateException. This issue is fixed with these changes, If table doesn't have any data, we will return zero and I have attached the test results as well. 
And projectId casing issue also fixed in this PR, as per official documentation big query projectId should contain only lowercase letters, numbers, and hyphens. So, we are converting projectId to lowercase to solve projectId casing issue.
Ref: https://cloud.google.com/resource-manager/docs/creating-managing-projects 

### DIFF
--- a/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryRecordHandler.java
+++ b/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryRecordHandler.java
@@ -127,7 +127,7 @@ public class BigQueryRecordHandler
                                      List<QueryParameterValue> parameterValues,
                                      BigQuery bigQueryClient) throws Exception
     {
-        String projectName = configOptions.get(BigQueryConstants.GCP_PROJECT_ID);
+        String projectName = configOptions.get(BigQueryConstants.GCP_PROJECT_ID).toLowerCase();
         String datasetName = fixCaseForDatasetName(projectName, recordsRequest.getTableName().getSchemaName(), bigQueryClient);
         String tableName = fixCaseForTableName(projectName, datasetName, recordsRequest.getTableName().getTableName(), bigQueryClient);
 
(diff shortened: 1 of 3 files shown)


athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryRecordHandler.java
```diff
@@ -127,7 +127,7 @@ private void handleStandardQuery(BlockSpiller spiller,
                                      List<QueryParameterValue> parameterValues,
                                      BigQuery bigQueryClient) throws Exception
     {
-        String projectName = configOptions.get(BigQueryConstants.GCP_PROJECT_ID);
```
AbdulR3hman, line 130: have we tested if BigQuery treats those project IDs sensitively?  meaning are they case sensitive?

athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryRecordHandler.java
```diff
@@ -263,7 +263,13 @@ private void getTableData(BlockSpiller spiller, ReadRecordsRequest recordsReques
                 // data available.  If no sessions are available for an anonymous (cached) table, consider
                 // writing results of a query to a named table rather than consuming cached results
                 // directly.
-                Preconditions.checkState(session.getStreamsCount() > 0);
```
AbdulR3hman, line 266: if we were throwing an exception before; are we returning zero now? 

how would that work for non count queries; meaning `select * from table` but returns 0 rows?

athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryRecordHandler.java
VenkatasivareddyTR: Hi @AbdulR3hman, yes, I have tested and please find attached testing doc for your reference. We can't create project with uppercase letters as per official documentation. Please refer attached document for more details.
[GBQ Project Id casing issue testing.docx](https://github.com/user-attachments/files/16140020/GBQ.Project.Id.casing.issue.testing.docx)

AbdulR3hman (approved):

PR #2075, opened 2024-07-09
[FEATURE] Add Support for querying Avro/Protobuf Data in athena-msk Connector

Jithendar12:
Support for querying Avro and Protobuf data from Amazon MSK.

Prerequisite: Define the Avro/Protobuf Schema in AWS Glue Schema Registry, select Data Format as Avro/Protocol Buffers while defining schema.
Note: Provide Glue Schema Name as MSK Topic Name with same case.
Example avro/protobuf schema in glue :
```
Avro:

{
  "type": "record",
  "namespace": "ABC_Organization",
  "name": "Employee",
  "fields": [
    {
      "name": "Name",
      "type": "string"
    },
    {
      "name": "Age",
      "type": "int"
    },
    {
      "name": "IsActive",
      "type": "boolean"
    },
    {
      "name": "Balance",
      "type": "double"
    },
    {
      "name": "Rating",
      "type": "float"
    }
  ]
}

Protobuf:

message protobuftest {
  optional string name = 1;
  optional int64 calories = 2;
  optional string colour = 3;
  optional bool isOrganic = 4;
  optional double price = 5;
  optional float weight = 6;
  optional int32 quantity = 7;
}
```

### DIFF
--- a/athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskMetadataHandler.java
+++ b/athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskMetadataHandler.java
@@ -37,6 +37,7 @@ import com.amazonaws.athena.connector.lambda.metadata.ListSchemasResponse;
 import com.amazonaws.athena.connector.lambda.metadata.ListTablesRequest;
 import com.amazonaws.athena.connector.lambda.metadata.ListTablesResponse;
 import com.amazonaws.athena.connector.util.PaginatedRequestIterator;
+import com.amazonaws.athena.connectors.msk.dto.AvroTopicSchema;
 import com.amazonaws.athena.connectors.msk.dto.SplitParameters;
 import com.amazonaws.athena.connectors.msk.dto.TopicPartitionPiece;
 import com.amazonaws.athena.connectors.msk.dto.TopicSchema;
@@ -50,6 +51,8 @@ import com.amazonaws.services.glue.model.ListSchemasResult;
 import com.amazonaws.services.glue.model.RegistryId;
 import com.amazonaws.services.glue.model.RegistryListItem;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.protobuf.Descriptors;
+import io.confluent.kafka.schemaregistry.protobuf.ProtobufSchema;
 import org.apache.arrow.vector.types.pojo.Field;
 import org.apache.arrow.vector.types.pojo.FieldType;
 import org.apache.arrow.vector.types.pojo.Schema;
@@ -68,7 +71,9 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static com.amazonaws.athena.connector.lambda.metadata.ListTablesRequest.UNLIMITED_PAGE_SIZE_VALUE;
+import static com.amazonaws.athena.connectors.msk.AmazonMskConstants.AVRO_DATA_FORMAT;
 import static com.amazonaws.athena.connectors.msk.AmazonMskConstants.MAX_RECORDS_IN_SPLIT;
+import static com.amazonaws.athena.connectors.msk.AmazonMskConstants.PROTOBUF_DATA_FORMAT;
 
 public class AmazonMskMetadataHandler extends MetadataHandler
 {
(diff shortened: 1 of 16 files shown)


athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskMetadataHandler.java
```diff
@@ -419,29 +432,67 @@ private Schema getSchema(String glueRegistryName, String glueSchemaName) throws
...
-            Field field = new Field(it.getName(), fieldType, null);
-            schemaBuilder.addField(field);
-        });
-
-        // Putting the additional schema level information into the metadata in ArrowType schema.
-        schemaBuilder.addMetadata("dataFormat", topicSchema.getMessage().getDataFormat());
+        String dataFormat = registryReader.getGlueSchemaType(glueRegistryName, glueSchemaName);
+        if (dataFormat.equalsIgnoreCase(AVRO_DATA_FORMAT)) {
+            AvroTopicSchema avroTopicSchema = registryReader.getGlueSchema(glueRegistryName, glueSchemaName, AvroTopicSchema.class);
+            // Creating ArrowType for each fields in the topic schema.
```
AbdulR3hman, line 438: nit; why copy the comments exactly as is? 

There is a difference between the first if condition and the else at the end. maybe it would serve the reader if the comments are clearer;

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskRecordHandler.java
```diff
@@ -80,35 +87,90 @@ public void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recordsR
...
+        TopicPartition partition = new TopicPartition(splitParameters.topic, splitParameters.partition);
+        Collection<TopicPartition> partitions = com.google.common.collect.ImmutableList.of(partition);
+        GlueRegistryReader registryReader = new GlueRegistryReader();
 
-        // Initiate new KafkaConsumer that MUST not belong to any consumer group.
-        try (Consumer<String, TopicResultSet> kafkaConsumer = AmazonMskUtils.getKafkaConsumer(recordsRequest.getSchema(), configOptions)) {
-            // Set which topic and partition we are going to read.
-            TopicPartition partition = new TopicPartition(splitParameters.topic, splitParameters.partition);
-            Collection<TopicPartition> partitions = com.google.common.collect.ImmutableList.of(partition);
+        String dataFormat = registryReader.getGlueSchemaType(recordsRequest.getTableName().getSchemaName(), recordsRequest.getTableName().getTableName());
```
AbdulR3hman, line 95: this method is getting really long; we have now three conditions; can we adopt a better pattern; split out the smaller logic into their own functions; if it is possible; extract common code to be used across all three conditions.

AbdulR3hman: Moreover; I believe we can adopt an [strategy pattern](https://www.baeldung.com/java-strategy-pattern) here and even extract all the `xxxConsume` & `xxxExecute` outside of this class to make it easier to read; most important maintain.

If we also have to introduce a new type we can also do this much easier.

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskMetadataHandler.java
Jithendar12: Hi @AbdulR3hman , removed.

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskRecordHandler.java
Jithendar12: Hi @AbdulR3hman , i have moved these consumers to other package.

AbdulR3hman (approved):

athena-msk/src/main/java/com/amazonaws/athena/connectors/msk/AmazonMskRecordHandler.java
AbdulR3hman: this is so much cleaner; easier to read and reason about; thank you.

AbdulR3hman:
@Jithendar12 it seems we can conflicts; if you can rebase and push again it would great; thank you

Jithendar12:
Hi @AbdulR3hman , rebased and pushed. Thank you

AbdulR3hman (approved):

PR #2077, opened 2024-07-09
V2 migration lambda

aimethed:
https://github.com/awslabs/aws-athena-query-federation/issues/2076

*Description of changes:*
Update all uses of Lambda SDK to v2.
Part of the effort to migrate to AWS SDK v2 for java.

Previously, a `LambdaFunctionException` was thrown by SerDe with a deprecated constructor that allowed setting the stack trace. This is no longer supported in v2, so the stack trace is added in the message.

In the sdk tools, declaring lambda interfaces is a v1 feature not supported in v2. These interfaces were removed and replaced with direct calls.


athena-federation-sdk-tools/src/main/java/com/amazonaws/athena/connector/validation/FederationService.java
```diff
@@ -1,30 +0,0 @@
-/*-
```
chngpe, line 1: Why is this file removed?

athena-federation-sdk-tools/src/main/java/com/amazonaws/athena/connector/validation/FederationServiceProvider.java
```diff
@@ -45,60 +48,58 @@ public class FederationServiceProvider
 
     private static final String VALIDATION_SUFFIX = "_validation";
 
-    private static final Map<String, FederationService> serviceCache = new ConcurrentHashMap<>();
+    private static final Map<String, Integer> serdeVersionCache = new ConcurrentHashMap<>();
```
chngpe, line 51: We should probably caching lambdaClient client instead of serDeVersion?

athena-federation-sdk-tools/src/main/java/com/amazonaws/athena/connector/validation/FederationService.java
aimethed: Interface definitions have been removed for lambda v2, so FederationServiceProvider.java updated and interfaces removed.

chngpe (approved):

Issue #507, opened 2021-10-21
[FEATURE] Add a lambda flag environment variable, translate_object_id, to athena-docdb, to control translating queries on _id properties


theFlyRoper:
**Problem**
When querying MongoDB, I can't filter or join on properties of type ObjectId, particularly `_id` properties.  Per documentation, all documents will contain an `_id` property, which will always be stored as a mongodb object id.  Right now, the query engine presents these values as strings.  Athena can display these values but filters on them don't work, because in MongoDB all queries that filter on `_id` must translate strings into the ObjectId value.

**Proposal**
The best way to translate a string into an objectId value is to use the mongodb internal function `ObjectId()`.  So, I propose a lambda environment variable that tells the function to convert any comparisons to an `_id` property to wrap whatever is being compared in that function.  

For example, an athena query like this:
`select *
from mongocollection
where _id = '444444d8e5930afaa0c92332'`
would convert to a mongodb findMany query, thus:
 `db.mongocolletion.findMany({ "_id": ObjectId('444444d8e5930afaa0c92332')})`

Controlling this with a flag would prevent breaking changes for existing users.

**Describe alternatives you've considered**
Most alternatives would involve changing or adding properties to documents, which I would prefer not to do.  There is no glue advanced type that covers this use case, and no way to define a custom type to translate this value.

AbdulR3hman:
this is now addressed in the latest release.

PR #2105, opened 2024-07-15
DocDB select query with _id issue fix

VenkatasivareddyTR:
*Issue #, if available:*
https://github.com/awslabs/aws-athena-query-federation/issues/507 
*Description of changes:*
Converting String to ObjectId when querying using _id column in docdb.
In DocumentDB _id column data type is ObjectId when we query with _id as String it won't return any results so converting String to ObjectId will return results. PFA testing doc for the same.
[DocumentDB test with _id.docx](https://github.com/user-attachments/files/16233803/DocumentDB.test.with._id.docx)


AbdulR3hman (changes requested):
can you please add a test case to make sure we track this correctly; thanks

Trianz-Akshay:
added unit test cases.

AbdulR3hman (approved):

Issue #62, opened 2019-11-22
[BUG] Redshift 'char' column constraint pushdown does not work


abhijeetvg:
Execution with constraint on char column returns empty results. 

Verified that correct value is set in the prepared statement. 

Similar push down works in MySql.

egriffith:
Greetings from 2023, I am fairly certain this bug is still present in the latest Athena Connector. Please let me know if I need to cut a different issue

My Reproduction Steps: 

1) Create a new redshift database, check the box to allow it to load the sample data. 
2) Setup the Athena connector to connect to Redshift's sample_data_dev database.

3) Run the following query in Redshift: 

```
SELECT count(*) as count, i_category
FROM tpcds.catalog_sales JOIN tpcds.item ON i_item_sk=cs_item_sk
                  JOIN tpcds.date_dim ON d_date_sk=cs_sold_date_sk
                  LEFT JOIN tpcds.catalog_returns ON (cs_order_number=cr_order_number 
                                            AND cs_item_sk=cr_item_sk)
WHERE i_category = 'Men'
GROUP BY i_category
order by 1
```

You'll get back 141762 results. 

Run the same query above in Athena using the connector, and you'll get back zero results. The reason is because i_category is padding 47 extra spaces onto the value and Athena is not trimming them out. A query that does work is...

```
SELECT count(*) as count, i_category
FROM tpcds.catalog_sales JOIN tpcds.item ON i_item_sk=cs_item_sk
                  JOIN tpcds.date_dim ON d_date_sk=cs_sold_date_sk
                  LEFT JOIN tpcds.catalog_returns ON (cs_order_number=cr_order_number 
                                            AND cs_item_sk=cr_item_sk)
WHERE i_category = 'Men                                               '
GROUP BY i_category
order by 1
```

or 

```
SELECT count(*) as count, i_category
FROM tpcds.catalog_sales JOIN tpcds.item ON i_item_sk=cs_item_sk
                  JOIN tpcds.date_dim ON d_date_sk=cs_sold_date_sk
                  LEFT JOIN tpcds.catalog_returns ON (cs_order_number=cr_order_number 
                                            AND cs_item_sk=cr_item_sk)
WHERE i_category LIKE 'Men%'
GROUP BY i_category
order by 1
```

However neither option should have to be used. Athena should be trimming the whitespace natively in order to match Redshift's behavior.

AbdulR3hman:
This has been address in the latest release.

PR #2117, opened 2024-07-24
[Fix] Remove Trailing Whitespaces from Char Column in Postgresql/Redshift Connector

Jithendar12:
*Issue #, if available:*
[ [BUG] Redshift 'char' column constraint pushdown does not work #62 ](https://github.com/awslabs/aws-athena-query-federation/issues/62)

*Description of changes:*
PostgreSQL/Redshift database will send data along with trailing whitespaces in CHAR datatype columns. So, removed those spaces using the RTRIM() function while building the SQL query.

### DIFF
--- a/athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
+++ b/athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
@@ -62,6 +62,7 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
```diff
@@ -445,4 +446,31 @@ protected PreparedStatement getMaterializedViewOrExternalTable(Connection connec
...
+    /**
+     * Retrieves the names of columns with the data type 'CHAR' for a specified table in a PostgreSQL/Redshift database.
+     *
+     * @param connection the JDBC connection to the database
+     * @param schema Postgresql/Redshift schema name
+     * @param table Postgresql/Redshift table name
+     * @return a list of column names that have the data type 'CHAR'
+     * @throws SQLException if a database access error occurs
+     */
+    public static List<String> getCharColumns(Connection connection, String schema, String table) throws SQLException
```
aimethed, line 459: Why is this method in the metadata handler and not where it is called in the query string builder?

athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
```diff
@@ -445,4 +446,31 @@ protected PreparedStatement getMaterializedViewOrExternalTable(Connection connec
...
+    {
+        List<String> charColumns = new ArrayList<>();
+        String query = "SELECT column_name " +
+                "FROM information_schema.columns " +
+                "WHERE table_schema = ? AND table_name = ? AND data_type = 'character'";
+        try (PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setString(1, schema);
+            statement.setString(2, table);
+            try (ResultSet resultSet = statement.executeQuery()) {
```
aimethed, line 468: Do we need to make an additional call to the postgresql database to determine this? Is the character datatype gotten from this query different from the datatype we can get from the tableSchema? An apache arrow type has the FieldType for each field in the arrow schema. Is there a case where the FieldType might not align with the type gotten from the postgresql database?

aimethed (changes requested):

athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
Jithendar12: Hi @aimethed , as other metadata related methods are present in this class so added it here.

athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
Jithendar12: Hi @aimethed , yes we make an additional call to postgresql to retreive fixed-length 'char' columns to trim any extra trailing spaces in the data. Yes the datatype obtained from the tableSchema is different which is 'varchar'. Yes it seems apache arrow type does not have a specific data type for fixed-length 'char'.

aimethed (approved):

AbdulR3hman:
This has been addressed in the latest release.

Issue #1306, opened 2023-06-02
[FEATURE] SQL Server Connector expects Dbo (i.e ROOT) privilege work - it lists only the databases created by user who is issqluser, Most prod deployment have DB created by user which is not issqluser


ssenathi:
**Is your feature request related to a problem? If yes, please describe.**
We are trying to query SQL Server where we have given a user who has all the query access, but when we connect thru SQL Server connector i get no databases listed . Upon looking into the code listDatabaseNames function it is filtering the databases where principle_id as user id and sqluser. Most of the production installation wont have this setup and thus make this connector useless.

**Describe the solution you'd like**
listDatabasenames should list all the databases without checking issqluser =1 

**Describe alternatives you've considered**
Alternative is to clone this repo and create own which though works, want to understand why the author has created this filter.

**Additional context**

ssenathi:
The function is in SQLServerMetadataHandler.java and function code is below and issue with the **_highlighted_** one

 private Set<String> listDatabaseNames(final Connection jdbcConnection)
            throws SQLException
    {
        String queryToListUserCreatedSchemas = "select s.name as schema_name from " +
                "sys.schemas s inner join sys.sysusers u **_on u.uid = s.principal_id_** " +
                "where u.issqluser = 1 " +
                "and u.name not in ('sys', 'guest', 'INFORMATION_SCHEMA') " +
                "order by s.name";
        try (Statement st = jdbcConnection.createStatement();
                ResultSet resultSet = st.executeQuery(queryToListUserCreatedSchemas)) {
            ImmutableSet.Builder<String> schemaNames = ImmutableSet.builder();
            while (resultSet.next()) {
                String schemaName = resultSet.getString("schema_name");
                schemaNames.add(schemaName);
            }
            return schemaNames.build();
        }
    }
}

ssenathi:
Sorry closed by mistake

josiahpeters:
@ssenathi  Thank you for opening up this issue. 

We have the same problem. Our databases were created by Windows Auth Accounts not SQL Auth Accounts. This prevents the metadata from being extracted by the lambda.

I can manually use the athena cli tool to look up the metadata in the tables for the database with the dbo schema. However we're prevented from seeing the tables and schema in the Athena Query Console.

Would you be available to share your forked version of the sql server federation lambda that solves this?

hdinizribeiro:
Any news on this topic? I'm having the same issue

aimethed:
[...]

PR #2127, opened 2024-07-31
sqlserver connector listSchema query changes

Trianz-Akshay:
*Issue #, if available:*
#1306 SQL Server Connector expects Dbo (i.e ROOT) privilege work - it lists only the databases created by user who is issqluser, Most prod deployment have DB created by user which is not issqluser

*Description of changes:*
Updated the list schema query to now fetch all user-defined schemas.

### DIFF
--- a/athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
+++ b/athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
@@ -130,7 +130,7 @@ public class SqlServerMetadataHandler extends JdbcMetadataHandler
             "WHERE t.object_id = (select object_id from sys.objects o where o.name = ? " +
             "and schema_id = (select schema_id from sys.schemas s where s.name = ?))";
     static final String VIEW_CHECK_QUERY = "select TYPE_DESC from sys.objects where name = ? and schema_id = (select schema_id from sys.schemas s where s.name = ?)";
-    static final String LIST_PAGINATED_TABLES_QUERY = "SELECT t.name AS \"TABLE_NAME\", s.name AS \"TABLE_SCHEM\" FROM sys.tables t INNER JOIN sys.schemas s ON t.schema_id = s.schema_id WHERE s.name = ? ORDER BY table_name OFFSET ? ROWS FETCH NEXT ? ROWS ONLY;";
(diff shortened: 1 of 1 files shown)


athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
```diff
@@ -507,9 +508,8 @@ private Set<String> listDatabaseNames(final Connection jdbcConnection)
             throws SQLException
     {
         String queryToListUserCreatedSchemas = "select s.name as schema_name from " +
-                "sys.schemas s inner join sys.sysusers u on u.uid = s.principal_id " +
-                "where u.issqluser = 1 " +
-                "and u.name not in ('sys', 'guest', 'INFORMATION_SCHEMA') " +
+                "sys.schemas s " +
+                "where s.name not in ('sys', 'guest', 'INFORMATION_SCHEMA', 'db_accessadmin', 'db_backupoperator', 'db_datareader', 'db_datawriter', 'db_ddladmin', 'db_denydatareader', 'db_denydatawriter', 'db_owner', 'db_securityadmin') " +
```
AbdulR3hman, line 512: where are we generating this list of the `NOT IN` condition; why specifically those schemas would replace the condition `issqluser = 1`?

AbdulR3hman (changes requested):

athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
Trianz-Akshay: Condition issqluser = 1 in previous query is used to filter user-defined schemas. find screenshot below.
sc_schema
But this condition is not true to fetch the screen for non-root users. so I excluded the system defined schema and fetched all other schema. this will be as same as JDBC metadata handler with excluding system schema. tested for non-root user with db_datareader and db_writer permission.

chngpe (approved):

AbdulR3hman:
also approving; just leaving one thing for @Trianz-Akshay to test; is across different versions of SQL Servers; can we confirm if those are the tables that belong to the `u.issqluser = 1` user; to make sure that we have same behavior across different versions of SQL Server (not a blocker for this PR for now);

AbdulR3hman (approved):

Trianz-Akshay:
> also approving; just leaving one thing for @Trianz-Akshay to test; is across different versions of SQL Servers; can we confirm if those are the tables that belong to the `u.issqluser = 1` user; to make sure that we have same behavior across different versions of SQL Server (not a blocker for this PR for now);

I have tested the sql server connector with SQL server 2016 13.00.6300.2.v1 and found it as expected. PFA results.
[sqlserver2016_13.00.6300.2.v1_test.docx](https://github.com/user-attachments/files/16691902/sqlserver2016_13.00.6300.2.v1_test.docx)

PR #2216, opened 2024-08-27
Collate Aware Postgresql String Builder

AbdulR3hman:
*Issue #, if available:*

*Description of changes:*
Postgresql default collate setting differs from Athena's and thus causes issues that results of missing records; for example:   PostgreSQL `'a' < 'A'` &  Trino `'A' < 'a'`. This causes the dynamic filtering range to miss records. 

This change will allow customers to enforce `Collate "C"` to any `varchar` based predicate. 

**warning**: this is only an experimental feature. 

### DIFF
--- a/athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlQueryStringBuilder.java
+++ b/athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlQueryStringBuilder.java
@@ -23,7 +23,10 @@ import com.amazonaws.athena.connector.lambda.domain.Split;
 import com.amazonaws.athena.connector.lambda.domain.predicate.Constraints;
 import com.amazonaws.athena.connectors.jdbc.manager.FederationExpressionParser;
 import com.amazonaws.athena.connectors.jdbc.manager.JdbcSplitQueryBuilder;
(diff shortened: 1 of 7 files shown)


athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlQueryStringBuilder.java
```diff
@@ -42,9 +47,13 @@
 public class PostGreSqlQueryStringBuilder
         extends JdbcSplitQueryBuilder
 {
-    public PostGreSqlQueryStringBuilder(final String quoteCharacters, final FederationExpressionParser federationExpressionParser)
+    private final java.util.Map<String, String> configOptions;
+    private final String postgresqlCollateExperimentalFlag = "postgresql_collate_experimental_flag";
```
aimethed, line 51: I assume this name will change once it is not experimental?

AbdulR3hman: I'll answer offline

burhan94:
Can we add a test case to the integ test package?

ty

aimethed (approved):

PR #2239, opened 2024-09-04
v2 changes for timestream

Trianz-Akshay:
*Description of changes:*
timestreamwrite and timestreamquery update to v2 

### DIFF
--- a/athena-timestream/src/test/java/com/amazonaws/athena/connectors/timestream/TestUtils.java
+++ b/athena-timestream/src/test/java/com/amazonaws/athena/connectors/timestream/TestUtils.java
@@ -19,32 +19,21 @@
  */
 package com.amazonaws.athena.connectors.timestream;
 
-import com.amazonaws.athena.connector.lambda.data.BlockUtils;
-import com.amazonaws.athena.connector.lambda.data.FieldResolver;
-import com.amazonaws.athena.connector.lambda.data.writers.GeneratedRowWriter;
-import com.amazonaws.athena.connector.lambda.data.writers.extractors.Extractor;
-import com.amazonaws.athena.connector.lambda.domain.predicate.ConstraintProjector;
-import com.amazonaws.services.timestreamquery.model.Datum;
-import com.amazonaws.services.timestreamquery.model.QueryResult;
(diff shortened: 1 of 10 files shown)


athena-timestream/src/test/java/com/amazonaws/athena/connectors/timestream/TestUtils.java
```diff
@@ -59,17 +48,17 @@ private TestUtils() {}
...
+    public static QueryResponse makeMockQueryResult(Schema schemaForRead, int numRows)
     {
         return makeMockQueryResult(schemaForRead, numRows, 100, true);
     }
 
-    public static QueryResult makeMockQueryResult(Schema schemaForRead, int numRows, int maxDataGenerationRow, boolean isRandomAZ)
+    public static QueryResponse makeMockQueryResult(Schema schemaForRead, int numRows, int maxDataGenerationRow, boolean isRandomAZ)
     {
-        QueryResult mockResult = mock(QueryResult.class);
+        QueryResponse mockResult = mock(QueryResponse.class);
```
aimethed, line 58: I approved it to build and it failed, likely because of this. With v2, the request and response objects are final or static I believe so you can't mock them. Instead you just build them and return it. So here, instead of mocking QueryResponse.class, you can just create a QueryResponse object and return it.

aimethed (changes requested):

athena-timestream/src/test/java/com/amazonaws/athena/connectors/timestream/TestUtils.java
Trianz-Akshay: The build was successful locally, but it failed in GitHub due to a logger line, as indicated by the logs. I have now addressed the issue by adjusting the logs accordingly. Additionally, I have added the Mockito-inline dependency to handle mocking of final classes, ensuring no test failures occur.

aimethed:
Looks like it is still failing

Trianz-Akshay:
It seems these issues are not being caught in my system, though I resolved it now.

aimethed (approved):

PR #2290, opened 2024-09-24
Oracle date issue fix

VenkatasivareddyTR:
*Issue #, if available:*
Customer has complained that in Oracle DB RDS they are seeing different date (offset by 1) for any dates prior to 1/1/1970. 
**Example**: 
• Database stores 7/20/1967 
o SQL*Plus query: 7/20/1967 
o Athena query: 7/21/1967 
• Database stores 7/20/1972 
o SQL*Plus query: 7/20/1972 
o Athena query: 7/20/1972

*Description of changes:*
This code changes resolves the issue and now we will see correct dates. Please refer attached doc for more details and testing results.

[Oracle Date Type Issue Testin2.docx](https://github.com/user-attachments/files/17115900/Oracle.Date.Type.Issue.Testin2.docx)

### DIFF
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcRecordHandler.java
@@ -82,11 +82,11 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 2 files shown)


AbdulR3hman:
I've done further deep dive into this; and I believe the problem is when the `java.sql.Date` is set to prior 1970 && midnight; it gets changed from UTC to whatever local date the machine is running. For example;

```
public class Main {
    public static void main(String[] args) {
        
        long dateInMillis = -315601200000L; // 1960/01/01
        java.sql.Date sqlDatePre = new java.sql.Date(dateInMillis);

        long daysPre70_Old = TimeUnit.MILLISECONDS.toDays(sqlDatePre.getTime());  //-3652
        long daysPre70_new = LocalDate.parse(sqlDatePre.toString()).toEpochDay(); //-3653

        System.out.println("Before: " + daysPre70_Old);
        System.out.println("After: " + daysPre70_new);
    }
}
```


This solution; affords us to stop the change of the timezone and treats it as a zone-less conversation; which results in not seeing this issue any more;

AbdulR3hman (approved):

PR #2310, opened 2024-10-02
Ecr deployment

aimethed:
*Description of changes:*
This PR adds Dockerfiles to each connector that take the produced jar/zip file and unpack it. The yaml files are then updated to deploy image-based lambdas instead of package based lambdas. This is done by:
* Removing the `Handler`, `CodeUri`, and `Runtime` properties.
* Adding the `PackageType`: `"Image"` property.
* Adding the `ImageUri` property with the uri pointing to the production ECR repository.
This currently works only for non-opt-in regions. An update will be made for opt-in regions to fetch the image from the correct accounts.

### DIFF
--- a/athena-aws-cmdb/Dockerfile
+++ b/athena-aws-cmdb/Dockerfile
@@ -1,0 +1,9 @@
+FROM public.ecr.aws/lambda/java:11
+
+# Copy function code and runtime dependencies from Maven layout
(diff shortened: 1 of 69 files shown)


aimethed:
Included in https://github.com/awslabs/aws-athena-query-federation/pull/2339

PR #2322, opened 2024-10-09
[Fix] athena-sqlserver: Add null check for partition function and partitioning column in getPartitionWhereClauses method

Jithendar12:
This change adds a null check for the partition function and partitioning column in Athena SQL Server connector. It ensures that data will be fetched using specific partition when the necessary partition information is available.

Note: customer's schema has not yet been verified.


aimethed (commented):
Is it possible to add a unit test where we recreate the situation the customer was describing?  `WHERE  $PARTITION.null(null) = 74`

aimethed (approved):

PR #2337, opened 2024-10-14
Feature/postgresql disable predicate pushdown

AbdulR3hman:
*Description of changes:*

Two changes; 

1. Revert https://github.com/awslabs/aws-athena-query-federation/pull/2216
2. Enable Engine hint for postgresql -- to allow the Cx to tell the engine that postgresql is using default collate setting. This will in turn disable predicate pushdown.

### DIFF
--- a/athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
+++ b/athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
@@ -39,6 +39,7 @@ import com.amazonaws.athena.connector.lambda.metadata.optimizations.DataSourceOp
 import com.amazonaws.athena.connector.lambda.metadata.optimizations.OptimizationSubType;
 import com.amazonaws.athena.connector.lambda.metadata.optimizations.pushdown.ComplexExpressionPushdownSubType;
 import com.amazonaws.athena.connector.lambda.metadata.optimizations.pushdown.FilterPushdownSubType;
+import com.amazonaws.athena.connector.lambda.metadata.optimizations.pushdown.HintsSubtype;
 import com.amazonaws.athena.connectors.jdbc.connection.DatabaseConnectionConfig;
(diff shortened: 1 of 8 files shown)


athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
```diff
@@ -143,6 +147,14 @@ public GetDataSourceCapabilitiesResponse doGetDataSourceCapabilities(BlockAlloca
         ));
 
         jdbcQueryPassthrough.addQueryPassthroughCapabilityIfEnabled(capabilities, configOptions);
+
+        //Provide a hint to the engine that postgresql is using default collate settings
+        //Which doesn't match Athena's Engine Collation; this disabling Predicate pushdown
+        boolean usingDefaultCollate = Boolean.valueOf(this.configOptions.getOrDefault(USING_DEFAULT_COLLATE, "false"));
```
ejeffrli, line 153: Is the default boolean for "using_default_collate" supposed to be false? The naming seems confusing here. It looks like capabilities is using NON_DEFAULT_COLLATE setting when using_default_collate is true. Is this intended?

AbdulR3hman: Using default collate is supposed to be false by default so we don't effect existing customers; and customer who chose to use PostgreSQL default collation; required to set this to true in order to disable dynamic filtering.

ejeffrli: Can we make the name more explicit for it seems like using_default_collate = true is enabling a NON_DEFAULT_COLLATE. Maybe using_postgresql_default_collate? or can change NON_DEFAULT_COLLATE naming

AbdulR3hman: changed the name to be clearer;

ejeffrli (approved):

PR #2340, opened 2024-10-17
Opt in imageuri

aimethed:
*Description of changes:*
Because AWS accounts can only opt-in to certain regions, different accounts needed to be used for the 2 opt-in regions SAR is available. This change adds conditions and if statements to account for these different accounts.

Originally, I intended to use a cfn mapping with a default value. That transform is not supported by SAM though (see https://github.com/aws/aws-sam-cli/issues/4835), so I could not do that. Instead, I followed this stack overflow post: https://stackoverflow.com/questions/35904204/default-value-on-mappings-aws-cloudformation

Once we transition off of SAR/SAM we should look to use a mapping.

### DIFF
--- a/athena-aws-cmdb/athena-aws-cmdb.yaml
+++ b/athena-aws-cmdb/athena-aws-cmdb.yaml
@@ -42,6 +42,8 @@ Parameters:
     Type: String
 Conditions:
   HasPermissionsBoundary: !Not [ !Equals [ !Ref PermissionsBoundaryARN, "" ] ]
+  IsRegionBAH: !Equals [!Ref "AWS::Region", "me-south-1"]
+  IsRegionHKG: !Equals [!Ref "AWS::Region", "ap-east-1"]
 Resources:
   ConnectorConfig:
     Type: 'AWS::Serverless::Function'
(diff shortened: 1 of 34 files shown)


athena-aws-cmdb/athena-aws-cmdb.yaml
```diff
@@ -53,7 +55,9 @@ Resources:
           spill_prefix: !Ref SpillPrefix
       FunctionName: !Ref AthenaCatalogName
       PackageType: "Image"
-      ImageUri: !Sub '292517598671.dkr.ecr.${AWS::Region}.amazonaws.com/athena-federation-repository-aws-cmdb:2022.47.1'
+      ImageUri: !Sub
+        - '${Account}.dkr.ecr.${AWS::Region}.amazonaws.com/athena-federation-repository-aws-cmdb:2022.47.1'
+        - Account: !If [IsRegionBAH, 084828588479, !If [IsRegionHKG, 183295418215, 292517598671]]
```
mschoeni1, line 60: Why are there 2 regions for HKG?

mschoeni1 (approved):

ejeffrli (approved):

Issue #2325, opened 2024-10-10
[BUG] Observed no data (0 items) returned for float type in WHERE clause with connector ElasticSearch


abhishekpoddar-trianz:
**Describe the bug**
Observed no data (0 items) returned for float type in WHERE clause.

**To Reproduce**
- Run the below query from Athena and it returns no value.
select * from "lambda:elasticsearchtest"."opensearch-domain-test".datatypetestnew where float_col=45.12
- However, on running the below query returns the item with float_col - 45.12

**Expected behavior**
When using WHERE clause in the SELECT query, expectation is that it returns the item with that value.

abhishekpoddar-trianz:
@macohen - can you please assign this ticket to @VenkatasivareddyTR

abhishekpoddar-trianz:
As the code changes are done and PR merged, so closing this issue for now.

PR #2341, opened 2024-10-18
Bug/elasticsearch float issue

VenkatasivareddyTR:
*Issue #, if available:*
https://github.com/awslabs/aws-athena-query-federation/issues/2325 
*Description of changes:*
While fetching data from Elasticsearch by applying “where” condition for float type it’s not working. So, we are converting float to double to fetch data. Please find attached document for more details.
[ElasticSearchFloat_Issue.docx](https://github.com/user-attachments/files/17435459/ElasticSearchFloat_Issue.docx)


aimethed (approved):
Looking through the doc, it seems like this is just a bandaid fix for a larger underlying problem that there is some difference between the way we compare types. Because Athena does support floats and doubles separately. But this does solve the issue, so approving.

Issue #2295, opened 2024-09-27
[RFC] JDK 17 Upgrade & JDK LTS 


macohen:
**Is your feature request related to a problem? If yes, please describe.**
This project builds using JDK 8 and 11 today. Updating to JDK 17 would help modernize the connectors and create a clearer path to stay modern and update to the latest JDKs regularly. 

**Describe the solution you'd like**
* Support up to three JDKs - this helps for folks building custom connectors who may not be using a newer JDK yet.
* Keep current with the latest version AWS Lambda supports
* Plan for an EOL for JDK support 
* Continue to build JDK 11 and support JDK 8
* Create a new release that works for JDK 17
Does this seem like a decent approach? Other suggestions? Pitfalls?

aimethed:
Since we will still need to support JDK8, by my understanding we can't use features provided later than that? So we just need to make all the connectors compilable using JDK 17?

macohen:
I think that's right, @aimethed. Until we can actually retire JDK8, we only need to build in JDK17.

abhishekpoddar-trianz:
@macohen - could you please assign this ticket to me.

macohen:
@abhishekpoddar-trianz, please ask any questions for clarification here. I think the priority is producing a build with JDK17. If we need to JDK features so it can run on a JDK8 VM instead of producing multiple builds, that's OK.

VenkatasivareddyTR:
While building with JDK 17, build was failing with below error message:
java.lang.NoClassDefFoundError: Could not initialize class org.apache.arrow.memory.util.MemoryUtil
        at org.apache.arrow.memory.ArrowBuf.setZero(ArrowBuf.java:1175)
		
We have fixed build issue by adding java environment variable to yaml file (JAVA_TOOL_OPTIONS: "--add-opens=java.base/java.nio=ALL-UNNAMED") as per the Apache Arrow recommendation https://arrow.apache.org/docs/java/install.html.
 
and we have added <argLine>--add-opens=java.base/java.nio=ALL-UNNAMED</argLine> to maven-surefire-plugin and maven-failsafe-plugin in pom.xml file as per the Apache Arrow recommendation https://arrow.apache.org/docs/java/install.html#java-install-maven-testing.
 
After adding these changes, we are able to build successfully. Now we are working on functional testing.

macohen:
Fantastic. Adding @mschoeni1 to work on this with you as we move along.

macohen:
[...]

PR #2342, opened 2024-10-18
Upgrade to JDK17.

VenkatasivareddyTR:
*Issue #, if available:*
https://github.com/awslabs/aws-athena-query-federation/issues/2295 
*Description of changes:*
Upgrade to JDK 17.

### DIFF
--- a/athena-aws-cmdb/Dockerfile
+++ b/athena-aws-cmdb/Dockerfile
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/lambda/java:11
+FROM public.ecr.aws/lambda/java:17
 
 # Copy function code and runtime dependencies from Maven layout
 COPY target/athena-aws-cmdb-2022.47.1.jar ${LAMBDA_TASK_ROOT}
--- a/athena-aws-cmdb/Dockerfile
+++ b/athena-aws-cmdb/Dockerfile
@@ -1,4 +1,12 @@
-FROM public.ecr.aws/lambda/java:17
+# Argument for Java version, defaulting to 11
+ARG JAVA_VERSION=11
+# Use the specified version of Java
+FROM public.ecr.aws/lambda/java:${JAVA_VERSION}
+
+# Argument for Java tool options, defaulting to an empty string
+ARG JAVA_TOOL_OPTIONS=""
+# Set the JAVA_TOOL_OPTIONS environment variable for Java 17
+ENV JAVA_TOOL_OPTIONS=${JAVA_TOOL_OPTIONS}
 
 # Copy function code and runtime dependencies from Maven layout
 COPY target/athena-aws-cmdb-2022.47.1.jar ${LAMBDA_TASK_ROOT}
--- a/athena-aws-cmdb/athena-aws-cmdb.yaml
+++ b/athena-aws-cmdb/athena-aws-cmdb.yaml
@@ -40,10 +40,19 @@ Parameters:
     Description: "(Optional) An IAM policy ARN to use as the PermissionsBoundary for the created Lambda function's execution role"
     Default: ''
     Type: String
+  JavaVersion:
+    Description: "(Optional) Specify the Java version for the Lambda function. Default is '17'; use '11' for Java 11 or '8' for Java 8."
+    Type: String
+    AllowedValues:
+      - "17"
+      - "11"
+      - "8"
+    Default: "17"
 Conditions:
   HasPermissionsBoundary: !Not [ !Equals [ !Ref PermissionsBoundaryARN, "" ] ]
   IsRegionBAH: !Equals [!Ref "AWS::Region", "me-south-1"]
   IsRegionHKG: !Equals [!Ref "AWS::Region", "ap-east-1"]
+  IsJava17: !Equals [!Ref JavaVersion, "17"]
 Resources:
   ConnectorConfig:
     Type: 'AWS::Serverless::Function'
@@ -53,6 +62,7 @@ Resources:
           disable_spill_encryption: !Ref DisableSpillEncryption
           spill_bucket: !Ref SpillBucket
           spill_prefix: !Ref SpillPrefix
+          JAVA_TOOL_OPTIONS: !If [IsJava17, "--add-opens=java.base/java.nio=ALL-UNNAMED", !Ref "AWS::NoValue"]
       FunctionName: !Ref AthenaCatalogName
       PackageType: "Image"
       ImageUri: !Sub
--- a/athena-aws-cmdb/athena-aws-cmdb.yaml
+++ b/athena-aws-cmdb/athena-aws-cmdb.yaml
@@ -40,19 +40,10 @@ Parameters:
     Description: "(Optional) An IAM policy ARN to use as the PermissionsBoundary for the created Lambda function's execution role"
     Default: ''
     Type: String
-  JavaVersion:
-    Description: "(Optional) Specify the Java version for the Lambda function. Default is '17'; use '11' for Java 11 or '8' for Java 8."
-    Type: String
-    AllowedValues:
-      - "17"
-      - "11"
-      - "8"
-    Default: "17"
 Conditions:
   HasPermissionsBoundary: !Not [ !Equals [ !Ref PermissionsBoundaryARN, "" ] ]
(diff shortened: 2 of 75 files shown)


macohen:
Thanks @VenkatasivareddyTR; can you check on the build error, please: 
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (default-compile) on project aws-athena-federation-sdk: Fatal error compiling: error: release version 17 not supported -> [Help 1]
```

https://github.com/awslabs/aws-athena-query-federation/actions/runs/11475342726/job/31973995753?pr=2342

VenkatasivareddyTR:
@macohen The build process appears to be running on JDK 11. To align with the codebase, we need to update the build environment to use JDK 17.

mschoeni1 (approved):

macohen:
the code looks fine, but the GH actions are breaking and won't work until we're building with JDK 17 it seems. Is that right? are you able to update the actions (I know this build still won't pass until those are merged)...

abhishekpoddar-trianz:
Yes correct, we would need to update the GH actions with JDK 17 to build this PR successfully. We have tested this in our local environments with JDK 17 setup.
 
To update the GH actions to take JDK 17, we are updating the respective YAML files (in .github/workflows/). We will be testing it in a fork before raising the PR for the same.

abhishekpoddar-trianz:
PR for GitHub Actions update for java 17.
https://github.com/awslabs/aws-athena-query-federation/pull/2377

macohen:
@mschoeni1 can we merge this into a new branch and then merge master on top of that? I think we need to work on producing JDK 11 AND JDK 17 builds in GH actions before that gets merged back to master. What do you think? cc: @aimethed

mschoeni1:
Thanks @Jithendar12, this looks great. Could you please just also update the README to include the command for users to be able to build with JDK17?

AbdulR3hman:
few questions; 

1. regarding `JAVA_TOOL_OPTIONS` being used as `ARG` vs `ENV`;  then we proceed by setting the empty ARG to the ENV; but the comment says we are setting it to JDK 17; I think I might be missing something. I assume from this comment https://github.com/awslabs/aws-athena-query-federation/issues/2295#issuecomment-2422484927 it is coming from yaml; where exactly? 

2. we're adding JDK 17 profiles to `athena-federation-sdk/pom.xml` | but at the same time we are adding both JDK8 and 17 to the main `pom.xml`; why the discrepancy? 
3. Following up from that; if we added the profile in the main maven; profiles are not inherited according to [maven docs](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#Profile_Inheritance); this is why I assume we are using [activation](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#JDK) in order to enable the profile on children POMs? 
4. Same goes for the plugins; the parent pom file has `maven-surefire-plugin` introduced to it; why are we introducing the same code to all children? But the plugins are inherited by children POM; so why are we doing this?  
5. Lastly; where are you plan to utilize the jdk 17 profile? that was introduced?

Jithendar12:
> few questions;
> 
>     1. regarding `JAVA_TOOL_OPTIONS` being used as `ARG` vs `ENV`;  then we proceed by setting the empty ARG to the ENV; but the comment says we are setting it to JDK 17; I think I might be missing something. I assume from this comment [[RFC] JDK 17 Upgrade & JDK LTS  #2295 (comment)](https://github.com/awslabs/aws-athena-query-federation/issues/2295#issuecomment-2422484927) it is coming from yaml; where exactly?
> 
>     2. we're adding JDK 17 profiles to `athena-federation-sdk/pom.xml` | but at the same time we are adding both JDK8 and 17 to the main `pom.xml`; why the discrepancy?
> 
>     3. Following up from that; if we added the profile in the main maven; profiles are not inherited according to [maven docs](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#Profile_Inheritance); this is why I assume we are using [activation](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#JDK) in order to enable the profile on children POMs?
> 
>     4. Same goes for the plugins; the parent pom file has `maven-surefire-plugin` introduced to it; why are we introducing the same code to all children? But the plugins are inherited by children POM; so why are we doing this?
> 
>     5. Lastly; where are you plan to utilize the jdk 17 profile? that was introduced?

Hi @AbdulR3hman ,

1) We pass JAVA_TOOL_OPTIONS as a build argument (--build-arg) when building the Docker image. During the build process, the ARG is used to set the ENV variable in the image.
Example command:
docker build --build-arg **JAVA_VERSION=17** --build-arg **JAVA_TOOL_OPTIONS**="--add-opens=java.base/java.nio=ALL-UNNAMED" -t athena-federation-repository-snowflake:2022.47.1 . 

2) The JDK 8 profile in the main pom.xml ensures that if a user is building the code with JDK 8 locally, then maven.compiler.release is overridden to 8. This avoids compatibility issues and ensures the correct compiler version is used.

3) Yes, you are correct. The activation ensures that the profiles are automatically enabled based on the detected JDK version. When the JDK version matches the activation criteria (e.g., jdk=17), the properties and plugins defined in the profile are applied automatically, even in child POMs.

4) Not all child pom, only few where we have explicility added this [plugin](https://github.com/awslabs/aws-athena-query-federation/blob/2d1bea3ff59074224530292117d21e32a4e86d92/athena-hbase/pom.xml#L137) instead of picking up from the main pom. This forced me to add profile again in those pom's.

5) The JDK 17 profile is  used for setting maven.compiler.release to 17 and
adding **<argLine>--add-opens=java.base/java.nio=ALL-UNNAMED</argLine>**  option to the Surefire and Failsafe plugins. This configuration is required to ensure the unit and integration tests run successfully on JDK 17, as documented [here.](https://arrow.apache.org/docs/java/install.html#java-install-maven-testing)

Please refer to my [wiki page](https://github.com/Jithendar12/aws-athena-query-federation/wiki/How-To-Build-And-Deploy-A-Connector-Using-JDK-17) on how we are building and deploying the connector using JDK 17. Thank you!

AbdulR3hman:
>Not all child pom, only few where we have explicility added this [plugin](https://github.com/awslabs/aws-athena-query-federation/blob/2d1bea3ff59074224530292117d21e32a4e86d92/athena-hbase/pom.xml#L137) instead of picking up from the main pom. This forced me to add profile again in those pom's.

Can't we remove this plugin from all children and reference it from the parent pom?

AbdulR3hman:
I did further reading; I found this answer to a stackoverflow that we are attempting to replicate; 

https://stackoverflow.com/a/53666166
 
The link above is to the answer not the main post; and explains how we could achieve children all inherting the plugin without having to be repetitive and add profiles or anything of that sort in the children pom. I wouldn't mind having this change as a separate PR; but could you at least experiment with it?

AbdulR3hman:
last question I have; don't we need to make change to our github actions or is the intention to continue to use JDK11 by default? (question to the team as well as the PR's Author)

AbdulR3hman (approved):

Jithendar12:
> > Not all child pom, only few where we have explicility added this [plugin](https://github.com/awslabs/aws-athena-query-federation/blob/2d1bea3ff59074224530292117d21e32a4e86d92/athena-hbase/pom.xml#L137) instead of picking up from the main pom. This forced me to add profile again in those pom's.
> 
> Can't we remove this plugin from all children and reference it from the parent pom?

Hi @AbdulR3hman , this plugin has been removed from child pom's.

Jithendar12:
> I did further reading; I found this answer to a stackoverflow that we are attempting to replicate;
> 
> https://stackoverflow.com/a/53666166
> 
> The link above is to the answer not the main post; and explains how we could achieve children all inherting the plugin without having to be repetitive and add profiles or anything of that sort in the children pom. I wouldn't mind having this change as a separate PR; but could you at least experiment with it?

Thanks for the reference @AbdulR3hman. I have incorporated the changes in this PR only.

abhishekpoddar-trianz:
> last question I have; don't we need to make change to our github actions or is the intention to continue to use JDK11 by default? (question to the team as well as the PR's Author)

@abdul - in case it's decided to move to JDK17 as default then for GH actions PR https://github.com/awslabs/aws-athena-query-federation/pull/2377 can be referred.

PR #2377, opened 2024-11-04
Upgrade to Java 17: GitHub Actions workflows and prepare_dev_env.sh

Jithendar12:
*Description of changes:*
Updated Github Actions to java 17.
Updated prepare_dev_env.sh to java 17.

### DIFF
--- a/.github/workflows/cut_release.yml
+++ b/.github/workflows/cut_release.yml
@@ -13,11 +13,11 @@ jobs:
           sudo apt-get install -y libxml2-utils python3
           pip3 install lxml beautifulsoup4
 
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
           distribution: 'corretto'
-          java-version: '11'
+          java-version: '17'
 
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -62,22 +62,22 @@ jobs:
           git log --format='  - %s' $PREVIOUS_RELEASE_TAG..HEAD >> /tmp/RELEASE_MESSAGE
           git commit -a -F /tmp/RELEASE_MESSAGE
 
-      - name: Build Java 11 jars
+      - name: Build Java 17 jars
         run: |
           mvn clean package -T 1C -DskipTests -Dorg.slf4j.simpleLogger.defaultLogLevel=WARN --no-transfer-progress
 
           # Copy over the sdk jars
-          mkdir -p /tmp/java11_sdk_jars/
-          cp ./athena-federation-sdk/target/aws-athena-federation-sdk-${NEW_RELEASE_VERSION}.jar /tmp/java11_sdk_jars/
-          cp ./athena-federation-sdk/target/aws-athena-federation-sdk-${NEW_RELEASE_VERSION}-withdep.jar /tmp/java11_sdk_jars/
+          mkdir -p /tmp/java17_sdk_jars/
+          cp ./athena-federation-sdk/target/aws-athena-federation-sdk-${NEW_RELEASE_VERSION}.jar /tmp/java17_sdk_jars/
+          cp ./athena-federation-sdk/target/aws-athena-federation-sdk-${NEW_RELEASE_VERSION}-withdep.jar /tmp/java17_sdk_jars/
 
           # Copy over the connector jars
-          mkdir -p /tmp/java11_connector_jars/
-          find athena-*/target -name "*.jar" -type f | grep -v "test" | grep -v "arrow" | grep -v "/original" | grep -v "example" | grep -v "\-sdk-" | grep -v "\-dsv2/" | grep -v "athena-jdbc" | xargs -I{} cp {} /tmp/java11_connector_jars/
+          mkdir -p /tmp/java17_connector_jars/
+          find athena-*/target -name "*.jar" -type f | grep -v "test" | grep -v "arrow" | grep -v "/original" | grep -v "example" | grep -v "\-sdk-" | grep -v "\-dsv2/" | grep -v "athena-jdbc" | xargs -I{} cp {} /tmp/java17_connector_jars/
 
           # Copy over the connector zips
-          mkdir -p /tmp/java11_connector_zips/
-          find athena-*/target/ -name "*.zip" -type f | grep -v "test" | grep -v "arrow" | grep -v "/original" | grep -v "example" | grep -v "\-sdk-" | grep -v "\-dsv2/" | xargs -I{} cp {} /tmp/java11_connector_zips/
+          mkdir -p /tmp/java17_connector_zips/
+          find athena-*/target/ -name "*.zip" -type f | grep -v "test" | grep -v "arrow" | grep -v "/original" | grep -v "example" | grep -v "\-sdk-" | grep -v "\-dsv2/" | xargs -I{} cp {} /tmp/java17_connector_zips/
 
       - name: Push the release branch
         run: |
@@ -107,17 +107,17 @@ jobs:
           Athena Federation SDK jars
           |CheckSum|File|
           |----------|----|
-          $(cd /tmp/java11_sdk_jars && ls | xargs -I{} cksum {} | sed 's/\([0-9]\) \([a-zA-Z]\)/\1|\2/g' | sed 's/^/|/g' | sed 's/$/|/g')
+          $(cd /tmp/java17_sdk_jars && ls | xargs -I{} cksum {} | sed 's/\([0-9]\) \([a-zA-Z]\)/\1|\2/g' | sed 's/^/|/g' | sed 's/$/|/g')
 
           Athena Federation Connector Lambda jars
           |CheckSum|File|
           |----------|----|
-          $(cd /tmp/java11_connector_jars && ls | xargs -I{} cksum {} | sed 's/\([0-9]\) \([a-zA-Z]\)/\1|\2/g' | sed 's/^/|/g' | sed 's/$/|/g')
+          $(cd /tmp/java17_connector_jars && ls | xargs -I{} cksum {} | sed 's/\([0-9]\) \([a-zA-Z]\)/\1|\2/g' | sed 's/^/|/g' | sed 's/$/|/g')
 
           Athena Federation Connector Lambda zips
           |CheckSum|File|
           |----------|----|
-          $(cd /tmp/java11_connector_zips && ls | xargs -I{} cksum {} | sed 's/\([0-9]\) \([a-zA-Z]\)/\1|\2/g' | sed 's/^/|/g' | sed 's/$/|/g')
+          $(cd /tmp/java17_connector_zips && ls | xargs -I{} cksum {} | sed 's/\([0-9]\) \([a-zA-Z]\)/\1|\2/g' | sed 's/^/|/g' | sed 's/$/|/g')
 
           ## What's Changed
           $(git log --format='- %s' v$PREVIOUS_RELEASE_VERSION..HEAD~1)
@@ -133,7 +133,7 @@ jobs:
             --notes-file /tmp/RELEASE_NOTES
           # Upload the jar attachments
           gh release upload "v$NEW_RELEASE_VERSION" \
-            $(find /tmp/java11_sdk_jars -type f) \
-            $(find /tmp/java11_connector_jars -type f) \
-            $(find /tmp/java11_connector_zips -type f) \
+            $(find /tmp/java17_sdk_jars -type f) \
+            $(find /tmp/java17_connector_jars -type f) \
+            $(find /tmp/java17_connector_zips -type f) \
             --clobber
--- a/.github/workflows/javadoc_sync.yaml
+++ b/.github/workflows/javadoc_sync.yaml
@@ -10,11 +10,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
           distribution: 'corretto'
-          java-version: '11'
+          java-version: '17'
 
       - name: Build Javadoc with Maven
         run: mvn -B javadoc:aggregate --file pom.xml -Dorg.slf4j.simpleLogger.defaultLogLevel=WARN --no-transfer-progress
--- a/.github/workflows/maven_push.yml
+++ b/.github/workflows/maven_push.yml
@@ -22,11 +22,11 @@ jobs:
     runs-on: aws-athena-query-federation_ubuntu-latest_16-core
     steps:
       - uses: actions/checkout@v4
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
           distribution: 'corretto'
-          java-version: '11'
+          java-version: '17'
       # Target java 8 to ensure that the source is compatible with java 8
       - name: Build with Maven
         env:
--- a/.github/workflows/publish_to_maven_central.yml
+++ b/.github/workflows/publish_to_maven_central.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Maven Central Repository
         uses: actions/setup-java@v4
         with:
-          java-version: '11'
+          java-version: '17'
           distribution: 'adopt'
           server-id: ossrh
           server-username: MAVEN_USERNAME
--- a/.github/workflows/run_release_tests.yml
+++ b/.github/workflows/run_release_tests.yml
@@ -21,11 +21,11 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 18
-      - name: Set up JDK 11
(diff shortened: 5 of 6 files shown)


.github/workflows/cut_release.yml
```diff
@@ -13,11 +13,11 @@ jobs:
           sudo apt-get install -y libxml2-utils python3
           pip3 install lxml beautifulsoup4
 
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
           distribution: 'corretto'
-          java-version: '11'
```
macohen, line 21: Are we able to support multiple versions using the matrix strategy? https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations

I'd also want to see how this would impact our release process if we were producing multiple JDK version releases. cc: @AbdulR3hman @ejeffrli @aimethed

Jithendar12: Hi @macohen ,
We have updated the GitHub workflow to use a matrix strategy for Java 11 and 17. A few unit tests were failing under Java 17, but these issues have been resolved, and the build now succeeds on both Java versions.
Attached is the GitHub Actions run from my fork. We are continuing thorough testing and will update the PR once it is completed. Thank you!

image

.github/workflows/cut_release.yml
```diff
@@ -51,46 +56,134 @@ jobs:
...
+        with:
+          name: java${{ matrix.java-version }}-artifacts
+          path: |
+            /tmp/java${{ matrix.java-version }}_sdk_jars/
+            /tmp/java${{ matrix.java-version }}_connector_jars/
+            /tmp/java${{ matrix.java-version }}_connector_zips/
 
-      - name: Push the release branch
+      - name: Wait for Java 11 artifacts
+        if: ${{ matrix.java-version == '17' }}
```
christov: Could you please elaborate on the need to wait for Java 11 artifacts when the Java version is 17?

.github/workflows/publish_to_maven_central.yml
```diff
@@ -26,6 +29,7 @@ jobs:
           gpg-passphrase: MAVEN_GPG_PASSPHRASE
 
       - name: Publish to Apache Maven Central
+        if: ${{ matrix.java-version == '17' }}
```
christov: Am I interpreting this correctly that after this change takes effect, only a release compiled with Java 17 will be published to Maven Central?

.github/workflows/cut_release.yml
Jithendar12: Hi @christov, we are waiting for the Java 11 artifacts in the Java 17 job because the final release (including GitHub release creation and artifact uploads) is coordinated by the Java 17 workflow.
The Java 17 job bundles both Java 11 and Java 17 builds into the release, so it needs to ensure that the Java 11 artifacts are available before proceeding.

.github/workflows/publish_to_maven_central.yml
Jithendar12: Yes, that’s correct — only the Java 17 job publishes artifacts to Maven Central.
Both Java 11 and Java 17 builds are executed for compatibility verification, but the actual deployment is restricted to the Java 17 build to ensure a single, consistent release artifact.

Jithendar12: Hi @christov , updated the workflow to publish both JDK 11 and JDK 17 artifacts to Maven Central.
Each artifact will have a version suffix, for example:
2025.43.1.jdk11 and 2025.43.1.jdk17.
Let me know if this approach is fine or if we should revert to only publishing JDK 11 artifacts. Thank you!

Jithendar12:
Hi @christov ,
We’ve updated this PR to follow the new design: all workflows now build with both Java 11 and Java 17, and only the Java 11 jobs publish artifacts. 
Please review and let us know if you have any feedback. Thank you!

.github/workflows/cut_release.yml
```diff
@@ -40,9 +44,11 @@ jobs:
           echo "NEW_RELEASE_VERSION=$NEW_RELEASE_VERSION" >> $GITHUB_ENV
 
       - name: Create branch locally on the runner
+        if: ${{ matrix.java-version == '11' }}
         run: git checkout -b release_${NEW_RELEASE_VERSION} origin/master
 
       - name: Bump the versions in the branch
+        if: ${{ matrix.java-version == '11' }}
```
ejeffrli: Since Java versions are only being bumped for 11 with this statement, are we sure when java 17 builds the JAR's it is using java 17 to build? Seems like it would just use the default JDK defined in pom.xml

ejeffrli: seems like 11 here. https://github.com/awslabs/aws-athena-query-federation/blob/a8717a4b571d0bf0836e7c8460ad340a2910a614/pom.xml#L14

ejeffrli: does this profile get triggered for java 17 case? Any relation with github actions? What is this profile's use case? https://github.com/awslabs/aws-athena-query-federation/blob/a8717a4b571d0bf0836e7c8460ad340a2910a614/pom.xml#L523

.github/workflows/cut_release.yml
```diff
@@ -81,16 +91,19 @@ jobs:
           find athena-*/target/ -name "*.zip" -type f | grep -v "test" | grep -v "arrow" | grep -v "/original" | grep -v "example" | grep -v "\-sdk-" | grep -v "\-dsv2/" | xargs -I{} cp {} /tmp/java11_connector_zips/
 
       - name: Push the release branch
+        if: ${{ matrix.java-version == '11' }}
         run: |
-          git push origin release_${NEW_RELEASE_VERSION}:release_${NEW_RELEASE_VERSION}
+          git push --force origin release_${NEW_RELEASE_VERSION}:release_${NEW_RELEASE_VERSION}
```
ejeffrli: Why is force necessary here? Just curious

.github/workflows/cut_release.yml
ejeffrli: Let's verify java 17 is being used in the github action run as well. Maybe a screenshot or link to github action

.github/workflows/javadoc_sync.yaml
```diff
@@ -6,31 +6,38 @@ on:
...
-      - name: Set up JDK 11
+      - name: Set up JDK ${{ matrix.java-version }}
         uses: actions/setup-java@v5
         with:
           distribution: 'corretto'
-          java-version: '11'
+          java-version: ${{ matrix.java-version }}
 
       - name: Build Javadoc with Maven
         run: mvn -B javadoc:aggregate --file pom.xml -Dorg.slf4j.simpleLogger.defaultLogLevel=WARN --no-transfer-progress
```
ejeffrli, line 20: Do we know if java 17 is being explicitly used here? We may have to pass in a flag or check what maven is using

.github/workflows/maven_push.yml
```diff
@@ -20,19 +20,21 @@ jobs:
...
           distribution: 'corretto'
-          java-version: '11'
-      # Target java 8 to ensure that the source is compatible with java 8
+          java-version: ${{ matrix.java-version }}
       - name: Build with Maven
         env:
           AWS_DEFAULT_REGION: us-east-1
           AWS_REGION: us-east-1
-        run: mvn -B clean package -T 1C --file pom.xml -Dmaven.compiler.release=11 -Dorg.slf4j.simpleLogger.defaultLogLevel=WARN --no-transfer-progress
+        run: mvn -B clean package -T 1C --file pom.xml -Dorg.slf4j.simpleLogger.defaultLogLevel=WARN --no-transfer-progress
```
ejeffrli, line 37: Let's make sure it's 17 here as well

.github/workflows/publish_to_maven_central.yml
```diff
@@ -9,12 +9,25 @@ jobs:
...
         uses: actions/checkout@v6
 
+      - name: Set up JDK ${{ matrix.java-version }}
+        uses: actions/setup-java@v5
+        with:
+          distribution: 'corretto'
+          java-version: ${{ matrix.java-version }}
+
+      - name: Build Java ${{ matrix.java-version }}
+        if: ${{ matrix.java-version == '17' }}
```
ejeffrli: Is this necessary to keep? This part seems to not to be too important with the previous steps essentially building the whole repo with the tests already. But, I understand if it is necessary to satisfy some sort of github action requirement I may not be aware of.

.github/workflows/run_release_tests.yml
```diff
@@ -40,7 +45,9 @@ jobs:
           aws-region: us-east-1
           role-duration-seconds: 18000
           role-to-assume: ${{ secrets.AWS_ASSUME_ROLE }}
+      # Run release tests ONLY on Java 11
       - name: Run tests
+        if: ${{ matrix.java-version == '11' }}
```
ejeffrli: Any reason why we don't run release tests on java 17 as well?

.github/workflows/cut_release.yml
Jithendar12: Hi @ejeffrli , thanks for reviewing.
The java17 profile is automatically activated whenever the build runs under JDK 17. It uses:
```
<activation>
<jdk>17</jdk>
</activation>
```
 
So, when actions/setup-java@v5 installs Java 17 (via matrix.java-version: '17'), Maven detects that JDK and automatically activates the corresponding profile. This sets maven.compiler.release=17, so no manual activation is required.
 
Additionally, I have added the exec-maven-plugin in the latest commit to log the active profile, Maven compiler release, and Java versions for verification.
 
Please find the attached screenshots that verify the Java version for the cut-release workflow (this warning message now appears in all other workflows as well), along with the link to the cut-release workflow in my fork.
Java 11:
image

Java 17:
image

link to cut-release:
[Cut release](https://github.com/Jithendar12/aws-athena-query-federation/actions/runs/19962713984)

.github/workflows/cut_release.yml
Jithendar12: If the workflow is re-run or if a previous failed run already created the branch, a normal push would fail due to a non-fast-forward update.
**--force** ensures the workflow can be safely re-run and the branch is always updated.

.github/workflows/javadoc_sync.yaml
Jithendar12: Yes. When the matrix value is 17, the workflow step:
 ```
 name: Set up JDK ${{ matrix.java-version }}
  uses: actions/setup-java@v5
```
installs JDK 17, so Maven runs under Java 17.
The java17 Maven profile then auto-activates and sets maven.compiler.release=17.

File: .github/workflows/maven_push.yml (hunk 7 above)
Comment (Jithendar12) [reply .github/workflows/maven_push.yml:37]: Yes, it uses Java 17 when matrix.java-version is set to '17'.
Additionally added the exec-maven-plugin in the latest commit to log the active profile, Maven compiler release, and Java versions for verification.

File: .github/workflows/publish_to_maven_central.yml (hunk 8 above)
Comment (Jithendar12) [reply .github/workflows/publish_to_maven_central.yml]: The Java 17 build is there to validate that the project compiles under Java 17 before publishing the Java 11 artifacts, which ensures forward-compatibility even when the workflow is triggered independently.
 
However, since **maven_push.yml** already builds both Java 11 and Java 17 on every PR/push, this step is technically redundant.
I can remove the Java 17 build from **publish_to_maven_central.yml** if preferred. Please let us know.

File: .github/workflows/run_release_tests.yml (hunk 9 above)
Comment (Jithendar12) [reply .github/workflows/run_release_tests.yml]: We currently run release tests only on Java 11 because those are the artifacts being published. Running the full release test suite on Java 17 would add an additional 2–3 hours to the daily pipeline. It may also lead to resource conflicts (creation/deletion of databases, Glue jobs, Athena data sources, etc.) when both jobs run in parallel.
 
If needed, we can enable release tests for Java 17 as well, but the Java 17 job would need to run sequentially to avoid these conflicts. Please let us know your preference.

File: .github/workflows/publish_to_maven_central.yml (hunk 8 above)
Comment (ejeffrli) [reply .github/workflows/publish_to_maven_central.yml]: ah i see. i think its fine to leave it

File: .github/workflows/run_release_tests.yml (hunk 9 above)
Comment (ejeffrli) [reply .github/workflows/run_release_tests.yml]: i see thanks. Let's leave it as just 11 for now

File: pom.xml [hunk 10]
```diff
@@ -507,6 +507,24 @@
...
+                <executions>
+                    <execution>
+                        <id>print-java-info</id>
+                        <phase>validate</phase>
+                        <goals>
+                            <goal>exec</goal>
+                        </goals>
+                        <configuration>
+                            <executable>bash</executable>
+                            <commandlineArgs>-c "echo '[WARNING] ========================================'; echo '[WARNING] Maven Build Configuration Verification'; echo '[WARNING] ========================================'; if [ '${maven.compiler.release}' = '8' ]; then echo '[WARNING] Activated Profile: java8'; elif [ '${maven.compiler.release}' = '17' ]; then echo '[WARNING] Activated Profile: java17'; else echo '[WARNING] Activated Profile: default (Java 11)'; fi; echo '[WARNING] maven.compiler.release = ${maven.compiler.release}'; echo '[WARNING] Java version = ${java.version}'; echo '[WARNING] ========================================'"</commandlineArgs>
```
christov: Are these warnings? Wondering whether something like `[INFO]` would be more appropriate because I believe these are informational messages rather than warnings. But, perhaps, there is some convention/other reasoning I an not aware of.

.github/workflows/cut_release.yml
```diff
@@ -7,18 +7,22 @@ jobs:
   cut_release:
     name: Cut a release
     runs-on: aws-athena-query-federation_ubuntu-latest_16-core
+    strategy:
+      matrix:
+        java-version: ['11', '17']
```
christov: This action is set up to run with 11 and 17, but then  if-statements are added to run the key steps with only 11.

If only the Java 11 artifacts will be released, I am wondering whether a matrix strategy should be used in the `cut_release` action. Perhaps, that action should remain specific to Java 11. I see that this is sort of a legacy from when this PR was aiming to release both Java 11 and 17 and I can see how keeping the matrix here is kind of a preparation when Java 17 releases start. But until then, it is probably a "premature optimization" plus it will be fairly easy to add the matrix back in the future. Is there any other strong reason for using a matrix strategy in this action?

Same comment for the `javadoc_sync` and `publish_to_maven_central` actions. I believe that `maven_push` would ensure that the code builds with 11 and 17 on each PR. Does adding 17 builds to any of the other actions add any value at this point?

Also, wondering whether a matrix should be used in the `codeql-analysis` action. I believe the kinds of vulnerabilities codeql checks for are independent of the Java version, but perhaps some are not?

pom.xml
Jithendar12: Hi @christov ,
These messages are meant purely for verification, not actual warnings. We used [WARNING] mainly to make the section visually stand out in long workflow logs, but functionally they are informational. We’re happy to switch them to [INFO] if we want to keep the log semantics consistent. Please let us know your preference.

.github/workflows/cut_release.yml
Jithendar12: Hi @christov , thanks for the review.
 
The Java 17 builds in the other workflows were originally included only to validate forward compatibility and to prepare the pipeline for future Java 17 publishing. This follows the team’s direction:
 
1. Keep the Java 11 build and publishing flow unchanged.
 
2. Build Java 17 in GitHub as part of the project build, but do not publish Java 17 artifacts.
 
Because of that, Java 17 was added only for build/validation, while all publishing continues to use Java 11. This also provides a forward-compatibility check even when the workflow is triggered independently.
 
**Regarding the matrix strategy:**
CodeQL: We kept the Java 11 + Java 17 matrix since the workflow builds the project before scanning, but CodeQL itself is Java-version-agnostic, so the matrix does not add meaningful value. We will revert this file.
 
Other workflows (cut_release, javadoc_sync, publish_to_maven_central, run_release_tests): Since maven_push.yml already validates both Java 11 and Java 17 on every PR, running Java 17 in those additional workflows is redundant. We can revert them to Java 11-only to simplify the pipeline.
 
Please let us know if you'd like us to proceed with removing the Java 17 matrix from these workflows and reverting them back, except for the maven_push.yml file.

christov: @Jithendar12, thank you for the explanation. Yes, please remove the matrix from the rest of the workflows except for the `maven_push` one.

pom.xml
christov: Let's use `[INFO]` given the nature of these log messages.

.github/workflows/cut_release.yml
Jithendar12: Hi @christov , removed. Thank you!

pom.xml
Jithendar12: Hi @christov , changed from [WARNING] to [INFO]. Thank you!

christov (approved):

PR #2388, opened 2024-11-14
Check if auth is not null

ejeffrli:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/connection/EnvironmentProperties.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/connection/EnvironmentProperties.java
@@ -95,7 +95,7 @@ public class EnvironmentProperties
     {
         Map<String, String> authMap = new HashMap<>();
 
-        if (StringUtils.isNotBlank(auth.secretArn())) {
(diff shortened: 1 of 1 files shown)


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/connection/EnvironmentProperties.java
```diff
@@ -95,7 +95,7 @@ private Map<String, String> authenticationConfigurationToMap(AuthenticationConfi
     {
         Map<String, String> authMap = new HashMap<>();
 
-        if (StringUtils.isNotBlank(auth.secretArn())) {
+        if (auth != null && StringUtils.isNotBlank(auth.secretArn())) {
```
macohen, line 98: With the previous change, auth being null throws an exception. Creating a new code path here, what happens if authMap is empty for calling methods?

aimethed (approved):

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/connection/EnvironmentProperties.java
ejeffrli: no tests covering. empty was expected before but it was just reaching this null exception before. empty is handled properly on line 57

PR #2417, opened 2024-11-22
use package for oracle and docdb

aimethed:
*Issue #, if available:*

*Description of changes:*
There was a regression with SSL capabilities for DocDB and Oracle connectors. This is likely because lambda controls the environment for package based lambdas (it has to set up an environment to run the zip in), while image based ones you provide the environment. The one lambda sets up for package based lambdas has the certs it seems already baked in. The ones we make with ECR do not (yet).

This returns DocDB and Oracle connectors to use package based lambdas


aimethed:
fixed

PR #2452, opened 2024-12-02
Add port  number as a DBS_PORT parameter in Teradata JDBC Connection String

Jithendar12:
Issue #, if available:
The Teradata JDBC connection string requires a specific format to include the port number.
Example URL:
jdbc:teradata://host/TMODE=ANSI,CHARSET=UTF8,DATABASE=TEST,**DBS_PORT=1025**,user=dummy,password=dummy

Previously, the port was included in the connection string as host:port, which was not compatible with the Teradata JDBC driver

Description of changes:
Overridden the connectionPropertiesToEnvironment method to include the port number as a DBS_PORT parameter.

### DIFF
--- a/athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataEnvironmentProperties.java
+++ b/athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataEnvironmentProperties.java
@@ -35,12 +35,10 @@ public class TeradataEnvironmentProperties extends JdbcEnvironmentProperties
     public Map<String, String> connectionPropertiesToEnvironment(Map<String, String> connectionProperties)
     {
         HashMap<String, String> environment = new HashMap<>();
-        // Default port for teradata is 1025
-        String port = connectionProperties.getOrDefault(PORT, "1025");
 
         // Construct the JDBC connection string and include the port as a DBS_PORT parameter
         String connectionString = getConnectionStringPrefix(connectionProperties) + connectionProperties.get(HOST)
-                + getDatabase(connectionProperties)  + ",DBS_PORT=" + port + getJdbcParameters(connectionProperties);
+                + getDatabase(connectionProperties)  + ",DBS_PORT=" + connectionProperties.getOrDefault(PORT, String.valueOf(TeradataConstants.TERADATA_DEFAULT_PORT)) + getJdbcParameters(connectionProperties);
 
         environment.put(DEFAULT, connectionString);
         return environment;
--- a/athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataEnvironmentProperties.java
+++ b/athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataEnvironmentProperties.java
@@ -21,12 +21,31 @@ package com.amazonaws.athena.connectors.teradata;
 
 import com.amazonaws.athena.connectors.jdbc.JdbcEnvironmentProperties;
 
+import java.util.HashMap;
 import java.util.Map;
 
 import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.DATABASE;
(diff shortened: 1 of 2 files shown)


athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataEnvironmentProperties.java
```diff
@@ -21,12 +21,31 @@
...
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.PORT;
 
 public class TeradataEnvironmentProperties extends JdbcEnvironmentProperties
 {
+    @Override
+    public Map<String, String> connectionPropertiesToEnvironment(Map<String, String> connectionProperties)
+    {
+        HashMap<String, String> environment = new HashMap<>();
+        // Default port for teradata is 1025
+        String port = connectionProperties.getOrDefault(PORT, "1025");
```
AbdulR3hman, line 39: Please move this to the Teradata constants; not here.

athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataEnvironmentProperties.java
```diff
@@ -21,12 +21,31 @@
...
+
+        // Construct the JDBC connection string and include the port as a DBS_PORT parameter
+        String connectionString = getConnectionStringPrefix(connectionProperties) + connectionProperties.get(HOST)
+                + getDatabase(connectionProperties)  + ",DBS_PORT=" + port + getJdbcParameters(connectionProperties);
```
AbdulR3hman, line 43: this is assuming a new change; will this break existing customers? what exactly promoted this change? I want to make sure this is backward compatible; i.e.; existing customers who are upgrading won't have their environments broken.

athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataEnvironmentProperties.java
```diff
@@ -21,12 +21,31 @@
...
 
 import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.DATABASE;
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.DEFAULT;
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.HOST;
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.PORT;
 
 public class TeradataEnvironmentProperties extends JdbcEnvironmentProperties
 {
+    @Override
+    public Map<String, String> connectionPropertiesToEnvironment(Map<String, String> connectionProperties)
```
AbdulR3hman, line 35: let's add a unit test once questions below are answered.

athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataEnvironmentProperties.java
Jithendar12: Removed.

athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataEnvironmentProperties.java
Jithendar12: Added. Thank you!

athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataEnvironmentProperties.java
Jithendar12: Hi @AbdulR3hman ,
No, this will not break existing customers because they will pass the DBS_PORT parameter(if teradata is running on non default port) as part of their connection string.

This issue was observed during Panama Phase 1 where we are explicitly constructing the connection string by taking the parameters from the Glue Connection, so this change is necessary.

PR #2487, opened 2024-12-20
Oracle case insensitive search

aimethed:
*Description of changes:*
Oracle by default upper cases everything. Glue does not support anything aside from lowercase. Originally, we were casing everything to upper case to match oracle. Now to support glue connections, we need to not do this. So we have three casing modes now, upper (default for legacy connectors), lower (default for connectors using glue connections), and now search, which will perform a case insensitive search.

### DIFF
--- a/athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
+++ b/athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
@@ -25,10 +25,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.Arrays;
-import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
@@ -42,11 +39,18 @@ import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConsta
 public class OracleCaseResolver
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(OracleCaseResolver.class);
+    private static final String SCHEMA_NAME_QUERY_TEMPLATE = "select distinct OWNER from all_tables where lower(OWNER) = ?";
+    private static final String TABLE_NAME_QUERY_TEMPLATE = "select distinct TABLE_NAME from all_tables where OWNER = ? and lower(TABLE_NAME) = ?";
+    private static final String SCHEMA_NAME_COLUMN_KEY = "OWNER";
+    private static final String TABLE_NAME_COLUMN_KEY = "TABLE_NAME";
 
     // the environment variable that can be set to specify which casing mode to use
     static final String CASING_MODE = "casing_mode";
 
-    private static final String ORACLE_QUOTE_CHARACTER = "\"";
+    // used for identifying database objects (ex: table names)
+    private static final String ORACLE_IDENTIFIER_CHARACTER = "\"";
+    // used in SQL statements for character strings (ex: where OWNER = 'example')
+    private static final String ORACLE_STRING_LITERAL_CHARACTER = "\'";
 
     private OracleCaseResolver() {}
 
@@ -101,58 +105,58 @@ public class OracleCaseResolver
     public static String getSchemaNameCaseInsensitively(final Connection connection, String schemaNameInput, Map<String, String> configOptions)
             throws SQLException
     {
-        String nameFromSnowFlake = null;
+        String nameFromOracle = null;
         int i = 0;
         try (PreparedStatement preparedStatement = new PreparedStatementBuilder()
                 .withConnection(connection)
                 .withQuery(SCHEMA_NAME_QUERY_TEMPLATE)
-                .withParameters(Arrays.asList(schemaNameInput.toLowerCase())).build();
+                .withParameters(Arrays.asList(convertToLiteral(schemaNameInput.toLowerCase()))).build();
                 ResultSet resultSet = preparedStatement.executeQuery()) {
             while (resultSet.next()) {
                 i++;
                 String schemaNameCandidate = resultSet.getString(SCHEMA_NAME_COLUMN_KEY);
                 LOGGER.debug("Case insensitive search on columLabel: {}, schema name: {}", SCHEMA_NAME_COLUMN_KEY, schemaNameCandidate);
-                nameFromSnowFlake = schemaNameCandidate;
+                nameFromOracle = schemaNameCandidate;
             }
         }
         catch (SQLException e) {
             throw new RuntimeException(e);
         }
 
-        if (i == 0 || i > 1) {
+        if (i != 1) {
             throw new RuntimeException(String.format("Schema name case insensitive match failed, number of match : %d", i));
         }
 
-        return nameFromSnowFlake;
+        return nameFromOracle;
     }
 
     public static String getTableNameCaseInsensitively(final Connection connection, String schemaName, String tableNameInput, Map<String, String> configOptions)
             throws SQLException
     {
         // schema name input should be correct case before searching tableName already
-        String nameFromSnowFlake = null;
+        String nameFromOracle = null;
         int i = 0;
         try (PreparedStatement preparedStatement = new PreparedStatementBuilder()
                 .withConnection(connection)
                 .withQuery(TABLE_NAME_QUERY_TEMPLATE)
-                .withParameters(Arrays.asList(schemaName, tableNameInput.toLowerCase())).build();
+                .withParameters(Arrays.asList(convertToLiteral(schemaName), convertToLiteral(tableNameInput.toLowerCase()))).build();
                 ResultSet resultSet = preparedStatement.executeQuery()) {
             while (resultSet.next()) {
                 i++;
                 String schemaNameCandidate = resultSet.getString(TABLE_NAME_COLUMN_KEY);
                 LOGGER.debug("Case insensitive search on columLabel: {}, schema name: {}", TABLE_NAME_COLUMN_KEY, schemaNameCandidate);
-                nameFromSnowFlake = schemaNameCandidate;
+                nameFromOracle = schemaNameCandidate;
             }
         }
         catch (SQLException e) {
             throw new RuntimeException(e);
         }
 
-        if (i == 0 || i > 1) {
+        if (i != 1) {
             throw new RuntimeException(String.format("Schema name case insensitive match failed, number of match : %d", i));
         }
 
-        return nameFromSnowFlake;
+        return nameFromOracle;
     }
 
     private static OracleCasingMode getCasingMode(Map<String, String> configOptions)
@@ -173,5 +177,24 @@ public class OracleCaseResolver
             LOGGER.error("Invalid input for:{}, input value:{}, valid values:{}", CASING_MODE, configOptions.get(CASING_MODE), Arrays.asList(OracleCasingMode.values()), ex);
             throw ex;
         }
+    }
+
+    public static TableName quoteTableName(TableName inputTable) {
+        String schemaName = inputTable.getSchemaName();
+        String tableName = inputTable.getTableName();
+        if (!schemaName.contains(ORACLE_IDENTIFIER_CHARACTER)) {
+            schemaName = String.join(ORACLE_IDENTIFIER_CHARACTER, schemaName, ORACLE_IDENTIFIER_CHARACTER);
+        }
+        if (!tableName.contains(ORACLE_IDENTIFIER_CHARACTER)) {
+            tableName = String.join(ORACLE_IDENTIFIER_CHARACTER, tableName, ORACLE_IDENTIFIER_CHARACTER);
+        }
+        return new TableName(schemaName, tableName);
+    }
+
+    public static String convertToLiteral(String input) {
+        if (!input.contains(ORACLE_STRING_LITERAL_CHARACTER)) {
+            input = String.join(ORACLE_STRING_LITERAL_CHARACTER, input, ORACLE_STRING_LITERAL_CHARACTER);
+        }
+        return input;
     }
 }
--- a/athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
+++ b/athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
@@ -57,7 +57,7 @@ public class OracleCaseResolver
     {
         LOWER,      // casing mode to lower case everything (glue and trino lower case everything)
         UPPER,      // casing mode to upper case everything (oracle by default upper cases everything)
-        SEARCH      // casing mode to perform case insensitive search
+        CASE_INSENSITIVE_SEARCH     // casing mode to perform case insensitive search
     }
 
     public static TableName getAdjustedTableObjectName(final Connection connection, TableName tableName, Map<String, String> configOptions)
@@ -65,7 +65,7 @@ public class OracleCaseResolver
     {
         OracleCasingMode casingMode = getCasingMode(configOptions);
         switch (casingMode) {
-            case SEARCH:
+            case CASE_INSENSITIVE_SEARCH:
                 String schemaNameCaseInsensitively = getSchemaNameCaseInsensitively(connection, tableName.getSchemaName());
                 String tableNameCaseInsensitively = getTableNameCaseInsensitively(connection, schemaNameCaseInsensitively, tableName.getTableName());
                 TableName tableNameResult = new TableName(schemaNameCaseInsensitively, tableNameCaseInsensitively);
@@ -89,7 +89,7 @@ public class OracleCaseResolver
     {
         OracleCasingMode casingMode = getCasingMode(configOptions);
         switch (casingMode) {
-            case SEARCH:
+            case CASE_INSENSITIVE_SEARCH:
                 LOGGER.info("casing mode is SEARCH: performing case insensitive search for Schema...");
                 return getSchemaNameCaseInsensitively(connection, schemaNameInput);
             case UPPER:
--- a/athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
+++ b/athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
@@ -38,8 +38,8 @@ import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConsta
 public class OracleCaseResolver
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(OracleCaseResolver.class);
-    static final String SCHEMA_NAME_QUERY_TEMPLATE = "select distinct OWNER from all_tables where lower(OWNER) = ?";
-    static final String TABLE_NAME_QUERY_TEMPLATE = "select distinct TABLE_NAME from all_tables where OWNER = ? and lower(TABLE_NAME) = ?";
+    static final String SCHEMA_NAME_QUERY_TEMPLATE = "SELECT DISTINCT OWNER as \"OWNER\" FROM all_tables where lower(OWNER) = ?";
+    static final String TABLE_NAME_QUERY_TEMPLATE = "select distinct TABLE_NAME as \"TABLE_NAME\" from all_tables where OWNER = ? and lower(TABLE_NAME) = ?";
     static final String SCHEMA_NAME_COLUMN_KEY = "OWNER";
     static final String TABLE_NAME_COLUMN_KEY = "TABLE_NAME";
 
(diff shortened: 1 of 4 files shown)


athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.Arrays;
+import java.util.Map;
+
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.DEFAULT_GLUE_CONNECTION;
+
+public class OracleCaseResolver
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(OracleCaseResolver.class);
+    static final String SCHEMA_NAME_QUERY_TEMPLATE = "SELECT DISTINCT OWNER as \"OWNER\" FROM all_tables where lower(OWNER) = ?";
+    static final String TABLE_NAME_QUERY_TEMPLATE = "select distinct TABLE_NAME as \"TABLE_NAME\" from all_tables where OWNER = ? and lower(TABLE_NAME) = ?";
```
AbdulR3hman, line 42: nit: can we keep the queries the same; if using capitals for the first query let's do that for the second; easier to read.

AbdulR3hman: I'd rather see all SQL in all caps.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,201 @@
...
+    private OracleCaseResolver() {}
+
+    private enum OracleCasingMode
+    {
+        LOWER,      // casing mode to use whatever the engine returns (in trino's case, lower)
+        UPPER,      // casing mode to upper case everything (oracle by default upper cases everything)
+        SEARCH      // casing mode to perform case insensitive search
+    }
+
+    public static TableName getAdjustedTableObjectName(final Connection connection, TableName tableName, Map<String, String> configOptions)
```
AbdulR3hman, line 63: are we assuming that if the table/schema is upper case then the columns are also upper case? and vice versa? 

If so we need to make sure we update our documentation to make this clear for the Cxs.

AbdulR3hman: so just to add; if we think Cx wants to use Upper case it means that; Schema, Table name, and Columns area ALL upper case;

aimethed: Right now we are only handling table/schema for case insensitivity. The documentation should reflect this.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,201 @@
...
+                String tableNameCaseInsensitively = getTableNameCaseInsensitively(connection, schemaNameCaseInsensitively, tableName.getTableName());
+                TableName tableNameResult = new TableName(schemaNameCaseInsensitively, tableNameCaseInsensitively);
+                LOGGER.info("casing mode is `SEARCH`: performing case insensitive search for TableName object. TableName:{}", tableNameResult);
+                return tableNameResult;
+            case UPPER:
+                TableName upperTableName = new TableName(tableName.getSchemaName().toUpperCase(), tableName.getTableName().toUpperCase());
+                LOGGER.info("casing mode is `UPPER`: adjusting casing from input to upper case for TableName object. TableName:{}", upperTableName);
+                return upperTableName;
+            case LOWER:
+                LOGGER.info("casing mode is `LOWER`: not adjust casing from input for TableName object. TableName:{}", tableName);
```
AbdulR3hman, line 79: I don't think we should return this as is; we need to make sure that we lower case this; we shouldn't trust the caller to pass the original table name in lower case. it should be the responsibility of this method to do so.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,201 @@
...
+    {
+        OracleCasingMode casingMode = getCasingMode(configOptions);
+        switch (casingMode) {
+            case SEARCH:
+                LOGGER.info("casing mode is SEARCH: performing case insensitive search for Schema...");
+                return getSchemaNameCaseInsensitively(connection, schemaNameInput);
+            case UPPER:
+                LOGGER.info("casing mode is `UPPER`: adjusting casing from input to upper case for Schema");
+                return schemaNameInput.toUpperCase();
+            case LOWER:
```
AbdulR3hman, line 97: same here; we shouldn't trust the caller to send the schema name lower cased; it's the responsibility of this method to ensure it.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,201 @@
...
+
+    public static String getSchemaNameCaseInsensitively(final Connection connection, String schemaNameInput)
+            throws SQLException
+    {
+        String nameFromOracle = null;
+        int i = 0;
+        try (PreparedStatement preparedStatement = new PreparedStatementBuilder()
+                .withConnection(connection)
+                .withQuery(SCHEMA_NAME_QUERY_TEMPLATE)
+                .withParameters(Arrays.asList(schemaNameInput.toLowerCase())).build();
```
AbdulR3hman, line 112: why are we lower casing the schema name here? also nit; shouldn't it be just called `schemaName` instead of `schemaNameInput`?

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
aimethed: The decision here was that the three options would be to upper case everything, search, or take whatever the engine gives us and use that. Moving to the third option being to lowercase everything would not be an issue, it is just not the decision we made previously (I know the name LOWER is misleading, I just used it to be consistent with other connectors)

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
aimethed: This is because the query where this is parameter is being used is:
```
"SELECT DISTINCT OWNER as \"OWNER\" FROM all_tables WHERE lower(OWNER) = ?"
```
We lowercase both sides to make it case insensitive

aimethed: Agree with nit though

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    public static String getSchemaNameCaseInsensitively(final Connection connection, String schemaNameInput)
+            throws SQLException
+    {
+        String nameFromOracle = null;
+        int i = 0;
+        try (PreparedStatement preparedStatement = new PreparedStatementBuilder()
+                .withConnection(connection)
+                .withQuery(SCHEMA_NAME_QUERY_TEMPLATE)
```
AbdulR3hman, line 111: this will cause an issue; if the user that is calling this query does not have admin access they won't be able to call `all_tables` as it belongs to the Admin and Admin schema; we need to test this again and make sure our assumptions are correct; if the only way to do that; then we need to revisit this; 

Documentation points says it points to the current user's tables but lets double check. Link to docs: https://docs.oracle.com/cd/B13789_01/server.101/b10755/statviews_1190.htm#sthref1478

more over I believe the schema for `all_tables` should be all caps

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,201 @@
...
+    public static String getTableNameCaseInsensitively(final Connection connection, String schemaName, String tableNameInput)
+            throws SQLException
+    {
+        // schema name input should be correct case before searching tableName already
+        String nameFromOracle = null;
+        int i = 0;
+        try (PreparedStatement preparedStatement = new PreparedStatementBuilder()
+                .withConnection(connection)
+                .withQuery(TABLE_NAME_QUERY_TEMPLATE)
+                .withParameters(Arrays.asList((schemaName), tableNameInput.toLowerCase())).build();
```
AbdulR3hman, line 141: same question as above.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,201 @@
...
+                String schemaNameCandidate = resultSet.getString(SCHEMA_NAME_COLUMN_KEY);
+                LOGGER.debug("Case insensitive search on columLabel: {}, schema name: {}", SCHEMA_NAME_COLUMN_KEY, schemaNameCandidate);
+                nameFromOracle = schemaNameCandidate;
+            }
+        }
+        catch (SQLException e) {
+            throw new RuntimeException(e);
+        }
+
+        if (i != 1) {
```
AbdulR3hman, line 125: shouldn't this be caught on line 114 instead? why are we waiting to finish the loop then throw the exception?

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,201 @@
...
+
+        return nameFromOracle;
+    }
+
+    private static OracleCasingMode getCasingMode(Map<String, String> configOptions)
+    {
+        boolean isGlueConnection = StringUtils.isNotBlank(configOptions.get(DEFAULT_GLUE_CONNECTION));
+        if (!configOptions.containsKey(CASING_MODE)) {
+            LOGGER.info("CASING MODE not set");
+            return isGlueConnection ? OracleCasingMode.LOWER : OracleCasingMode.UPPER;
```
AbdulR3hman, line 166: why are we always using lower casing mode for glue connection?

AbdulR3hman: adding to this; Oracle by default uses upper casing no? shouldn't we default to that instead?

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
aimethed: I believe this table has the behavior we want. I tested it only with the admin user (agree testing with another would be good) and it worked as intended. I believe the behavior of a user only being able to see tables/schemas they have permission for is also the behavior we want. A user shouldn't be able to query for tables they don't have permissions for.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        try {
+            OracleCasingMode oracleCasingMode = OracleCasingMode.valueOf(configOptions.get(CASING_MODE).toUpperCase());
+            LOGGER.info("CASING MODE enable: {}", oracleCasingMode.toString());
+            return oracleCasingMode;
+        }
+        catch (Exception ex) {
```
AbdulR3hman, line 174: we are catching a much wider exception here; the run time exceptions should be propagated back to the engine; in the above scenario; on line 170; if we don't have a proper casing mode set by the Cx we should check there and send an appropriate error instead of trying to catch a wider exception here.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
aimethed: Glue lower cases everything.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
aimethed: What do you mean by we should check there?
As for catching the error, we could change it to catch only the exception thrown by using `valueOf` on a value that doesn't exist in the enum. But I don't think there are any other possible errors thrown here anyway so it shouldn't change the overall result.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
chngpe: >(I know the name LOWER is misleading, I just used it to be consistent with other connectors)

let's fix this, which other connector has lower but not lowering it. we should make it consistent

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
chngpe: >Glue lower cases everything.

+1, We should update documentation. Stating that for connector with Glue connection. EVERYTHING IS lower cased unless you specify otherwise.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
aimethed: Sorry I misremembered, this is set to LOWER because we did a release with this as LOWER before implementing this new Case Resolver class. So for backwards compatibility, I kept it as LOWER. If we want it to be something else (ex: DEFAULT or ENGINE) we should still support any customers who may have it set to LOWER currently by supporting LOWER

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
aimethed: I updated this to catch the IllegalArgumentException instead of the generic Exception

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
aimethed: nit addressed

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
aimethed: we now lower case it

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
aimethed: Changed to lower case it

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
AbdulR3hman: ack

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleMetadataHandler.java
```diff
@@ -95,11 +93,9 @@ public class OracleMetadataHandler
     static final String BLOCK_PARTITION_COLUMN_NAME = "PARTITION_NAME".toLowerCase();
     static final String ALL_PARTITIONS = "0";
     static final String PARTITION_COLUMN_NAME = "PARTITION_NAME".toLowerCase();
-    static final String CASING_MODE = "casing_mode";
     private static final Logger LOGGER = LoggerFactory.getLogger(OracleMetadataHandler.class);
     private static final int MAX_SPLITS_PER_REQUEST = 1000_000;
     private static final String COLUMN_NAME = "COLUMN_NAME";
-    private static final String ORACLE_QUOTE_CHARACTER = "\"";
 
     static final String LIST_PAGINATED_TABLES_QUERY = "SELECT TABLE_NAME as \"TABLE_NAME\", OWNER as \"TABLE_SCHEM\" FROM all_tables WHERE owner = ? ORDER BY TABLE_NAME OFFSET ? ROWS FETCH NEXT ? ROWS ONLY";
```
chngpe, line 100: owner = ? should have case intensively match as well?

chngpe (changes requested):
getPaginatedTables and listPaginatedTables also needs to support insensitive search. Can we making sure we are passing correct parameters?

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleMetadataHandler.java
aimethed: Before we assumed that the case insensitive search from a previous call would pass through because that is the behavior of the trino engine. But other engines in the future may not do this so we should case insensitive search the schema name before passing here. This was just done.

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,203 @@
...
+    // used for identifying database objects (ex: table names)
+    private static final String ORACLE_IDENTIFIER_CHARACTER = "\"";
+    // used in SQL statements for character strings (ex: where OWNER = 'example')
+    private static final String ORACLE_STRING_LITERAL_CHARACTER = "\'";
+
+    private OracleCaseResolver() {}
+
+    private enum OracleCasingMode
+    {
+        LOWER,      // casing mode to use whatever the engine returns (in trino's case, lower)
```
chngpe, line 58: Update Lower comment, as we are adjusting it to lower

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,203 @@
...
+                ResultSet resultSet = preparedStatement.executeQuery()) {
+            while (resultSet.next()) {
+                i++;
+                String schemaNameCandidate = resultSet.getString(SCHEMA_NAME_COLUMN_KEY);
+                LOGGER.debug("Case insensitive search on columLabel: {}, schema name: {}", SCHEMA_NAME_COLUMN_KEY, schemaNameCandidate);
+                nameFromOracle = schemaNameCandidate;
+            }
+        }
+        catch (SQLException e) {
+            throw new RuntimeException(e);
```
chngpe, line 124: Can do you an error message, getSchemaNameCaseInsensitively for `schemaName`

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
```diff
@@ -0,0 +1,203 @@
...
+                ResultSet resultSet = preparedStatement.executeQuery()) {
+            while (resultSet.next()) {
+                i++;
+                String schemaNameCandidate = resultSet.getString(TABLE_NAME_COLUMN_KEY);
+                LOGGER.debug("Case insensitive search on columLabel: {}, schema name: {}", TABLE_NAME_COLUMN_KEY, schemaNameCandidate);
+                nameFromOracle = schemaNameCandidate;
+            }
+        }
+        catch (SQLException e) {
+            throw new RuntimeException(e);
```
chngpe, line 153: nit: better message please

athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleCaseResolver.java
AbdulR3hman: I'm not exactly following; we are checking `i=1` meaning we are expecting only 1 and only 1 result from the `resultSet`; if we get more that means the search is not conclusive since we have more than one table/schema that match the same string. So I am not sure why we can't throw the error in the loop instead of going over the entire loop.

AbdulR3hman: discussed offline; please ignore

aimethed: Oh I see, I thought you were talking about the Runtime Exception thrown when catching the SQLException (the github preview was not clear). If we move the thrown error into the while loop, then we will still need to check for `i=0` outside of the loop. We also lose the information on how many matches there were (knowing 0 matches vs 2+ is helpful)

AbdulR3hman (approved):
pending other comments from Peter addressed

chngpe (approved):

PR #2522, opened 2025-01-16
Upgrade apache-arrow dependency to 18.1.0 and bouncycastle to bcprov-jdk18on:1.79

MarioRial22:
- We are upgrading apache-arrow and bouncy castle dependencies to the current up to date versions.
- Added the maven wrapper and a script to build and publish ECR images of connectors to private ECR repos.


AbdulR3hman:
not needed; done on https://github.com/awslabs/aws-athena-query-federation/pull/2576

PR #2543, opened 2025-01-22
Infer schema from gremlin/sparql query results for Neptune qpt

VenkatasivareddyTR:
*Issue #, if available:*

*Description of changes:*
Currently we are generating schema from aws glue for Neptune qpt as well
[Neptune_QPT_Tests.xlsx](https://github.com/user-attachments/files/18505470/Neptune_QPT_Tests.xlsx)
, this change will get schema from gremlin/sparql query results for Neptune qpt.

### DIFF
--- a/athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneMetadataHandler.java
+++ b/athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneMetadataHandler.java
@@ -22,8 +22,6 @@ package com.amazonaws.athena.connectors.neptune;
 import com.amazonaws.athena.connector.lambda.QueryStatusChecker;
 import com.amazonaws.athena.connector.lambda.data.BlockAllocator;
 import com.amazonaws.athena.connector.lambda.data.BlockWriter;
-import com.amazonaws.athena.connector.lambda.data.FieldBuilder;
-import com.amazonaws.athena.connector.lambda.data.SchemaBuilder;
 import com.amazonaws.athena.connector.lambda.domain.Split;
 import com.amazonaws.athena.connector.lambda.domain.TableName;
 import com.amazonaws.athena.connector.lambda.domain.spill.SpillLocation;
@@ -47,9 +45,7 @@ import com.amazonaws.athena.connectors.neptune.qpt.NeptuneQueryPassthrough;
 import com.amazonaws.athena.connectors.neptune.rdf.NeptuneSparqlConnection;
 import com.google.common.collect.ImmutableMap;
 import org.apache.arrow.util.VisibleForTesting;
-import org.apache.arrow.vector.types.Types;
 import org.apache.arrow.vector.types.pojo.Field;
-import org.apache.arrow.vector.types.pojo.FieldType;
 import org.apache.arrow.vector.types.pojo.Schema;
 import org.apache.tinkerpop.gremlin.driver.Client;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
@@ -64,7 +60,6 @@ import software.amazon.awssdk.services.glue.model.Table;
 import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -282,6 +277,7 @@ public class NeptuneMetadataHandler extends GlueMetadataHandler
         queryPassthrough.verify(qptArguments);
         String schemaName = qptArguments.get(NeptuneQueryPassthrough.DATABASE);
         String tableName = qptArguments.get(NeptuneQueryPassthrough.COLLECTION);
+        String componentTypeValue = qptArguments.get(NeptuneQueryPassthrough.COMPONENT_TYPE);
         TableName tableNameObj = new TableName(schemaName, tableName);
         Schema schema;
         Enums.GraphType graphType = Enums.GraphType.PROPERTYGRAPH;
@@ -304,7 +300,7 @@ public class NeptuneMetadataHandler extends GlueMetadataHandler
                     if (responseObj instanceof Map && gremlinQuery.contains(Constants.GREMLIN_QUERY_SUPPORT_TYPE)) {
                         logger.info("NeptuneMetadataHandler doGetQueryPassthroughSchema gremlinQuery with valueMap");
                         Map graphTraversalObj = (Map) responseObj;
-                        schema = getSchemaFromResults(graphTraversalObj);
+                        schema = NeptuneSchemaUtils.getSchemaFromResults(graphTraversalObj, componentTypeValue, tableName);
                         return new GetTableResponse(request.getCatalogName(), tableNameObj, schema);
                     }
                     else {
@@ -324,8 +320,8 @@ public class NeptuneMetadataHandler extends GlueMetadataHandler
                 NeptuneSparqlConnection neptuneSparqlConnection = (NeptuneSparqlConnection) neptuneConnection;
                 neptuneSparqlConnection.runQuery(sparqlQuery);
                 if (neptuneSparqlConnection.hasNext()) {
-                    Map<String, Object> resultsMap = neptuneSparqlConnection.next(true);
-                    schema = getSchemaFromResults(resultsMap);
+                    Map<String, Object> resultsMap = neptuneSparqlConnection.next();
+                    schema = NeptuneSchemaUtils.getSchemaFromResults(resultsMap, componentTypeValue, tableName);
                     return new GetTableResponse(request.getCatalogName(), tableNameObj, schema);
                 }
                 else {
@@ -336,72 +332,4 @@ public class NeptuneMetadataHandler extends GlueMetadataHandler
                 throw new IllegalArgumentException("Unsupported graphType: " + graphType);
         }
     }
-
-    private Schema getSchemaFromResults(Map resultsMap)
-    {
-        Schema schema;
-        SchemaBuilder schemaBuilder = SchemaBuilder.newBuilder();
-        //Building schema from gremlin/sparql query results.
-        resultsMap.forEach((columnName, columnValue) -> buildSchema(columnName.toString(), columnValue,  schemaBuilder));
-        schema = schemaBuilder.build();
-
-        return schema;
-    }
-
-    private void buildSchema(String columnName, Object columnValue, SchemaBuilder schemaBuilder)
-    {
-        schemaBuilder.addField(getArrowFieldForNeptune(columnName, columnValue));
-    }
-
-    /**
-     * Infers the type of a field from Neptune data.
-     *
-     * @param key The key of the field we are attempting to infer.
-     * @param value A value from the key whose type we are attempting to infer.
-     * @return The Apache Arrow field definition of the inferred key/value.
-     */
-    private Field getArrowFieldForNeptune(String key, Object value)
-    {
-        if (value instanceof String) {
-            return new Field(key, FieldType.nullable(Types.MinorType.VARCHAR.getType()), null);
-        }
-        else if (value instanceof Integer) {
-            return new Field(key, FieldType.nullable(Types.MinorType.INT.getType()), null);
-        }
-        else if (value instanceof Long) {
-            return new Field(key, FieldType.nullable(Types.MinorType.BIGINT.getType()), null);
-        }
-        else if (value instanceof Boolean) {
-            return new Field(key, FieldType.nullable(Types.MinorType.BIT.getType()), null);
-        }
-        else if (value instanceof Float) {
-            return new Field(key, FieldType.nullable(Types.MinorType.FLOAT4.getType()), null);
-        }
-        else if (value instanceof Double) {
-            return new Field(key, FieldType.nullable(Types.MinorType.FLOAT8.getType()), null);
-        }
-        else if (value instanceof java.util.Date) {
-            return new Field(key, FieldType.nullable(Types.MinorType.DATEMILLI.getType()), null);
-        }
-        else if (value instanceof java.util.UUID) {
-            return new Field(key, FieldType.nullable(Types.MinorType.VARCHAR.getType()), null);
-        }
-        else if (value instanceof List) {
-            Field child;
-            if (((List<?>) value).isEmpty()) {
-                logger.warn("getArrowFieldForNeptune: Encountered an empty List for field[{}], defaulting to List<String> due to type erasure.", key);
-                return FieldBuilder.newBuilder(key, Types.MinorType.LIST.getType()).addStringField("").build();
-            }
-            else {
-                child = getArrowFieldForNeptune("", ((List<?>) value).get(0));
-            }
-            return new Field(key, FieldType.nullable(Types.MinorType.LIST.getType()),
-                    Collections.singletonList(child));
-        }
-
-        String className = (value == null || value.getClass() == null) ? "null" : value.getClass().getName();
-        logger.warn("Unknown type[{}] for field[{}], defaulting to varchar.", className, key);
-        return new Field(key, FieldType.nullable(Types.MinorType.VARCHAR.getType()), null);
-    }
-
 }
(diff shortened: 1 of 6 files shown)


athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneMetadataHandler.java
```diff
@@ -296,7 +291,7 @@ public GetTableResponse doGetQueryPassthroughSchema(BlockAllocator allocator, Ge
                 Client client = neptuneConnection.getNeptuneClientConnection();
                 GraphTraversalSource graphTraversalSource = neptuneConnection.getTraversalSource(client);
                 String gremlinQuery = qptArguments.get(NeptuneQueryPassthrough.QUERY);
-                gremlinQuery = gremlinQuery.concat(".limit(1)");
+                gremlinQuery = gremlinQuery.concat(".limit(10)");
```
chngpe, line 294: Is it possible that the first 10 row data is different than the rest of rows?

Example:
Can first 10 rows contains a different result set which has no column field call `resource_type` and 11~20 rows has that? If yes, the schema inferred will drop the column? (and vice-versa, if no column in first 10 rows and data coming from 11~20, when reading data will there be an issue?)

It is okay, if not supported but i want to document the behavior as documentation

Thanks team

VenkatasivareddyTR: Yes, we can update documentation about this limitation.

We determine the schema by examining the first 10 records. If the data exhibits schema variations in further rows, then we may not accurately capture the complete schema. It seems similar approach is followed in DynamoDB connector as well.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/rdf/NeptuneSparqlConnection.java
```diff
@@ -96,61 +96,66 @@ public boolean hasNext()
...
-                    logger.debug("Int" + varName + "=" + oval);
-                } 
-                else if (type.equals(XMLSchema.INTEGER)) {
-                    oval = lit.integerValue();
-                    logger.debug("Integer " + varName + "=" + oval);
-                } 
-                else if (type.equals(XMLSchema.LONG)) {
-                    oval = lit.longValue();
-                    logger.debug("Long " + varName + "=" + oval);
+            if (val != null) {
```
AbdulR3hman, line 106: are we saying here that if value is null; we are still adding it to `ret` map?

AbdulR3hman: alternatively; if we want know that if `value == null` then we return as is; why not check for that and `continue;` 

that way the next code is not indent and easier to read.

VenkatasivareddyTR: > are we saying here that if value is null; we are still adding it to `ret` map?

Yes, to maintain consistency with SPARQL query results, which include column names with empty values, null values are also added to the 'ret' map.

VenkatasivareddyTR: > alternatively; if we want know that if `value == null` then we return as is; why not check for that and `continue;`
> 
> that way the next code is not indent and easier to read.

If we use `value == null` then it's not consistent with SPARQL query results.
image

SPARQL query results:
image

With existing PR code changes:
image

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneSchemaUtilsTest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.ArrayList;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(MockitoJUnitRunner.class)
+public class NeptuneSchemaUtilsTest {
+    private static final Logger logger = LoggerFactory.getLogger(NeptuneSchemaUtilsTest.class);
+
+    private final Map<String, Object> objectMap = ImmutableMap.of(
```
AbdulR3hman, line 40: can we also simulate a null as a test

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/rdf/NeptuneSparqlConnection.java
AbdulR3hman: understood; sounds good.

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneSchemaUtilsTest.java
VenkatasivareddyTR: Hi @AbdulR3hman, If I understand correctly, a null value needs to be added to the objectMap? If so, I have added a null value to objectMap. 
Alternatively, if you expect the objectMap itself to be null, that is not applicable here, as it is a NotNull parameter.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/qpt/NeptuneQueryPassthrough.java
```diff
@@ -56,7 +57,7 @@ public String getFunctionName()
     @Override
     public List<String> getFunctionArguments()
     {
-        return Arrays.asList(DATABASE, COLLECTION, QUERY);
+        return Arrays.asList(DATABASE, COLLECTION, QUERY, COMPONENT_TYPE);
```
AbdulR3hman: this class should be split into two classes; QPT can support multiple signatures; 

We should have different implementation for SPRAQL and Gremlin; one should be called `query` and the other `traverse`

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneMetadataHandler.java
AbdulR3hman: Please declare this as a static variable in `com.amazonaws.athena.connectors.neptune.Constants` class

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneSchemaUtils.java
```diff
@@ -0,0 +1,101 @@
...
+        SchemaBuilder schemaBuilder = SchemaBuilder.newBuilder();
+        //Building schema from gremlin/sparql query results.
+        resultsMap.forEach((columnName, columnValue) -> buildSchema(columnName.toString(), columnValue,  schemaBuilder));
+        schemaBuilder.addMetadata(Constants.SCHEMA_COMPONENT_TYPE, componentTypeValue);
+        schemaBuilder.addMetadata(Constants.SCHEMA_GLABEL, tableName);
+        schema = schemaBuilder.build();
+        return schema;
+    }
+
+    private static void buildSchema(String columnName, Object columnValue, SchemaBuilder schemaBuilder)
```
AbdulR3hman, line 55: We have two clients here; Gremlin and SPARQL; are we saying that the Neptune connection's result will return the same types regardless of the client we are using? based on what are we making this assumption? 

If not; we should have different schema builder and type conversion from Grelmin to arrow; and another for SPARQL?

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/qpt/NeptuneQueryPassthrough.java
VenkatasivareddyTR: Thank you @AbdulR3hman for reviewing, I have added code to support multiple signatures for neptune qpt.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneSchemaUtils.java
VenkatasivareddyTR: Hi @AbdulR3hman Yes, that's correct. 

For Gremlin queries, we are currently limiting support to valueMap() traversals. These traversals return Object types, which our connection logic explicitly converts into a Map representation.

For SPARQL queries, the direct response from Neptune is already structured as a Map.

Therefore, both Gremlin (after conversion) and SPARQL results converge to a Map type. This enables us to use a single, unified function for building the schema from these results.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/qpt/NeptuneGremlinQueryPassthrough.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public Logger getLogger()
+    {
+        return LOGGER;
+    }
+
+    @Override
+    public void customConnectorVerifications(Map<String, String> engineQptArguments)
```
fal-bharadwaj, line 78: I think we should add more validations based on the arguments like null checks and gremlin syntax validations?

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneSchemaUtilsTest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(MockitoJUnitRunner.class)
+public class NeptuneSchemaUtilsTest {
```
fal-bharadwaj: The test can be better organized with a setup that does initialization, and separate test cases for positive and negative usecases including edge cases.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/qpt/NeptuneGremlinQueryPassthrough.java
```diff
@@ -0,0 +1,85 @@
...
+    {
+        // Verify no mixed operations (SPARQL and Gremlin in same request)
+        if (engineQptArguments.containsKey(QUERY)) {
```
fal-bharadwaj, line 85: I think we should add more validations based on the arguments like null checks and gremlin syntax validations?

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneSchemaUtils.java
fal-bharadwaj: Just so I understand correctly, 

 > For Gremlin queries, we are currently limiting support to valueMap() traversals.

- Is that limitation for neptune, federated neptune connector or for QPT feature?

- even though both Gremlin and SPARQL return a Map type, do the returned objects contain same fields/are the fields named identically? are there any particular keys/fields that would need a special handling in either usecase?

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneSchemaUtils.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Infers the type of a field from Neptune data.
+     *
+     * @param key The key of the field we are attempting to infer.
+     * @param value A value from the key whose type we are attempting to infer.
+     * @return The Apache Arrow field definition of the inferred key/value.
+     */
+    private static Field getArrowFieldForNeptune(String key, Object value)
```
fal-bharadwaj, line 67: Are we missing support for some fields for SPARQL here like DECIMAL etc, comparing it with NeptuneSparqlConnection.java I see some are missing, are they being converted to the ones below somewhere?

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/rdf/RDFHandler.java
```diff
@@ -174,11 +174,9 @@ else if (queryMode.equals(Constants.QUERY_MODE_CLASS)) {
         final GeneratedRowWriter rowWriter = builder.build();
 
         // get results
-        String strim = recordsRequest.getSchema().getCustomMetadata().get(Constants.SCHEMA_STRIP_URI);
-        boolean trimURI = strim == null ? false : Boolean.parseBoolean(strim);
```
fal-bharadwaj, line 178: Why did we delete this? We no longer need to trim the result?

fal-bharadwaj (changes requested):
Please take a look at the comments, thanks.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneSchemaUtils.java
VenkatasivareddyTR: > Is that limitation for neptune, federated neptune connector or for QPT feature?

This limitation is only for QPT feature.

> even though both Gremlin and SPARQL return a Map type, do the returned objects contain same fields/are the fields named identically? are there any particular keys/fields that would need a special handling in either usecase?

Even though both Gremlin and SPARQL return a Map type, the returned objects generally do not contain the same fields or identically named fields. The map structure is dynamic and specific to each query result, where keys represent column names and values are the corresponding column values.

We process this dynamic map to prepare the schema on the fly based on the query output. Therefore, no special handling is required for particular keys/fields in either the Gremlin or SPARQL use case, as our schema preparation logic is designed to be adaptive.

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/NeptuneSchemaUtilsTest.java
VenkatasivareddyTR: We have added more tests to cover positive and negative use-cases including edge cases.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/qpt/NeptuneGremlinQueryPassthrough.java
VenkatasivareddyTR: Added validation for Unsupported gremlin query format. Null checks and empty checks are already handled in parent class.
We have removed Unsupported gremlin query format validation check from NeptuneMetadataHandler as we are handling here.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/NeptuneSchemaUtils.java
VenkatasivareddyTR: No, we are not missing. We are converting to varchar, as we don't have support for process decimal values in SparqlRowWriter.writeRowTemplate()/VertexRowWriter.writeRowTemplate() functions. Even if we get decimal value in NeptuneSparqlConnection.java class we are converting to varchar. Let us know if you have any doubts we can discuss further.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/rdf/RDFHandler.java
VenkatasivareddyTR: We never used 'trimURI in NeptuneSparqlConnection.next() function, so we cleaned it.

VenkatasivareddyTR:
> Please take a look at the comments, thanks. Some automated checks also failed, please review them as well.

Thank you @fal-bharadwaj for reviewing, we have addressed your review comments, please let us know if you have any other review comments. Fixed build issue after rebasing.

athena-neptune/src/main/java/com/amazonaws/athena/connectors/neptune/rdf/RDFHandler.java
fal-bharadwaj: Not sure I understand, looks like we did use it in line 181 below and then removed it from being used in neptuneConnection.next()?

VenkatasivareddyTR: In 181 we are passing to neptuneConnection.next() function, however we never used inside next() function. Inside next() function we are using class level variable. Please see below screenshot for your reference.

image

fal-bharadwaj (approved):
looks good, thanks for addressing the comments!

PR #2546, opened 2025-01-23
CLOUDERA-IMPALA Connection String Fix

ritiktrianz:
*Issue #, if available:*
The connection string was not forming correctly, which caused an error. 

Expected Connection string should be in "jdbc:impala://50.100.00.00:49172/default;UID=username;PWD=pass"  format but we were getting in "jdbc:impala://50.000.00.00:49172/?UID=username;PWD=pass".

After making these code changes, it is now successful. PFA test doc for LF testing with glue.
[CLOUDERA_IMPALA_PHASE2.xlsx](https://github.com/user-attachments/files/18520945/CLOUDERA_IMPALA_PHASE2.xlsx)


AbdulR3hman (approved):

AbdulR3hman:
same here; will approve; and request doc changes prior to release cut

PR #2571, opened 2025-02-03
add package templates

aimethed:
*Description of changes:*
For all connectors that have jar/zip files small enough to be deployed directly to lambda, the templates are being added back. These templates will not be deployed to standard partition regions but can be used to manually deploy the connectors.

Snowflake and Bigquery connector zip/jar files are too large. GCS and Vertica connector zip/jar files are very close, so they are included for now to test.


aimethed:
PR where the teradata was initially removed: https://github.com/awslabs/aws-athena-query-federation/pull/2294/files

GCS and Vertica were tested and the unzipped size was too large, so the templates have been removed.

chngpe (approved):

PR #2588, opened 2025-02-13
Msk connector was throwing null pointer exception when entire record is null, issue fix.

VenkatasivareddyTR:
*Issue #, if available:*

*Description of changes:*
Msk connector was throwing null pointer exception when it encountered a completely null record. This change resolves that issue.
PFA testing doc for reference.
[MSK NullPointerException Issue Fix.docx](https://github.com/user-attachments/files/18782772/MSK.NullPointerException.Issue.Fix.docx)


ejeffrli (approved):
Known codecov partial issue. record == null is not hit because record cannot be mocked to null.

PR #2591, opened 2025-02-14
Docdb credential refactor

ejeffrli:
*Issue #, if available:*

*Description of changes:*
Moved RdsSecretsCredentialProvider, StaticCredentialsProvider, JdbcCredentialProvider under athena-federation-sdk
Renamed RdsSecretsCredentialProvider to DefaultCredentialsProvider since this is also used for docdb and elasticsearch
Renamed JdbcCredentialProvider to CredentialsProvider since this is no longer only used by JDBC

Docdb currently does not support credentials in username password json format below.
```
{
username: test
password: test
}
```

This PR refactors RdsSecretsCredentialsProvider to become usable by all connectors since more than jdbc connectors follow this json format. Then, the code checks if secret is parse-able to username/password map (in docdb connector only). If not, we use default behavior of embedding string directly. Adding new resolveWithDefaultCredentials function for docdb use case, but can be further used by classes who have secrets in this format. Hbase uses resolveSecrets, but its secrets have no format limitation. I decided it was not possible to modify resolveSecrets safely and added a new function, resolveWithDefaultCredentials, that catches possible formatting error and instead performs the same operation as resolveSecrets.


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/CachableSecretsManager.java
```diff
@@ -85,6 +87,31 @@ public String resolveSecrets(String rawString)
...
     }
 
+    public String resolveWithDefaultCredentials(String rawString)
+    {
+        Matcher m = PATTERN.matcher(rawString);
+        String result = rawString;
+        while (m.find()) {
+            String nextSecret = m.group(1);
+            Matcher m1 = NAME_PATTERN.matcher(nextSecret);
+            m1.find();
```
chngpe, line 97: do we want to check if m1.find = true then proceed to next?

ejeffrli: no i don't think it's necessary. m1.find() is a subset pattern of m. It is trying to get the secret_name string within '${secret_name}' from nextSecret

chngpe (approved):

PR #2595, opened 2025-02-16
Fixed Encoding Warning

AbdulR3hman:
*Issue #, if available:*
All our current build have thousands of line of encoding warning making any github action that has a maven build extremely hard to read; this should fix it. 

*Description of changes:*
Enforce encoding setting on the parent project properties; this is following [stack overflow suggestion](https://stackoverflow.com/questions/3017695/how-can-i-configure-encoding-in-maven/3018152#3018152)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,6 +9,8 @@
     <description>The Amazon Athena Query Federation SDK allows you to customize Amazon Athena with your own code.</description>
     <url>https://github.com/awslabs/aws-athena-query-federation</url>
     <properties>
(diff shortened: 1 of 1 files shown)


AbdulR3hman:
the fix clearly fixed the warning of the encoding; for more info this is on the Maven FAQ: [How do I prevent “[WARNING] Using platform encoding (Cp1252 actually) to copy filtered resources, i.e. build is platform dependent!”](https://maven.apache.org/general.html#encoding-warning)

chngpe (approved):

PR #2615, opened 2025-02-18
Fix Synapse Connector to Support Case-Sensitive Collation

Jithendar12:
*Issue #, if available:*
1) Invalid object name SYS.COLUMNS error occurs in Synapse Connector when case-sensitive collation is enabled.
image

2) Unable to query uppercase table names when case-sensitive collation is enabled.
image

*Description of changes:*
1) Uppercase SYS.COLUMNS system view is not available in Synapse when case-sensitive collation is enabled. Changed it to lowercase sys.columns.
2) Added a method getExactCaseObjectName() which retrieves the exact schema and table name from Synapse instead of converting it to uppercase.

Please find attached functional test results.
[SYNAPSE_FUNCTIONAL_TEST_2025-02-18_10_47_47.742843.csv](https://github.com/user-attachments/files/18843005/SYNAPSE_FUNCTIONAL_TEST_2025-02-18_10_47_47.742843.csv)

[synapse-cs-collation-issue.odt](https://github.com/user-attachments/files/18843383/synapse-cs-collation-issue.odt)

### DIFF
--- a/athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseCaseInsensitiveResolver.java
+++ b/athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseCaseInsensitiveResolver.java
@@ -34,6 +34,7 @@ public class SynapseCaseInsensitiveResolver
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(SynapseCaseInsensitiveResolver.class);
     private static final String OBJECT_NAME_QUERY_TEMPLATE = "SELECT TABLE_SCHEMA, TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE LOWER(TABLE_SCHEMA) = ? AND LOWER(TABLE_NAME) = ?";
+    private static final String SCHEMA_NAME_QUERY_TEMPLATE = "SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE LOWER(SCHEMA_NAME) = ?";
     private static final String CASING_MODE = "casing_mode";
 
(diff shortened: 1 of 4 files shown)


chngpe (changes requested):
I think we will need to override `doListTables` method from JDBC Metadata Handler. 
Imaging this use case, customer send request for list table, if we don't have case insensitive search, we will not get the correct schema name, there won't be any tables return due to schema mismatch.


Example1: Snowflake
https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java#L394-L413

Example2: Oracle
https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleMetadataHandler.java#L251-L260

Jithendar12:
> I think we will need to override `doListTables` method from JDBC Metadata Handler. Imaging this use case, customer send request for list table, if we don't have case insensitive search, we will not get the correct schema name, there won't be any tables return due to schema mismatch.
> 
> Example1: Snowflake https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java#L394-L413
> 
> Example2: Oracle https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-oracle/src/main/java/com/amazonaws/athena/connectors/oracle/OracleMetadataHandler.java#L251-L260

Hi @chngpe , 
Thank you for reviewing.  I've incorporated the changes.

chngpe (approved):

PR #2667, opened 2025-03-19
Fix sqlserver casing mode

ritiktrianz:
*Issue #, if available:*
On Athena SQLServer was returning empty table list when Case Sensitive Collation is enabled at Database Level.
After making these code changes,it is working fine for different casing modes like 'UPPER', 'lower' and 'CASE_INSENSITIVE_SEARCH'.

### DIFF
--- a/athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
+++ b/athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
@@ -99,8 +99,8 @@ public class SqlServerMetadataHandler extends JdbcMetadataHandler
      * (When non-partitioned table have indexes partition_number will be 1, so these rows will be skipped with where condition)
      * table have both index and partition - rows will be returned with different partition_number as we are using distinct in the query.
      */
-    static final String GET_PARTITIONS_QUERY = "select distinct PARTITION_NUMBER from SYS.DM_DB_PARTITION_STATS where object_id = OBJECT_ID(?) and partition_number > 1 "; //'dbo.MyPartitionTable'
-    static final String ROW_COUNT_QUERY = "select count(distinct PARTITION_NUMBER) as row_count from SYS.DM_DB_PARTITION_STATS where object_id = OBJECT_ID(?) and partition_number > 1 ";
+    static final String GET_PARTITIONS_QUERY = "select distinct PARTITION_NUMBER from sys.dm_db_partition_stats where object_id = OBJECT_ID(?) and partition_number > 1 "; //'dbo.MyPartitionTable'
+    static final String ROW_COUNT_QUERY = "select count(distinct PARTITION_NUMBER) as row_count from sys.dm_db_partition_stats where object_id = OBJECT_ID(?) and partition_number > 1 ";
 
     private static final int MAX_SPLITS_PER_REQUEST = 1000_000;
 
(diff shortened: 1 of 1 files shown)


athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
```diff
@@ -127,7 +127,7 @@ public class SqlServerMetadataHandler extends JdbcMetadataHandler
             "WHERE t.object_id = (select object_id from sys.objects o where o.name = ? " +
             "and schema_id = (select schema_id from sys.schemas s where s.name = ?))";
     static final String VIEW_CHECK_QUERY = "select TYPE_DESC from sys.objects where name = ? and schema_id = (select schema_id from sys.schemas s where s.name = ?)";
-    static final String LIST_PAGINATED_TABLES_QUERY = "SELECT o.name AS \"TABLE_NAME\", s.name AS \"TABLE_SCHEM\" FROM sys.objects o INNER JOIN sys.schemas s ON o.schema_id = s.schema_id WHERE o.type IN ('U', 'V') and s.name = ? ORDER BY table_name OFFSET ? ROWS FETCH NEXT ? ROWS ONLY;";
```
chngpe, line 130: List pagination table is using this class which is looking for UPPER case name. Are we sure this change will works?

https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtil.java#L206-L226

chngpe: Line 177

athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
```diff
@@ -423,8 +423,8 @@ protected Schema getSchema(Connection jdbcConnection, TableName tableName, Schem
             throws Exception
     {
         String dataTypeQuery = "SELECT C.NAME AS COLUMN_NAME, TYPE_NAME(C.USER_TYPE_ID) AS DATA_TYPE " +
-                "FROM SYS.COLUMNS C " +
-                "JOIN SYS.TYPES T " +
+                "FROM sys.columns C " +
```
chngpe, line 426: is this bacward compatible of customer does not enable `Case Sensitive Collation`

athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
ritiktrianz: @chngpe , following your comment, we debugged the issue more and found that "ORDER BY table_name(lower case)" didn't match with the column name "TABLE_NAME (UPPER case)" due to case-sensitive collation at the database level. We updated the changes to align with the class and preserve the casing of "TABLE_NAME" and "TABLE_SCHEM". Please review and let us know if further any changes are needed.

athena-sqlserver/src/main/java/com/amazonaws/athena/connectors/sqlserver/SqlServerMetadataHandler.java
ritiktrianz: Hi @chngpe ,
 We successfully tested the backward compatibility of the code changes with case-sensitive collation disabled at the database level. PFA doc with screenshots of testing and functional testing.
[SQLSERVER_FUNCTIONAL_TEST_2025-03-20_14_06_05.956407.csv](https://github.com/user-attachments/files/19364061/SQLSERVER_FUNCTIONAL_TEST_2025-03-20_14_06_05.956407.csv)
[sqlserver_backward_compatibility.docx](https://github.com/user-attachments/files/19364062/sqlserver_backward_compatibility.docx)

PR #2698, opened 2025-04-01
Updated rds truststore to default truststore for oracle and docdb.

VenkatasivareddyTR:
*Issue #, if available:*

*Description of changes:*
Enabling SSL for the Oracle connector resulted in a 'certificate path not found' error. This was resolved by using the default truststore. To maintain consistency, the same truststore configuration was applied to DocDB.

### DIFF
--- a/athena-docdb/Dockerfile
+++ b/athena-docdb/Dockerfile
@@ -5,8 +5,8 @@ FROM public.ecr.aws/lambda/java:${JAVA_VERSION}
 
 # Argument for Java tool options, defaulting to an empty string
 ARG JAVA_TOOL_OPTIONS=""
-# Set the JAVA_TOOL_OPTIONS environment variable for Java 17
-ENV JAVA_TOOL_OPTIONS=${JAVA_TOOL_OPTIONS}
+# Set the JAVA_TOOL_OPTIONS environment variable for Java 17 and for ssl
+ENV JAVA_TOOL_OPTIONS="${JAVA_TOOL_OPTIONS} -Djavax.net.ssl.trustStore=/var/lang/lib/security/cacerts -Djavax.net.ssl.trustStorePassword=changeit"
 
 # Install necessary tools
 RUN yum update -y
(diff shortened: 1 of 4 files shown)


athena-docdb/Dockerfile
```diff
@@ -5,8 +5,8 @@ FROM public.ecr.aws/lambda/java:${JAVA_VERSION}
 
 # Argument for Java tool options, defaulting to an empty string
 ARG JAVA_TOOL_OPTIONS=""
-# Set the JAVA_TOOL_OPTIONS environment variable for Java 17
-ENV JAVA_TOOL_OPTIONS=${JAVA_TOOL_OPTIONS}
+# Set the JAVA_TOOL_OPTIONS environment variable for Java 17 and for ssl
+ENV JAVA_TOOL_OPTIONS="${JAVA_TOOL_OPTIONS} -Djavax.net.ssl.trustStore=/var/lang/lib/security/cacerts -Djavax.net.ssl.trustStorePassword=changeit"
```
aimethed, line 9: Can you add these in the `ARG JAVA_TOOL_OPTIONS` above since they are options?

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBConnectionFactory.java
```diff
@@ -56,8 +56,8 @@ public synchronized MongoClient getOrCreateConn(String connStr)
             //Setup SSL Trust Store:
             if (connStr.toLowerCase().contains("ssl=true")) {
                 logger.info("MongoClient is using SSL; thus setting up System properties for trust store");
-                System.setProperty("javax.net.ssl.trustStore", "rds-truststore.jks");
-                System.setProperty("javax.net.ssl.trustStorePassword", "federationStorePass");
+                System.setProperty("javax.net.ssl.trustStoreType", "JKS");
+                System.setProperty("javax.net.ssl.trustStorePassword", "changeit");
```
aimethed, line 60: This is repetitive right? Since this variable is set in the Dockerfile in `JAVA_TOOL_OPTIONS`?

aimethed (changes requested):
Is there a need to change the truststore and storepassword environment variables/`JAVA_TOOL_OPTIONS` from what they were? These appear to just be the password and location of the certs, so I don't think they need to be updated.

athena-docdb/Dockerfile
VenkatasivareddyTR: @aimethed Added to ARG JAVA_TOOL_OPTIONS.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBConnectionFactory.java
VenkatasivareddyTR: Hi @aimethed the stuststore password has been removed from the Dockerfile, enabling unified code for both package and image deployments, eliminating redundancy.

VenkatasivareddyTR:
> Is there a need to change the truststore and storepassword environment variables/`JAVA_TOOL_OPTIONS` from what they were? These appear to just be the password and location of the certs, so I don't think they need to be updated.

Yes, @aimethed to resolve 'Unable to find valid certification path to the requested target' error, which occurred with the custom RDS certificates in the existing truststore and password, we are implementing the default truststore and password.

aimethed (approved):

PR #2732, opened 2025-04-21
Added unit tests for JDBC module

Trianz-Akshay:
*Issue #, if available:*

*Description of changes:*
Added unit test cases for JDBC module.

### DIFF
--- a/athena-db2/src/main/java/com/amazonaws/athena/connectors/db2/Db2Constants.java
+++ b/athena-db2/src/main/java/com/amazonaws/athena/connectors/db2/Db2Constants.java
@@ -35,6 +35,9 @@ public class Db2Constants
             "from syscat.tables " +
             "where type in ('T', 'U', 'V', 'W') and " +
             "tabschema = ? order by tabname";
+    public static final String LIST_PAGINATED_TABLES_QUERY = "select tabschema AS \"TABLE_SCHEM\", tabname AS \"TABLE_NAME\" " +
+            "from syscat.tables where type in ('T', 'U', 'V', 'W') and tabschema = ? " +
+            "ORDER BY tabname OFFSET ? ROWS FETCH NEXT ? ROWS ONLY";
 
     static final String PARTITION_QUERY = "SELECT DATAPARTITIONID FROM SYSCAT.DATAPARTITIONS WHERE TABSCHEMA = ? AND TABNAME = ? AND SEQNO > 0";
     static final String COLUMN_INFO_QUERY = "select colname, typename from syscat.columns where tabschema = ? AND tabname = ?";
--- a/athena-db2/src/main/java/com/amazonaws/athena/connectors/db2/Db2MetadataHandler.java
+++ b/athena-db2/src/main/java/com/amazonaws/athena/connectors/db2/Db2MetadataHandler.java
@@ -37,12 +37,15 @@ import com.amazonaws.athena.connector.lambda.metadata.GetSplitsResponse;
 import com.amazonaws.athena.connector.lambda.metadata.GetTableLayoutRequest;
 import com.amazonaws.athena.connector.lambda.metadata.ListSchemasRequest;
 import com.amazonaws.athena.connector.lambda.metadata.ListSchemasResponse;
+import com.amazonaws.athena.connector.lambda.metadata.ListTablesRequest;
+import com.amazonaws.athena.connector.lambda.metadata.ListTablesResponse;
 import com.amazonaws.athena.connector.lambda.metadata.optimizations.DataSourceOptimizations;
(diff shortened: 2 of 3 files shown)


burhan94:
Can we consider using GenAI here to add even more unit tests? I'd like to see 30-50 UT's get generated that test the different expressions and the resulting translations

ritiktrianz:
Hi @burhan94  ,Thank you for the feedback. We already tried GenAI using Claude Sonnet models, for unit test generation. It generated large volume of test cases covering multiple scenarios, such as edge cases, negative paths, and broader input variations. We felt that not all of these test cases may be relevant and may create unnecessary overhead. So, we removed a few tests and kept only the relevant ones.
 
However, now we have added the generated tests after reviewing them manually.

burhan94:
Can we add some tests for ORDER BY, and then a test that has ORDER BY, LIMIT, and a complex expression? I'd like to also validate that we're testing the different complex expressions supported by the framework.

Trianz-Akshay:
> Can we add some tests for ORDER BY, and then a test that has ORDER BY, LIMIT, and a complex expression? I'd like to also validate that we're testing the different complex expressions supported by the framework.

Thanks for the review! We already had tests for ORDER BY and LIMIT functionality. Now we have added comprehensive tests for complex expressions functionality.

burhan94 (dismissed):

athena-jdbc/src/test/java/com/amazonaws/athena/connectors/jdbc/JdbcEnvironmentPropertiesTest.java
```diff
@@ -0,0 +1,154 @@
...
+        String specialSecret = "secret/name#1";
+        connectionProperties.put(SECRET_NAME, specialSecret);
+        
+        Map<String, String> result = jdbcEnvironmentProperties.connectionPropertiesToEnvironment(connectionProperties);
+        String expected = BASE_CONNECTION_STRING + "?${" + specialSecret + "}";
+        assertEquals(expected, result.get(DEFAULT));
+    }
+
+    @Test
+    public void testMultipleSecretsAndParams() {
```
ejeffrli: do we allow multiple secrets? I think we do not. Maybe possible to remove this test

athena-jdbc/src/test/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtilTest.java
```diff
@@ -121,22 +126,22 @@ public void createJdbcMetadataHandlerMap()
...
+        Assert.assertEquals(FakeDatabaseJdbcMetadataHandler.class, catalogs.get("lambda:functionName").getClass());
     }
 
     @Test(expected = IllegalArgumentException.class)
     public void createJdbcMetadataHandlerMapDifferentDatabasesThrows()
     {
-        Map<String, JdbcMetadataHandler> catalogs = JDBCUtil.createJdbcMetadataHandlerMap(ImmutableMap.<String, String>builder()
+        JDBCUtil.createJdbcMetadataHandlerMap(ImmutableMap.<String, String>builder()
                 .put("testCatalog1_connection_string", CONNECTION_STRING1)
                 .put("testCatalog2_connection_string", CONNECTION_STRING2)
```
ejeffrli: Can we make this test more clear why it throws IllegalArgumentException? I think it is because CONNECTION_STRING1 and CONNECTION_STRING2 are prepended by the wrong engine type.

ejeffrli: Maybe instead of just testCatalog1_connection_string and testCatalog2_connection_String, one can be "wrong_connection_string". 

Please check why this is throwing IllegalArgumentException before

athena-jdbc/src/test/java/com/amazonaws/athena/connectors/jdbc/JdbcEnvironmentPropertiesTest.java
ritiktrianz: Hi @ejeffrli , thanks for the review comment. We’ve removed this test as suggested.

athena-jdbc/src/test/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtilTest.java
ritiktrianz: @ejeffrli , We’ve updated the test and verified that exception occurs because one of the provided connection strings is for the "notrealdb" engine, which is not aligned with the "fakedatabase" engine type that the handler is configured for.

PR #2733, opened 2025-04-23
added unit tests for athena-redshift.

VenkatasivareddyTR:
*Issue #, if available:*

*Description of changes:*
Added unit tests for athena-redshift connector. Coverage is increase from 38% to 84%.


burhan94:
Can you add some UT's to cover the redshift SQL generation from the constraints object? Basically create a bunch of different scenarios that test every predicate type for redshift that we support today. I'd like to see all of the supported complex expressions, order by, limit, top N, etc.

ritiktrianz:
Hi @burhan, thank you for the review comment. We’ve verified that the class for generating redshift SQL generation is extended from PostgreSQL and we have added UT's to cover supported predicate types for Postgres. Kindly review.

PR #2742, opened 2025-04-29
add mock TPCDSMockMetadataHandler for pagination testing

chngpe:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/athena-tpcds/src/main/java/com/amazonaws/athena/connectors/tpcds/mock/pagination/TPCDSMockMetadataHandler.java
+++ b/athena-tpcds/src/main/java/com/amazonaws/athena/connectors/tpcds/mock/pagination/TPCDSMockMetadataHandler.java
@@ -165,10 +165,10 @@ public class TPCDSMockMetadataHandler extends TPCDSMetadataHandler
 
     private List<TableName> getPaginatedList(List<TableName> inputList, int startToken, int pageSize)
     {
-        if (startToken >= 125) {
+        if (startToken > inputList.size()) {
             return List.of();
         }
-        int endToken = Math.min(startToken + pageSize, 125);
+        int endToken = Math.min(startToken + pageSize, inputList.size());
 
         return inputList.subList(startToken, endToken);
     }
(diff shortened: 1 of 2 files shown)


athena-tpcds/src/main/java/com/amazonaws/athena/connectors/tpcds/mock/pagination/TPCDSMockMetadataHandler.java
```diff
@@ -0,0 +1,190 @@
...
+        if (tableName.getTableName().startsWith(MOCK_TABLE_NAME_PREFIX)) {
+            return Table.CUSTOMER;
+        }
+
+        return TPCDSUtils.validateTable(tableName);
+    }
+
+    private List<TableName> getPaginatedList(List<TableName> inputList, int startToken, int pageSize)
+    {
+        if (startToken >= 125) {
```
aimethed, line 168: Why 125? Seems like it should be a constant, either with a name that explains or a comment for that constant

chngpe: Let me move it as constant. We currently only generated fake table up to `125`

athena-tpcds/src/main/java/com/amazonaws/athena/connectors/tpcds/mock/pagination/TPCDSMockMetadataHandler.java
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Used to get the list of static tables from TerraData's TPCDS generator.
+     *
+     * @see MetadataHandler
+     */
+    @Override
+    public ListTablesResponse doListTables(BlockAllocator allocator, ListTablesRequest request)
+    {
+        logger.info("doListTables: enter - " + request);
+        List<TableName> tableList = getTableList(request);
```
samarsar, line 113: Does this pull all tables into memory regardless of if pageSize == UNLIMITED_PAGE_SIZE_VALUE?

should this go into an else clause after the check `if (request.getPageSize() != UNLIMITED_PAGE_SIZE_VALUE) {` so that it would pull all results only if we're not doing pagination?

samarsar (approved):

samarsar (approved):

PR #2745, opened 2025-05-01
adding fake pagination for fallback jdbc logic

samarsar:
*Issue #, if available:*
Adding Fake Pagination to default listPaginatedTables in jdbcMetadataHandler.

*Description of changes:*
Every connector should have the method listPaginatedTables overridden and implemented since each connectors pagination logic. Some connectors don't have it overridden in which case the logic defaults to the method `listPaginatedTables` in jdbcMetadataHandler - which did not do any pagination.

We want to 'fake' pagination here. All tables are being pulled into memory, however, we will return a subset of those tables based on the nextToken and pageSize provided in the ListTablesRequest object to mimic the effects of pagination. 

### DIFF
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
@@ -241,7 +241,7 @@ public abstract class JdbcMetadataHandler
     protected ListTablesResponse listPaginatedTables(final Connection connection, final ListTablesRequest listTablesRequest) throws SQLException
     {
         String adjustedSchemaName = caseResolver.getAdjustedSchemaNameString(connection, listTablesRequest.getSchemaName(), configOptions);
-        LOGGER.info("Request is asking for pagination, but true pagination has not been implemented.");
+        LOGGER.debug("Request is asking for pagination, but true pagination has not been implemented.");
 
         int startToken;
         try {
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
@@ -230,8 +230,9 @@ public abstract class JdbcMetadataHandler
     }
 
     /**
-     * This is default getAllTables no pagination.
-     * Override this if you want to support the behavior.
+     * This is default getAllTables without true pagination.
+     * Paginated list of tables will be returned by retrieving all tables first, then returning subset based off request.
+     * Override this if you want to support true pagination behavior.
      * @param connection
      * @param listTablesRequest
      * @return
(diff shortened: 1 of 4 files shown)


athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
```diff
@@ -230,21 +230,53 @@ public ListTablesResponse doListTables(final BlockAllocator blockAllocator, fina
...
      * @return
      * @throws SQLException
      */
     protected ListTablesResponse listPaginatedTables(final Connection connection, final ListTablesRequest listTablesRequest) throws SQLException
     {
-        // no-op is call listTables
-        // override this function to implement pagination
         String adjustedSchemaName = caseResolver.getAdjustedSchemaNameString(connection, listTablesRequest.getSchemaName(), configOptions);
-        LOGGER.debug("Request is asking for pagination, but pagination has not been implemented");
+        LOGGER.info("Request is asking for pagination, but true pagination has not been implemented.");
```
samarsar, line 244: I accidentally changed this from debug to info for testing purposes. Can change it back in next revision after addressing other comments.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
```diff
@@ -230,21 +230,53 @@ public ListTablesResponse doListTables(final BlockAllocator blockAllocator, fina
...
      * @param connection
      * @param listTablesRequest
      * @return
      * @throws SQLException
      */
     protected ListTablesResponse listPaginatedTables(final Connection connection, final ListTablesRequest listTablesRequest) throws SQLException
     {
-        // no-op is call listTables
-        // override this function to implement pagination
         String adjustedSchemaName = caseResolver.getAdjustedSchemaNameString(connection, listTablesRequest.getSchemaName(), configOptions);
```
chngpe, line 243: I see we are checking `if (pageSize == UNLIMITED_PAGE_SIZE_VALUE ||` 2 times below. Given this is for paginated table. this method should only called when the pageSize not equal to UNLIMITED_PAGE_SIZE_VALUE. We can peroform validation early on this.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
```diff
@@ -230,21 +230,53 @@ public ListTablesResponse doListTables(final BlockAllocator blockAllocator, fina
...
+        if (startToken >= allTables.size()) {
+            return new ListTablesResponse(listTablesRequest.getCatalogName(), List.of(), null);
+        }
+
+        int endToken = Math.min(startToken + pageSize, allTables.size());
+        if (pageSize == UNLIMITED_PAGE_SIZE_VALUE) {
+            endToken = allTables.size();
+        }
+
+        String nextToken;
```
chngpe, line 269: next token can be simplify as startToken + pagesize >= allTables.size then null. else startToken+pagesize.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
samarsar: So in the outer method `doListTables` we check for the condition `if pageSize == UNLIMITED_PAGE_SIZE_VALUE && nextToken == null`. This means that we're starting from zero and pageSize is unlimited so retrieve all items (no pagination)

In this next case in `listPaginatedTables`, there is a possibility where we start at token 2, but pageSize is UNLIMITED_PAGE_SIZE_VALUE. So we will retrieve all tables after index 2. 

Let me know if this makes sense.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
samarsar: This doesn't work because UNLIMITED_PAGE_SIZE_VALUE is actually equal to -1. So in the case where startToken is 2 and pageSize is -1, the condition you listed above will evaluate to 

Hypothetically:
startToken = 2
pageSize = -1
allTables.size = 1-

2 + (-1) >= 10 [false] so endToken will equal 2 + (-1) which would not be correct.

I could do this: `int endToken = (pageSize == UNLIMITED_PAGE_SIZE_VALUE) ? allTables.size() : Math.min(startToken + pageSize, allTables.size());` however, I prefer more verbose code as opposed to ternary because it's easier to read IMO.

samarsar: Starting at anything other than 0 with pageSize unlimited doesn't feel like a very common case though

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
AbdulR3hman: I would say revert it here because we get a lot of customers who would this this is an error;

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
```diff
@@ -230,21 +230,53 @@ public ListTablesResponse doListTables(final BlockAllocator blockAllocator, fina
...
-        // override this function to implement pagination
         String adjustedSchemaName = caseResolver.getAdjustedSchemaNameString(connection, listTablesRequest.getSchemaName(), configOptions);
-        LOGGER.debug("Request is asking for pagination, but pagination has not been implemented");
+        LOGGER.info("Request is asking for pagination, but true pagination has not been implemented.");
+
+        int startToken;
+        try {
+            startToken = listTablesRequest.getNextToken() == null ? 0 : Integer.parseInt(listTablesRequest.getNextToken());
+        }
+        catch (NumberFormatException e) {
```
AbdulR3hman, line 250: can we please add a quick test case for this;

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
samarsar: changed back to LOGGER.debug in next revision.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
samarsar: Added in next revision. It gets translated to IllegalArugmentException

AbdulR3hman (approved):

PR #2758, opened 2025-05-09
Added pagination for synapse connector

VenkatasivareddyTR:
*Issue #, if available:*
Synapse Connector pagination is not implemented.

*Description of changes:*
Added pagination for synapse connector

### DIFF
--- a/athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
+++ b/athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
@@ -321,7 +321,7 @@ public class SynapseMetadataHandler extends JdbcMetadataHandler
     @Override
     public ListTablesResponse listPaginatedTables(final Connection connection, final ListTablesRequest listTablesRequest) throws SQLException
     {
-        LOGGER.debug("Starting listPaginatedTables for Snowflake.");
+        LOGGER.debug("Starting listPaginatedTables for Synapse.");
         String token = listTablesRequest.getNextToken();
         int pageSize = listTablesRequest.getPageSize();
 
--- a/athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
+++ b/athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
@@ -35,6 +35,8 @@ import com.amazonaws.athena.connector.lambda.metadata.GetDataSourceCapabilitiesR
 import com.amazonaws.athena.connector.lambda.metadata.GetSplitsRequest;
 import com.amazonaws.athena.connector.lambda.metadata.GetSplitsResponse;
 import com.amazonaws.athena.connector.lambda.metadata.GetTableLayoutRequest;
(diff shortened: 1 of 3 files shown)


athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
```diff
@@ -310,6 +318,41 @@ private String encodeContinuationToken(int partition)
...
+
+        int t = token != null ? Integer.parseInt(token) : 0;
+
+        String adjustedSchemaName = caseResolver.getAdjustedSchemaNameString(connection, listTablesRequest.getSchemaName(), configOptions);
+
+        LOGGER.info("Starting pagination at {} with page size {}", t, pageSize);
+        List<TableName> paginatedTables = getPaginatedTables(connection, adjustedSchemaName, t, pageSize);
+        LOGGER.info("{} tables returned. Next token is {}", paginatedTables.size(), t + pageSize);
+
+        String nextToken = paginatedTables.isEmpty() || paginatedTables.size() < pageSize ? null : Integer.toString(t + pageSize);
```
samarsar, line 336: There's a chance to get into `listPaginatedTables` method when startToken is not null, but pageSize is UNLIMITED_PAGE_SIZE_VALUE (-1).

With that being said, in the case where start token is 2 and pageSize is -1, this condition will return the incorrect next token. 

paginatedTables.isEmpty() = false
paginatedTables.size() < -1 = false

Then nextToken would be (2 + -1) which would be incorrect.

athena-synapse/src/test/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandlerTest.java
```diff
@@ -342,6 +341,58 @@ public void doGetSplitsContinuation()
...
+        resultSet = mockResultSet(schema, values, new AtomicInteger(-1));
+        Mockito.when(preparedStatement.executeQuery()).thenReturn(resultSet);
+        listTablesResponse = this.synapseMetadataHandler.doListTables(
+                blockAllocator, new ListTablesRequest(this.federatedIdentity, "testQueryId",
+                        "testCatalog", "testSchema", "1", 1));
+        Assert.assertEquals("2", listTablesResponse.getNextToken());
+        Assert.assertArrayEquals(expected, listTablesResponse.getTables().toArray());
+
+
+        // Test 3: Testing single table returned when requesting pageSize 2 signifying end of pagination where nextToken is null.
```
samarsar, line 380: Can we add a fourth test where pageSize = UNLIMITED_PAGE_SIZE_VALUE and startToken is not null?

athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
samarsar: This mistake is also in my snowflake pagination review. Will get that fixed ASAP.

samarsar (changes requested):

VenkatasivareddyTR:
Following our Slack discussion, we've decided to skip implementing true pagination for Synapse due to our use of ROW_NUMBER() instead of offset. Therefore, I'm closing this PR.

Issue #2744, opened 2025-04-30
[FEATURE] Support OAuth for Snowflake


macohen:
**Is your feature request related to a problem? If yes, please describe.**
In November 2025, Snowflake will turn off username/password authentication: https://www.snowflake.com/en/blog/blocking-single-factor-password-authentification/ rendering all existing connectors broken. 

**Describe the solution you'd like**
Adding support in the connector to use OAuth access tokens: https://docs.snowflake.com/en/user-guide/oauth-intro?_ga=2.68114131.359822656.1746026802-1756365676.1743197756. 

**Describe alternatives you've considered**
Authentication through key-pairs is being added in a different PR #2735.

AbdulR3hman:
assigned to @VenkatasivareddyTR

VenkatasivareddyTR:
@macohen Can you please assign this ticket to me (@VenkatasivareddyTR).

ozakidai:
@macohen Sure, will work on that.

macohen:
This has been released in https://github.com/awslabs/aws-athena-query-federation/releases/tag/v2025.34.1.

PR #2773, opened 2025-05-14
Support Snowflake OAuth authentication for Athena Snowflake connector

Trianz-Akshay:
*Issue #, if available:*
https://github.com/awslabs/aws-athena-query-federation/issues/2744

*Description of changes:*
This PR introduces support for OAuth-based authentication for the Snowflake connector by securely fetching and storing  access tokens and refresh token using AWS Secrets Manager.
Changes:

- OAuth Token Handling:
Implements logic to fetch the initial access token using the authorization_code grant.
Stores full token response (including refresh_token) in AWS secrets manager on first auth.
On subsequent requests, checks token expiry based on fetched_at and expires_in.
Uses the refresh_token to obtain a new access token when the current one expires.
Updates only access_token, expires_in, and fetched_at in the existing token JSON, preserving the original refresh_token.

- Integration with Snowflake JDBC:
Injects the OAuth access_token as the JDBC password property.
Sets the authenticator property to oauth.

- Secrets Manager Integration:
Reads OAuth client configuration (client ID, secret, redirect URI, etc.) and auth_code from AWS Secrets Manager.

- S3 Token Storage:
Reads/writes token JSON from/to AWS Secrets Manager.

[Snowflake_Oauth.docx](https://github.com/user-attachments/files/20220442/Snowflake_Oauth.docx)


burhan94:
I'm not really in favor of having to store the oath token in S3. Can we create a background thread that manages the credential refresh and updates the oath token directly in secrets manager?

Otherwise, we will need to encrypt the token with a customer-provided KMS key.

Trianz-Akshay:
> I'm not really in favor of having to store the oath token in S3. Can we create a background thread that manages the credential refresh and updates the oath token directly in secrets manager?
> 
> Otherwise, we will need to encrypt the token with a customer-provided KMS key.

@burhan94 Following your recommendation, we have implemented the change to securely store the access token and refresh token in the Secret Manager rather than in Amazon S3.

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeOAuthJdbcConnectionFactory.java
```diff
@@ -0,0 +1,262 @@
...
+        else {
+            long expiresIn = Long.parseLong(oauthConfig.get(EXPIRES_IN));
+            long fetchedAt = Long.parseLong(oauthConfig.getOrDefault(FETCHED_AT, String.valueOf(0L)));
+            long now = System.currentTimeMillis() / 1000;
+
+            if ((now - fetchedAt) < expiresIn - 60) {
+                LOGGER.debug("Access token still valid.");
+            }
+            else {
+                LOGGER.debug("Access token expired. Using refresh_token...");
```
burhan94: If the token expires while the query is running or when we pull data from the driver, does anything fail?

Trianz-Akshay: It should not cause any failure in the middle of the request.
Token validation is performed only once at the time of establishing the JDBC connection.
As long as the connection remains alive, the driver continues to use the same session and does not revalidate the token for ongoing queries or result processing.
So, in normal conditions, if the connection remains open throughout the query execution, there should not be any error due to token expiry.

reference - https://community.snowflake.com/s/article/FAQs-Snowflake-OAuth

burhan94 (approved):

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeOAuthJdbcConnectionFactory.java
```diff
@@ -0,0 +1,262 @@
...
+    @VisibleForTesting
+    public SnowflakeOAuthJdbcConnectionFactory(DatabaseConnectionConfig databaseConnectionConfig,
+                                               Map<String, String> properties,
+                                               DatabaseConnectionInfo databaseConnectionInfo, Map<String, String> configOptions,
+                                               SecretsManagerClient secretsClient)
+    {
+        super(databaseConnectionConfig, properties, databaseConnectionInfo);
+        this.databaseConnectionInfo = Validate.notNull(databaseConnectionInfo, "databaseConnectionInfo must not be null");
+        this.databaseConnectionConfig = Validate.notNull(databaseConnectionConfig, "databaseConnectionConfig must not be null");
+        this.configOptions = configOptions;
```
chngpe: This part of code is duplicate compare to above, can we modularized it?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeOAuthJdbcConnectionFactory.java
```diff
@@ -0,0 +1,262 @@
...
+        try {
+            final String derivedJdbcString;
+            GetSecretValueRequest getSecretValueRequest = GetSecretValueRequest.builder()
+                    .secretId(this.oauthSecretName)
+                    .build();
+
+            GetSecretValueResponse secretValue = this.secretsClient.getSecretValue(getSecretValueRequest);
+            Map<String, String> oauthConfig = new ObjectMapper().readValue(secretValue.secretString(), Map.class);
+            if (oauthConfig.containsKey(AUTH_CODE) && !oauthConfig.get(AUTH_CODE).isEmpty()) {
+                if (credentialsProvider != null) {
```
chngpe: what is credential provider is null?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeOAuthJdbcConnectionFactory.java
Trianz-Akshay: Thanks @chngpe for pointing out. Used constructor chaining via `this()` to eliminate duplication and delegate initialization between constructors.

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeOAuthJdbcConnectionFactory.java
Trianz-Akshay: The credential provider has been removed because the secret values (such as username) are now directly available in the OAuthConfig, and providing a secret is mandatory as per the recent changes. Therefore, the username is retrieved directly from the OAuthConfig.

For non-OAuth configurations, the rest of the flow remains unchanged and continues to follow the standard behavior from GenericJdbcConnectionFactory.

chngpe (approved):

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeOAuthJdbcConnectionFactory.java
```diff
@@ -0,0 +1,251 @@
...
+        this.jdbcProperties = new Properties();
+        if (properties != null) {
+            this.jdbcProperties.putAll(properties);
+        }
+        this.secretsClient = secretsClient;
+        this.oauthSecretName = Validate.notNull(databaseConnectionConfig.getSecret(), "Missing required property: secret name");
+    }
+
+    @Override
+    public Connection getConnection(final CredentialsProvider credentialsProvider)
```
chngpe: Several comments regarding to this:
1. This is Connection factory, but majority of code here are credential related. Can we expands credential provider like this one ? https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/DefaultCredentialsProvider.java. We can have a SnowFlakeCredentiasProvider.java 

2. With #1 changes, this should let us continue to use GenericJDBCConnectorFactory: https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/connection/GenericJdbcConnectionFactory.java#L86-L87.  But instead of of calling `jdbcProperties.put("user", credentialsProvider.getCredential().getUser());` and `jdbcProperties.put("password", credentialsProvider.getCredential().getPassword());` we should introduce a method in CredentialsProvider to return Map<String, String> and which we will just do jdbcProperteis.putAll(CredentialsProvider.getmap()). (see point 3 for detail)

Also, given this is overridden connection factory, we also get rid of the connection pool :( which is we should preserved the behavior (per suggestion, if we change this to credential provider then we don't have to worry about this)

3. Credential provider interface doesn't have an interface of getCredentialMap, https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/CredentialsProvider.java We will need to add that, and from https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/connection/GenericJdbcConnectionFactory.java#L86-L87.  just call that method.

4. Can we reuse https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/CachableSecretsManager.java? It is in based handler example: https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java#L122

Given we need to write please update interface correctly. Please make sure we are refreshing from cache correctly given it is Oauth

Trianz-Akshay: We have updated the code as per your review comments. The refactoring separates OAuth credential management from the connection factory and introduces a dedicated credentials provider for improved modularity and clarity.
 
This approach was not considered earlier, as it required changes to both the Athena SDK and the JDBC module.
Kindly review the changes and let us know if there's anything we might have missed.

chngpe (approved):

Issue #2789, opened 2025-05-23
[BUG] poorly implemented query validation with connector Timestream


Moreno-Gentili:
The current implementation of `customConnectorVerifications` in the Timestream connector is problematic, to say the least. In my case it makes the whole connector unusable.

This is the part I'm referring to:
https://github.com/awslabs/aws-athena-query-federation/blob/5df26c2a30a9a093454eb716a6f8726753f535f7/athena-timestream/src/main/java/com/amazonaws/athena/connectors/timestream/qpt/TimestreamQueryPassthrough.java#L73-L91

As you can see, queries must start with `SELECT`, which is blocking perfectly valid CTE queries that start with `WITH`.

Also, the `disallowedKeywords` blindly blocks perfectly valid uses of the words `update` and `create`, like when they're used in field names such as `last_update` or in the Timestream aggregation function [CREATE_TIME_SERIES](https://docs.aws.amazon.com/timestream/latest/developerguide/timeseries-specific-constructs.views.html#timeseries-specific-constructs.views.CREATE_TIME_SERIES).

Please add a way to bypass that validation completely, before it gets a proper implementation.

Trianz-Akshay:
Hi @macohen , could you please assign this issue to me? I’d be happy to take this up and provide updates. Thanks!

PR #2790, opened 2025-05-25
Timestream connector: removes the check on disallowed words and allows queries to start with `SELECT` and `WITH`

Moreno-Gentili:
*Issue #2789*

*Removes the check on disallowed words and allows queries to start with `SELECT` and `WITH`*

### DIFF
--- a/athena-clickhouse/pom.xml
+++ b/athena-clickhouse/pom.xml
@@ -22,7 +22,7 @@
         <dependency>
             <groupId>com.clickhouse</groupId>
             <artifactId>clickhouse-jdbc</artifactId>
-            <version>0.8.5</version>
+            <version>0.8.6</version>
             <classifier>all</classifier>
         </dependency>
         <dependency>


athena-timestream/src/main/java/com/amazonaws/athena/connectors/timestream/qpt/TimestreamQueryPassthrough.java
```diff
@@ -76,18 +76,8 @@ public void customConnectorVerifications(Map<String, String> engineQptArguments)
...
-        if (!upperCaseStatement.startsWith("SELECT")) {
-            throw new UnsupportedOperationException("Statement does not start with SELECT.");
-        }
-
-        // List of disallowed keywords
-        Set<String> disallowedKeywords = ImmutableSet.of("INSERT", "UPDATE", "DELETE", "CREATE", "DROP", "ALTER");
-
-        // Check if the statement contains any disallowed keywords
-        for (String keyword : disallowedKeywords) {
-            if (upperCaseStatement.contains(keyword)) {
```
fal-bharadwaj, line 88: Instead of completely deleting the disallowed words logic, it will be a better idea to check for them using a word boundary regex pattern match. This will allow for field names like "last_update" to be allowed but still catch any cases where these words are used as keywords.

burhan94:
Closing PR due to no response. Please re-open if you find time to address the comments.

PR #2821, opened 2025-06-09
Query federation SDK changes for Managed Connector Approach

DilipReddyGaddam:
*Issue #, if available:*
N/A

*Description of changes:*
As we plan support managed connector feature we have to made below SDK changes 

1. Change to handle credentials fetch from Connector request made to lambda instead of using lambda execution role. We have created a KMS decryption provider class to decrypt tokens present in the request and use them in calls the made from connector.

2. Change the Spill location generation logic, If query id is already present in the prefix we don't append it again.

3. Added QueryPlan object to constraints object, This Query plan will have serialized Substrait plan of the query.

Tested the changes for backwards compatibility by deploying athena ddb connector and running athena queries.

We have tested the changes from MaxDome updating the connector lambda and making sure that metadata requests from GDC are successful and query editor requests are also successful.

Some logs from lambda

```
2025-07-02T13:25:59.943-07:00
	
2025-07-02 20:25:59 35155224-a073-462d-aad7-7dffa6299b5f DEBUG VersionedObjectMapperFactory:64 - Athena SDK SerDe Version: 6
	
2025-07-02T13:25:59.943-07:00
	
2025-07-02 20:25:59 35155224-a073-462d-aad7-7dffa6299b5f DEBUG CompositeHandler:112 - Client's SerDe mis-matched with connector version:, attempt with lower version: '5'
	
2025-07-02T13:25:59.943-07:00
	
2025-07-02 20:25:59 35155224-a073-462d-aad7-7dffa6299b5f DEBUG VersionedObjectMapperFactory:64 - Athena SDK SerDe Version: 5
	
2025-07-02T13:25:59.950-07:00
	
2025-07-02 20:25:59 35155224-a073-462d-aad7-7dffa6299b5f INFO CompositeHandler:120 - Parsing request with resolvedSerDeVersion: '5', connector SerDeVersion: '6'
	
2025-07-02T13:25:59.958-07:00
	
2025-07-02 20:25:59 35155224-a073-462d-aad7-7dffa6299b5f INFO MetadataHandler:285 - doHandleRequest: request[GetTableLayoutRequest{queryId=65d8d25b-28bd-4aaa-81ed-741a677b77ee, tableName=TableName{schemaName=default, tableName=ddblower}, constraints=Constraints{summary={}expression=[]orderByClause=[]limit=-1queryPassthroughArguments={}queryPlan=null}, schema=Schema<count: Decimal(38, 9, 128), date: Utf8, name: Utf8, id: Utf8>, partitionCols=[]}] 
```


athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/resolver/DynamoDBTableResolver.java
```diff
@@ -69,17 +70,20 @@ public DynamoDBTableResolver(ThrottlingInvoker invoker, DynamoDbClient ddbClient
     }
 
     /**
-     * Fetches the list of tables from DynamoDB via paginated ListTables calls
+     * Fetches the list of tables from DynamoDB vis paginated ListTables calls
```
ZhitongYan: small typo

ZhitongYan (commented):
1. Is it possible to have a default `null` value for all addition of AwsRequestOverrideConfiguration parameter, so that we can keep backwards compatibility?

athena-federation-sdk/pom.xml
```diff
@@ -133,6 +133,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-encryption-sdk-java</artifactId>
```
ZhitongYan, line 144: Looks like its SDK v1 dependency, is there a v2 available?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
```diff
@@ -85,6 +86,31 @@ public Constraints(@JsonProperty("summary") Map<String, ValueSet> summary,
...
+    }
+
+    /**
+     *
+     * @param summary
+     * @param expression
+     * @param orderByClause
+     * @param limit
+     * @param queryPassthroughArguments
+     */
```
ZhitongYan, line 65: missing queryPlan doc

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
```diff
@@ -52,6 +52,7 @@ public class Constraints
     private final List<OrderByField> orderByClause;
     private long limit;
     private Map<String, String> queryPassthroughArguments;
+    private final QueryPlan queryPlan;
```
ZhitongYan, line 55: is this serializable? [Got answered below]

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/S3SpillLocation.java
```diff
@@ -169,7 +170,14 @@ public Builder withSplitId(String splitId)
 
         public S3SpillLocation build()
         {
-            String key = prefix + SEPARATOR + queryId + SEPARATOR + splitId;
+            String bucketPrefix;
+            if (StringUtils.isBlank(queryId)) {
```
ZhitongYan, line 174: under which scenarios will queryId be blank? its not the case before? seems like a bigger change here

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/FederationRequestHandler.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.Map;
+import java.util.Objects;
+
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.FAS_TOKEN;
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.SERVICE_KMS;
+
+public interface FederationRequestHandler extends RequestStreamHandler
+{
+    default AwsCredentials getSessionCredentials(String kmsKeyId,
```
ZhitongYan, line 39: is this method used somewhere else? seems we can just pack it in `getRequestOverrideConfig`

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/FederationRequestHandler.java
```diff
@@ -0,0 +1,88 @@
...
+    {
+        if (Objects.nonNull(awsRequestOverrideConfiguration) &&
+                awsRequestOverrideConfiguration.credentialsProvider().isPresent()) {
+            AwsCredentialsProvider awsCredentialsProvider = awsRequestOverrideConfiguration.credentialsProvider().get();
+            return S3Client.builder()
+                    .credentialsProvider(awsCredentialsProvider)
+                    .build();
+        }
+        else {
+            return S3Client.create();
```
ZhitongYan: 1. is this using default credentials provider?
2. thinking Athena SDK will either: use aws clients provided directly from caller, OR use override config to create one with FAS. So seems no such cases for `S3Client.create()`?
3. Also can these clients creation be singleton to avoid leak?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -126,7 +130,8 @@ public abstract class MetadataHandler
     private final String spillBucket;
     private final String spillPrefix;
     private final String sourceType;
-    private final SpillLocationVerifier verifier;
+    private SpillLocationVerifier verifier;
+    private final KmsEncryptionProvider kmsEncryptionProvider;
```
ZhitongYan, line 138: why not also move this into `interface FederationRequestHandler` and just have `public AwsRequestOverrideConfiguration getRequestOverrideConfig(Map<String, String> configOptions)` there

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -179,6 +185,7 @@ public MetadataHandler(
         this.spillPrefix = spillPrefix;
         this.verifier = new SpillLocationVerifier(S3Client.create());
         this.athenaInvoker = ThrottlingInvoker.newDefaultBuilder(ATHENA_EXCEPTION_FILTER, configOptions).build();
+        this.kmsEncryptionProvider = new KmsEncryptionProvider(KmsClient.create());
```
ZhitongYan, line 194: KMS client does not use FAS?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -293,7 +301,12 @@ protected final void doHandleRequest(BlockAllocator allocator,
                 }
                 return;
             case GET_SPLITS:
-                verifier.checkBucketAuthZ(spillBucket);
+                FederatedIdentity federatedIdentity = req.getIdentity();
+                Map<String, String> connectorRequestOptions = federatedIdentity.getConfigOptions();
+                if (connectorRequestOptions != null && connectorRequestOptions.get(FAS_TOKEN) != null) {
+                    AwsRequestOverrideConfiguration awsRequestOverrideConfiguration = getRequestOverrideConfig(connectorRequestOptions);
+                    verifier = new SpillLocationVerifier(getS3Client(awsRequestOverrideConfiguration));
+                }
```
ZhitongYan, line 323: missing Authorization check? `verifier.checkBucketAuthZ(spillBucket);`

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/RecordHandler.java
```diff
@@ -192,12 +199,16 @@ public RecordResponse doReadRecords(BlockAllocator allocator, ReadRecordsRequest
             throws Exception
     {
         logger.info("doReadRecords: {}:{}", request.getSchema(), request.getSplit().getSpillLocation());
+        FederatedIdentity federatedIdentity = request.getIdentity();
+        AwsRequestOverrideConfiguration overrideConfig = getRequestOverrideConfig(federatedIdentity.getConfigOptions());
         SpillConfig spillConfig = getSpillConfig(request);
+        AthenaClient athenaClient = getAthenaClient(overrideConfig);
```
ZhitongYan: every get records call will create brand new clients?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/RecordHandler.java
```diff
@@ -192,12 +199,16 @@ public RecordResponse doReadRecords(BlockAllocator allocator, ReadRecordsRequest
...
+        S3Client s3Client = getS3Client(overrideConfig);
         try (ConstraintEvaluator evaluator = new ConstraintEvaluator(allocator,
                 request.getSchema(),
                 request.getConstraints());
-                S3BlockSpiller spiller = new S3BlockSpiller(amazonS3, spillConfig, allocator, request.getSchema(), evaluator, configOptions);
-                QueryStatusChecker queryStatusChecker = new QueryStatusChecker(athena, athenaInvoker, request.getQueryId())
+                S3BlockSpiller spiller = new S3BlockSpiller(s3Client, spillConfig, allocator, request.getSchema(), evaluator, configOptions);
+                QueryStatusChecker queryStatusChecker = new QueryStatusChecker(athenaClient, athenaInvoker, request.getQueryId())
```
ZhitongYan, line 211: should these clients fall back to use original amazons3 / athena from constructor when override config is not provided?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
ZhitongYan: Did not see usage for this in this CR, maybe can remove from this change and add it together with related changes with context?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/KmsEncryptionProvider.java
```diff
@@ -0,0 +1,121 @@
...
+        this.kmsClient = kmsClient;
+        this.cache = CacheBuilder.newBuilder()
+                .maximumSize(MAX_CACHE_SIZE)
+                .expireAfterWrite(CACHE_DURATION, TimeUnit.HOURS)
+                .build();
+        awsCrypto = AwsCrypto.builder().build();
+    }
+
+    @VisibleForTesting
+    public KmsEncryptionProvider(final KmsClient kmsClient,
```
ZhitongYan: protected might also work just for testing purpose without exposing the signature

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/KmsEncryptionProvider.java
```diff
@@ -0,0 +1,121 @@
...
+        return new String(awsCrypto.decryptData(cachingCmm, decode).getResult());
+    }
+
+    private CryptoMaterialsManager getCachingCmm(String kmsKeyId)
+    {
+        CryptoMaterialsManager cachingCmm = cache.getIfPresent(kmsKeyId);
+        if (cachingCmm == null) {
+            CryptoMaterialsCache cache = new LocalCryptoMaterialsCache(CACHE_SIZE);
+            MasterKeyProvider<KmsMasterKey> keyProvider = KmsMasterKeyProvider.builder()
+                    .customRegionalClientSupplier((region) -> kmsClient)
```
ZhitongYan, line 109: region agnostic?

athena-federation-sdk/pom.xml
DilipReddyGaddam: This is library which is internally using all V2 clients. Please refer the read me in this https://github.com/aws/aws-encryption-sdk-java

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
DilipReddyGaddam: This PR includes all the changes required in Connector SDK and also some changes in DDB connector. Will send out a new PR which uses this newly added objects.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/spill/S3SpillLocation.java
DilipReddyGaddam: We are expecting the prefix object will have queryid and we don't want to add this again in the prefix. We expect callers of our service to build the prefix with the queryid before making a request to us.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/FederationRequestHandler.java
DilipReddyGaddam: Yes, If the request OverrideConfig is null we use default credentials. This is existing  functionality in athena connectors they use default lambda execution role credentials to create these clients. 

We will not be able to create a singleton credentials as we want to override credentials based on the request.

I have updated to use only one client across invocations instead of creating a new one.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
DilipReddyGaddam: This will be service owned KMS key, We don't want to use FAS tokens for this.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
DilipReddyGaddam: Thank you. Added it back

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/RecordHandler.java
DilipReddyGaddam: Yes. That is expected. We want to use FAS credentials present in the request to create these clients.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/RecordHandler.java
DilipReddyGaddam: Yes added this in the logic of  generating new clients.

athena-federation-sdk/pom.xml
```diff
@@ -264,6 +269,12 @@
             <version>${assertj.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
```
burhan94: no lombok please. We don't use lombok anywhere in this package and im generally not a fan of lombok either.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
```diff
@@ -52,6 +52,7 @@ public class Constraints
...
 
     @Deprecated
     public Constraints(@JsonProperty("summary") Map<String, ValueSet> summary,
```
burhan94: This method is deprecated. Please delete this method, create a new method (already done), and deprecate the existing one.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
```diff
@@ -73,7 +75,6 @@ public Constraints(@JsonProperty("summary") Map<String, ValueSet> summary,
      * @param limit
      * @param queryPassthroughArguments
      */
-    @JsonCreator
     public Constraints(@JsonProperty("summary") Map<String, ValueSet> summary,
```
burhan94: @deprecated

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/FASToken.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package com.amazonaws.athena.connector.lambda.security;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
```
burhan94: remove lombok

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/FederatedIdentity.java
```diff
@@ -37,8 +38,8 @@ public class FederatedIdentity
     private final String account;
     private final Map<String, String> principalTags;
     private final List<String> iamGroups;
+    private final Map<String, String> configOptions;
 
-    @JsonCreator
```
burhan94, line 41: mark as deprecated and delete.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/FederatedIdentity.java
```diff
@@ -48,6 +49,21 @@ public FederatedIdentity(@JsonProperty("arn") String arn,
...
         this.iamGroups = iamGroups;
+        this.configOptions = new HashMap<>();
+    }
+
+    @JsonCreator
+    public FederatedIdentity(@JsonProperty("arn") String arn,
+                             @JsonProperty("account") String account,
+                             @JsonProperty("principalTags") Map<String, String> principalTags,
+                             @JsonProperty("iamGroups") List<String> iamGroups,
+                             @JsonProperty("configOptions") Map<String, String> configOptions)
```
burhan94, line 47: you'll need to make the corresponding serialization changes as well.

burhan94 (changes requested):

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
DilipReddyGaddam: There are so many instances which are using this constructor. Do you want me to update all those ? 

Yes I will mark the existing one as deprecated.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/FASToken.java
DilipReddyGaddam: sure will update this, and remove lombok

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/FederatedIdentity.java
DilipReddyGaddam: Sure will mark this as deprecated

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/FederatedIdentity.java
DilipReddyGaddam: Thanks for pointing this. I will make the changes to serializer class too

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v5/ConstraintsSerDeV5.java
```diff
@@ -150,7 +159,13 @@ public Constraints doDeserialize(JsonParser jparser, DeserializationContext ctxt
                 queryPassthroughArguments.put(jparser.getCurrentName(), jparser.getValueAsString());
             }
 
-            Constraints constraints = new Constraints(summaryMap.build(), federationExpression.build(), orderByClauseBuilder.build(), limit, queryPassthroughArguments);
+            QueryPlan queryPlan = null;
+            assertFieldName(jparser, QUERY_PLAN);
+            if (jparser.nextToken() != JsonToken.VALUE_NULL) {
```
burhan94: You need to create a ConstraintsSerDeV6 with these changes. Otherwise we'll run into issues when we serialize with a V5 client and the connector is using V6

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/PingRequestSerDe.java
```diff
@@ -84,8 +84,10 @@ public FederationRequest deserialize(JsonParser jparser, DeserializationContext
                 throws IOException
         {
             validateObjectStart(jparser.getCurrentToken());
-            return doDeserialize(jparser, ctxt);
-            // do not validate object end to allow forwards compatibility
+            FederationRequest result = doDeserialize(jparser, ctxt);
```
burhan94: Why did PingRequest need to change?

burhan94:
Can you deploy a connector with these changes, and test it with Athena? We need to make sure that these changes are backwards compatible.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/PingRequestSerDe.java
DilipReddyGaddam: This is was changed by Q while fixing some unit tests. Will remove this.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v6/ObjectMapperFactoryV6.java
```diff
@@ -0,0 +1,385 @@
...
+            disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+        }
+    }
+
+    public static ObjectMapper create(BlockAllocator allocator)
+    {
+        return new StrictObjectMapper(allocator);
+    }
+
+    private static FederationRequestSerDeV5.Serializer createRequestSerializer()
```
burhan94: I know its not required, but I think we should make a FederationRequestSerDeV6 to be consistent. This way the object mapper version can directly be associated with a FederationRequest version.

burhan94 (commented):
You need to bump the default SerDeVersion here: https://github.com/awslabs/aws-athena-query-federation/blob/f3648629fda99b5b883cccd7eaf0c6943f0abad9/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/SerDeVersion.java

Once you do this, you need to retest everything and make sure there are no failures. Please also make sure you test with GDC integration.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v6/ObjectMapperFactoryV6.java
DilipReddyGaddam: sure will add FederationRequestSerDeV6 and update this.

DilipReddyGaddam:
> You need to bump the default SerDeVersion here: https://github.com/awslabs/aws-athena-query-federation/blob/f3648629fda99b5b883cccd7eaf0c6943f0abad9/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/SerDeVersion.java
> 
> Once you do this, you need to retest everything and make sure there are no failures. Please also make sure you test with GDC integration.

I was planning to upgrade once we start integrating with connector. I will update this and test again with athena query editor and GDC too

burhan94 (approved):

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v2/ConstraintsSerDe.java
```diff
@@ -91,7 +91,8 @@ public Constraints doDeserialize(JsonParser jparser, DeserializationContext ctxt
             return new Constraints(summaryMap.build(),
                     Collections.emptyList(),
                     Collections.emptyList(),
-                    NO_LIMIT_VALUE);
```
AbdulR3hman, line 94: I don't think we should be touching ANY v2/v3/v4 SerDes; if this goes into the engine; it will break any customers who are using those older serdes. Hence why we have to add a new version. If this object needs to be changed; it needs to be added to v6 (the new version) and keep the old version as is.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v4/ConstraintsSerDeV4.java
```diff
@@ -140,7 +141,7 @@ public Constraints doDeserialize(JsonParser jparser, DeserializationContext ctxt
 
             long limit = getNextLongField(jparser, LIMIT_FIELD);
 
-            return new Constraints(summaryMap.build(), federationExpression.build(), orderByClauseBuilder.build(), limit);
+            return new Constraints(summaryMap.build(), federationExpression.build(), orderByClauseBuilder.build(), limit, Collections.emptyMap(), null);
```
AbdulR3hman, line 144: same above; I don't think we should be making any changes to existing SerDes!

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v5/ConstraintsSerDeV5.java
```diff
@@ -150,8 +150,8 @@ public Constraints doDeserialize(JsonParser jparser, DeserializationContext ctxt
                 queryPassthroughArguments.put(jparser.getCurrentName(), jparser.getValueAsString());
             }
 
-            Constraints constraints = new Constraints(summaryMap.build(), federationExpression.build(), orderByClauseBuilder.build(), limit, queryPassthroughArguments);
-            return constraints;
+            return new Constraints(summaryMap.build(), federationExpression.build(), orderByClauseBuilder.build(),
+                    limit, queryPassthroughArguments, null);
```
AbdulR3hman, line 154: same above; I don't think we should be making any changes to existing SerDes!

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/v2/ConstraintsSerDe.java
AbdulR3hman: I believe the way to do this; is to introduce a new constructor with the extra parameters; and set the previous constructor as `@Deprecated`

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
```diff
@@ -52,18 +52,7 @@ public class Constraints
     private final List<OrderByField> orderByClause;
     private long limit;
     private Map<String, String> queryPassthroughArguments;
-
-    @Deprecated
-    public Constraints(@JsonProperty("summary") Map<String, ValueSet> summary,
```
AbdulR3hman, line 57: I believe this needs to stay if we are going to not change v2 SerDe

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/Constraints.java
```diff
@@ -72,19 +61,22 @@ public Constraints(@JsonProperty("summary") Map<String, ValueSet> summary,
      * @param orderByClause
      * @param limit
      * @param queryPassthroughArguments
+     * @param queryPlan
      */
     @JsonCreator
     public Constraints(@JsonProperty("summary") Map<String, ValueSet> summary,
```
AbdulR3hman, line 67: same for here; we need to introduce a new constructor with the new parameter; and assign this as deprecated.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/serde/FederatedIdentitySerDe.java
```diff
@@ -104,7 +104,25 @@ public FederatedIdentity doDeserialize(JsonParser jparser, DeserializationContex
             Map<String, String> principalTags = getNextStringMap(jparser, TAGS_FIELD);
             List<String> groups = getNextStringArray(jparser, GROUPS_FIELD);
 
-            return new FederatedIdentity(arn, account, principalTags, groups);
+            // Handle configOptions field for backwards compatibility
```
AbdulR3hman, line 107: this can't change. SerDe needs to be backward compatible; We need to introduce V6 version into the new SerDe. Every previous SerDe has to continue to work.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/FederatedIdentity.java
```diff
@@ -37,17 +37,20 @@ public class FederatedIdentity
     private final String account;
     private final Map<String, String> principalTags;
     private final List<String> iamGroups;
+    private final Map<String, String> configOptions;
 
     @JsonCreator
     public FederatedIdentity(@JsonProperty("arn") String arn,
```
pearlswan, line 43: @burhan94  @DilipReddyGaddam do you know why none of the request models compare the federated identity object in the equals method? 

https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/GetSplitsRequest.java#L222

burhan94:
one last comment:

Does the DynamoDBRecordHandler not need changes? 

`private Iterator<Map<String, AttributeValue>> getIterator` has two calls to DDB, but we're not overriding the request with the new credentials.

DilipReddyGaddam:
> one last comment:
> 
> Does the DynamoDBRecordHandler not need changes?
> 
> `private Iterator<Map<String, AttributeValue>> getIterator` has two calls to DDB, but we're not overriding the request with the new credentials.

I have made changes to that and they are nit included in this PR. I will send out a new PR after this one and that will have DDB record handler changes.

burhan94 (approved):

AbdulR3hman (approved):
talked to @burhan94 and confirmed that my concerns regarding the SerDe are not warranted; and these changes won't break backward compatibility. Still think we need to deploy this as part of the engine to be 100% sure;

PR #2822, opened 2025-06-11
added unit tests for athena-postgresql.

ritiktrianz:
*Issue #, if available:*

*Description of changes:*
Added unit tests for athena-postgresql connector.

### DIFF
--- a/athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlEnvironmentPropertiesTest.java
+++ b/athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlEnvironmentPropertiesTest.java
@@ -1,0 +1,57 @@
+/*-
+ * #%L
+ * athena-postgresql
+ * %%
+ * Copyright (C) 2019 - 2025 Amazon Web Services
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package com.amazonaws.athena.connectors.postgresql;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.DATABASE;
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.DEFAULT;
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.HOST;
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.PORT;
+import static com.amazonaws.athena.connector.lambda.connection.EnvironmentConstants.SECRET_NAME;
+import static org.junit.Assert.assertEquals;
+
+public class PostGreSqlEnvironmentPropertiesTest
+{
+    Map<String, String> connectionProperties;
+    PostGreSqlEnvironmentProperties postGreSqlEnvironmentProperties;
+
+    @Before
+    public void setUp() {
+        connectionProperties = new HashMap<>();
+        connectionProperties.put(HOST, "postgres-endpoint");
+        connectionProperties.put(DATABASE, "testdb");
+        connectionProperties.put(SECRET_NAME, "postgres-secret");
+        connectionProperties.put(PORT, "5000");
+        postGreSqlEnvironmentProperties = new PostGreSqlEnvironmentProperties();
+    }
+
+    @Test
+    public void postGreSqlConnectionPropertiesTest() {
+        Map<String, String> postGreSqlConnectionProperties = postGreSqlEnvironmentProperties.connectionPropertiesToEnvironment(connectionProperties);
+
+        String expectedConnectionString = "postgres://jdbc:postgresql://postgres-endpoint:5000/testdb?${postgres-secret}";
+        assertEquals(expectedConnectionString, postGreSqlConnectionProperties.get(DEFAULT));
+    }
+}
--- a/athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
+++ b/athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
@@ -31,6 +31,8 @@ import com.amazonaws.athena.connector.lambda.metadata.GetSplitsRequest;
 import com.amazonaws.athena.connector.lambda.metadata.GetSplitsResponse;
 import com.amazonaws.athena.connector.lambda.metadata.GetTableLayoutRequest;
 import com.amazonaws.athena.connector.lambda.metadata.GetTableLayoutResponse;
+import com.amazonaws.athena.connector.lambda.metadata.GetDataSourceCapabilitiesResponse;
+import com.amazonaws.athena.connector.lambda.metadata.GetDataSourceCapabilitiesRequest;
 import com.amazonaws.athena.connector.lambda.metadata.GetTableRequest;
 import com.amazonaws.athena.connector.lambda.metadata.GetTableResponse;
 import com.amazonaws.athena.connector.lambda.metadata.ListTablesRequest;
@@ -40,12 +42,25 @@ import com.amazonaws.athena.connectors.jdbc.TestBase;
 import com.amazonaws.athena.connectors.jdbc.connection.DatabaseConnectionConfig;
 import com.amazonaws.athena.connectors.jdbc.connection.JdbcConnectionFactory;
 import com.amazonaws.athena.connector.credentials.CredentialsProvider;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.amazonaws.athena.connector.lambda.metadata.optimizations.DataSourceOptimizations;
+import com.amazonaws.athena.connector.lambda.metadata.optimizations.OptimizationSubType;
+import com.amazonaws.athena.connector.lambda.metadata.optimizations.pushdown.ComplexExpressionPushdownSubType;
+import com.amazonaws.athena.connector.lambda.metadata.optimizations.pushdown.FilterPushdownSubType;
+import com.amazonaws.athena.connector.lambda.metadata.optimizations.pushdown.HintsSubtype;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
+import static org.mockito.ArgumentMatchers.anyString;
+
 import org.apache.arrow.vector.types.DateUnit;
 import org.apache.arrow.vector.types.FloatingPointPrecision;
 import org.apache.arrow.vector.types.pojo.ArrowType;
 import org.apache.arrow.vector.types.pojo.Field;
 import org.apache.arrow.vector.types.pojo.Schema;
+import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
@@ -72,7 +87,6 @@ import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
-import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.nullable;
 import static org.mockito.Mockito.when;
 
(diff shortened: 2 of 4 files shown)


athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
```diff
@@ -210,6 +245,89 @@ public void doGetTableLayout()
         Mockito.verify(preparedStatement, Mockito.times(1)).setString(2, tableName.getTableName());
     }
 
+    @Test
+    public void testListTablesWithValidConnection() {
```
AbdulR3hman: throw exception here; try/catch is not needed. let the exception bubble.

athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
```diff
@@ -210,6 +245,89 @@ public void doGetTableLayout()
...
+            Mockito.verify(matViewStatement).executeQuery();
+        } catch (Exception e) {
+            fail("Unexpected exception occurred: " + e.getMessage());
+        }
+    }
+
+
+    @Test
+    public void testListTablesWithNoMaterializedViews() {
+        try {
```
AbdulR3hman: same here; please let the exception bubble up.

athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
```diff
@@ -210,6 +245,89 @@ public void doGetTableLayout()
...
+    public void testListTablesWithValidConnection() {
+        try {
+            String databaseName = "testSchema";
+            Connection jdbcConnection = Mockito.mock(Connection.class);
+            PreparedStatement tablesStatement = Mockito.mock(PreparedStatement.class);
+            PreparedStatement matViewStatement = Mockito.mock(PreparedStatement.class);
+            ResultSet tablesResultSet = Mockito.mock(ResultSet.class);
+            ResultSet matViewResultSet = Mockito.mock(ResultSet.class);
+
+            when(jdbcConnection.prepareStatement(anyString())).thenReturn(tablesStatement, matViewStatement);
```
AbdulR3hman: this could lead to vagueness when it failed; giving that we are saying any string in the preprared statement; when that's not entirely correct; See below; where we know exactly what the string to list tables is. 

https://github.com/awslabs/aws-athena-query-federation/blob/de87534712adb9befb4cd4e3b65ec428c6403cb4/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JDBCUtil.java#L155

athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
```diff
@@ -210,6 +245,89 @@ public void doGetTableLayout()
...
+
+            List<TableName> result = postGreSqlMetadataHandler.listTables(jdbcConnection, databaseName);
+
+            List<TableName> expected = ImmutableList.of(
+                    new TableName(databaseName, "table1"),
+                    new TableName(databaseName, "matview1")
+            );
+
+            assertEquals("Expected tables and materialized views", expected, result);
+            Mockito.verify(tablesStatement).executeQuery();
```
AbdulR3hman: I believe we should verify ahead of testing for equality. given that would be the product of a correct execution order that is confirmed by the verify.

athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
```diff
@@ -503,6 +620,166 @@ public void doGetTableMaterializedView()
...
+
+        List<OptimizationSubType> hints = capabilities.get(DATA_SOURCE_HINTS);
+        assertNotNull("Expected " + DATA_SOURCE_HINTS + " capability to be present", hints);
+        assertEquals(HINTS_SIZE, hints.size());
+        Assert.assertTrue(hints.stream().anyMatch(subType -> subType.getSubType().equals(NON_DEFAULT_COLLATE)));
+    }
+
+    @Test
+    public void testGetCharColumnsWithCharacterColumns() {
+        try {
```
AbdulR3hman: same here;

chngpe (approved):

athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
Jithendar12: Added constants TABLES_SQL and MATERIALIZED_VIEWS_SQL at class level with exact SQL queries used in the implementation.

athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
Jithendar12: Moved Mockito.verify() calls before assertions.

athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
Jithendar12: Removed try/catch blocks and added throws SQLException.

athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
Jithendar12: Removed try/catch blocks and added throws SQLException.

athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
Jithendar12: Thank you for reviewing @AbdulR3hman.
Removed try/catch blocks and added throws SQLException.

burhan94:
Can we consider using GenAI here to add all of the missing cases and add even more robust UT's? We should be able to get 30-50 new test cases here.

ritiktrianz:
Hi @burhan94 ,Thank you for the feedback.  We already tried GenAI using Claude Sonnet models, for unit test generation. It generated large volume of test cases covering multiple scenarios, such as edge cases, negative paths, and broader input variations. We felt that not all of these test cases may be relevant and may create unnecessary overhead. So, we removed a few tests and kept only the relevant ones.
 
However, now We have added the generated tests after reviewing them manually.

burhan94:
Can you add some UT's to cover the postgres SQL generation from the constraints object? Basically create a bunch of different scenarios that test every predicate type for postgres that we support today. I'd like to see all of the below:
1. Complex expressions
2. Top N
3. Limit
4. order by

ritiktrianz:
Hi @burhan94 , thank you for the review comment. We’ve added unit tests to cover all supported predicate types for Postgres.

athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandlerTest.java
```diff
@@ -205,9 +266,89 @@ public void doGetTableLayout()
...
+        when(matViewStatement.executeQuery()).thenReturn(emptyResultSet);
+
+        List<TableName> result = postGreSqlMetadataHandler.listTables(mockConn, TEST_SCHEMA);
+
+        List<TableName> expected = ImmutableList.of(new TableName(TEST_SCHEMA, "table1"));
+        assertEquals("Expected only tables", expected, result);
+    }
+
+    @Test
+    public void testListTablesWithDatabaseConnectionError() {
```
ejeffrli: can we not use try-catch here for this test?

ritiktrianz: Hi @ejeffrli , As suggested, we've removed the try-catch block from the test and replaced it with assertThrows for cleaner exception handling. Please review and let us know if you have any further feedback.

athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlEnvironmentPropertiesTest.java
```diff
@@ -0,0 +1,139 @@
...
+        connectionProperties.put(SECRET_NAME, "postgres-secret");
+        connectionProperties.put(PORT, DEFAULT_PORT);
+        postGreSqlEnvironmentProperties = new PostGreSqlEnvironmentProperties();
+    }
+
+    @Test
+    public void testBasicConnectionString() {
+        Map<String, String> postGreSqlConnectionProperties = postGreSqlEnvironmentProperties.connectionPropertiesToEnvironment(connectionProperties);
+
+        String expectedConnectionString = JDBC_URL_PREFIX + DEFAULT_HOST + ":" + DEFAULT_PORT + "/" + DEFAULT_DATABASE + "?" + SECRET_PLACEHOLDER;
```
ejeffrli: nit: For expected string assertions, can we write out the whole string explicitly? for example, it would be "postgres://jdbc:postgresql://..." This way developers can quickly verify differences.

ejeffrli: This is more readable
```
 assertEquals(
        "postgres://jdbc:postgresql://postgres-endpoint:5000/testdb?${postgres-secret}",
        postGreSqlConnectionProperties.get(DEFAULT)
    );
    ```

Comment (ritiktrianz) [reply athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlEnvironmentPropertiesTest.java]: > nit: For expected string assertions, can we write out the whole string explicitly? for example, it would be "postgres://jdbc:postgresql://..." This way developers can quickly verify differences.

Hi @ejeffrli , thank you for your suggestion. We’ve updated the assertions to improve readability. Please review and let us know if you have any further feedback.

File: athena-postgresql/src/test/java/com/amazonaws/athena/connectors/postgresql/PostgreSqlFederationExpressionParserTest.java [hunk 8]
```diff
@@ -0,0 +1,119 @@
...
+
+    @Test
+    public void testWriteArrayConstructorClauseWithSpecialCharacters() {
+        List<String> arguments = Arrays.asList("'Hello, World!'", "'It''s working'", "'Semi;Colon'");
+        String result = parser.writeArrayConstructorClause(new ArrowType.Utf8(), arguments);
+        assertEquals("'Hello, World!', 'It''s working', 'Semi;Colon'", result);
+    }
+
+    @Test
+    public void testWriteArrayConstructorClauseWithNull() {
```
ejeffrli: is this actually testing null? Seems like null is the same as any string here

ritiktrianz: > is this actually testing null? Seems like null is the same as any string here

Hi @ejeffrli , thank you. We’ve replaced the null string test with an actual null argument test. Kindly review.

ejeffrli: I'm not sure if this is the right change. I believe arguments should be SQL examples. In this case, would it be a SQL null? That means it would be a string like "null"

Let me know if there was another intent with this test.

ritiktrianz: > I'm not sure if this is the right change. I believe arguments should be SQL examples. In this case, would it be a SQL null? That means it would be a string like "null"
> 
> Let me know if there was another intent with this test.

Hi @ejeffrli , thanks for the review. We verified further and confirmed that there is a null check earlier in the flow, so this method would never receive a Java null. The original intent of this test was to cover a negative scenario where "null" is treated as a string value. Since similar coverage already exists, we have removed the SQL "null" test as well.
Please review and let us know if you have any feedback.

ejeffrli: which test is the similar coverage?

VenkatasivareddyTR: Hi @ejeffrli, these two tests provide the coverage we mentioned for String types, as "null" value also treated as String type only. Here are the reference links for each:
https://github.com/awslabs/aws-athena-query-federation/pull/2822/files#diff-fc042a21655b9e3353e4024d4c649d3426eefc86952b5a8e7ea19434844485cfR43

https://github.com/awslabs/aws-athena-query-federation/pull/2822/files#diff-fc042a21655b9e3353e4024d4c649d3426eefc86952b5a8e7ea19434844485cfR64

PR #2715, opened 2025-04-10
added unit tests in athena-federation-sdk.

VenkatasivareddyTR:
*Issue #, if available:*

*Description of changes:*
Added unit tests for athena-federation-sdk module.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2834, opened 2025-06-20
added unit tests for athena-cloudera-hive.

ritiktrianz:
*Issue #, if available:*

*Description of changes:*
Added unit tests for athena-cloudera-hive connector.

### DIFF
--- a/athena-clickhouse/src/main/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseConstants.java
+++ b/athena-clickhouse/src/main/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseConstants.java
@@ -39,6 +39,7 @@ public final class ClickHouseConstants
     // 9000; // Native-TCP
     // 9004; // MySQL
     static final Map<String, String> JDBC_PROPERTIES = Collections.singletonMap("databaseTerm", "SCHEMA");
+    static final int FETCH_SIZE = 1000;
 
     private ClickHouseConstants() {}
 }
--- a/athena-clickhouse/src/main/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseMetadataHandler.java
+++ b/athena-clickhouse/src/main/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseMetadataHandler.java
@@ -27,11 +27,14 @@ import com.amazonaws.athena.connector.lambda.data.SchemaBuilder;
 import com.amazonaws.athena.connector.lambda.domain.Split;
 import com.amazonaws.athena.connector.lambda.domain.TableName;
 import com.amazonaws.athena.connector.lambda.domain.spill.SpillLocation;
+import com.amazonaws.athena.connector.lambda.metadata.GetDataSourceCapabilitiesRequest;
+import com.amazonaws.athena.connector.lambda.metadata.GetDataSourceCapabilitiesResponse;
 import com.amazonaws.athena.connector.lambda.metadata.GetSplitsRequest;
 import com.amazonaws.athena.connector.lambda.metadata.GetSplitsResponse;
 import com.amazonaws.athena.connector.lambda.metadata.GetTableLayoutRequest;
 import com.amazonaws.athena.connector.lambda.metadata.ListTablesRequest;
 import com.amazonaws.athena.connector.lambda.metadata.ListTablesResponse;
+import com.amazonaws.athena.connector.lambda.metadata.optimizations.OptimizationSubType;
 import com.amazonaws.athena.connectors.clickhouse.resolver.ClickhouseJDBCCaseResolver;
 import com.amazonaws.athena.connectors.jdbc.connection.DatabaseConnectionConfig;
 import com.amazonaws.athena.connectors.jdbc.connection.DatabaseConnectionInfo;
@@ -40,6 +43,7 @@ import com.amazonaws.athena.connectors.jdbc.connection.JdbcConnectionFactory;
 import com.amazonaws.athena.connectors.jdbc.manager.JDBCUtil;
 import com.amazonaws.athena.connectors.mysql.MySqlMetadataHandler;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import org.apache.arrow.vector.complex.reader.FieldReader;
 import org.apache.arrow.vector.types.Types;
@@ -109,6 +113,15 @@ public class ClickHouseMetadataHandler
     }
 
     @Override
+    public GetDataSourceCapabilitiesResponse doGetDataSourceCapabilities(BlockAllocator allocator, GetDataSourceCapabilitiesRequest request)
+    {
+        ImmutableMap.Builder<String, List<OptimizationSubType>> capabilities = ImmutableMap.builder();
+
+        jdbcQueryPassthrough.addQueryPassthroughCapabilityIfEnabled(capabilities, configOptions);
+        return new GetDataSourceCapabilitiesResponse(request.getCatalogName(), capabilities.build());
+    }
+
+    @Override
     public Schema getPartitionSchema(final String catalogName)
     {
         SchemaBuilder schemaBuilder = SchemaBuilder.newBuilder()
@@ -135,6 +148,11 @@ public class ClickHouseMetadataHandler
             final BlockAllocator blockAllocator, final GetSplitsRequest getSplitsRequest)
     {
         LOGGER.debug("{}: Catalog {}, table {}", getSplitsRequest.getQueryId(), getSplitsRequest.getTableName().getSchemaName(), getSplitsRequest.getTableName().getTableName());
+        if (getSplitsRequest.getConstraints().isQueryPassThrough()) {
+            LOGGER.info("QPT Split Requested");
+            return setupQueryPassthroughSplit(getSplitsRequest);
+        }
+
         int partitionContd = decodeContinuationToken(getSplitsRequest);
         Set<Split> splits = new HashSet<>();
         Block partitions = getSplitsRequest.getPartitions();
--- a/athena-clickhouse/src/main/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseRecordHandler.java
+++ b/athena-clickhouse/src/main/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseRecordHandler.java
@@ -91,10 +91,16 @@ public class ClickHouseRecordHandler
     public PreparedStatement buildSplitSql(Connection jdbcConnection, String catalogName, TableName tableName, Schema schema, Constraints constraints, Split split)
             throws SQLException
     {
-        PreparedStatement preparedStatement = jdbcSplitQueryBuilder.buildSql(jdbcConnection, null, tableName.getSchemaName(), tableName.getTableName(), schema, constraints, split);
+        PreparedStatement preparedStatement;
 
-        // Disable fetching all rows.
-        preparedStatement.setFetchSize(Integer.MIN_VALUE);
+        if (constraints.isQueryPassThrough()) {
+            preparedStatement = buildQueryPassthroughSql(jdbcConnection, constraints);
+        }
+        else {
+            preparedStatement = jdbcSplitQueryBuilder.buildSql(jdbcConnection, null, tableName.getSchemaName(), tableName.getTableName(), schema, constraints, split);
+        }
+        // Set fetch size to a reasonable value for ClickHouse JDBC driver.
+        preparedStatement.setFetchSize(ClickHouseConstants.FETCH_SIZE);
 
         return preparedStatement;
     }
--- a/athena-clickhouse/src/test/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseMetadataHandlerTest.java
+++ b/athena-clickhouse/src/test/java/com/amazonaws/athena/connectors/clickhouse/ClickHouseMetadataHandlerTest.java
@@ -25,6 +25,8 @@ import com.amazonaws.athena.connector.lambda.data.SchemaBuilder;
 import com.amazonaws.athena.connector.lambda.domain.Split;
 import com.amazonaws.athena.connector.lambda.domain.TableName;
 import com.amazonaws.athena.connector.lambda.domain.predicate.Constraints;
+import com.amazonaws.athena.connector.lambda.metadata.GetDataSourceCapabilitiesRequest;
+import com.amazonaws.athena.connector.lambda.metadata.GetDataSourceCapabilitiesResponse;
 import com.amazonaws.athena.connector.lambda.metadata.GetSplitsRequest;
 import com.amazonaws.athena.connector.lambda.metadata.GetSplitsResponse;
 import com.amazonaws.athena.connector.lambda.metadata.GetTableLayoutRequest;
@@ -38,6 +40,7 @@ import com.amazonaws.athena.connectors.jdbc.connection.JdbcConnectionFactory;
 import com.amazonaws.athena.connector.credentials.CredentialsProvider;
 import org.apache.arrow.vector.types.pojo.Field;
 import org.apache.arrow.vector.types.pojo.Schema;
+import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
@@ -54,6 +57,7 @@ import java.sql.SQLException;
 import java.sql.Types;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
@@ -65,9 +69,16 @@ import static org.mockito.ArgumentMatchers.nullable;
 public class ClickHouseMetadataHandlerTest
         extends TestBase
 {
-    private DatabaseConnectionConfig databaseConnectionConfig = new DatabaseConnectionConfig("default", 
-    ClickHouseConstants.NAME,
-    "clickhouse://jdbc:clickhouse://localhost/user=A&password=B");
+    private static final String TEST_CATALOG = "testCatalog";
+    private static final String TEST_SCHEMA = "testSchema";
+    private static final String TEST_TABLE = "testTable";
+    private static final String TEST_QUERY_ID = "testQueryId";
+    private static final String TEST_SECRET = "testSecret";
+    private static final String TEST_CONNECTION_STRING = "clickhouse://jdbc:clickhouse://localhost/user=A&password=B";
+    private static final String TEST_SECRET_STRING = "{\"username\": \"testUser\", \"password\": \"testPassword\"}";
+    private DatabaseConnectionConfig databaseConnectionConfig = new DatabaseConnectionConfig("default",
+            ClickHouseConstants.NAME,
+            TEST_CONNECTION_STRING);
     private ClickHouseMetadataHandler metadataHandler;
     private JdbcConnectionFactory jdbcConnectionFactory;
     private Connection connection;
@@ -85,10 +96,18 @@ public class ClickHouseMetadataHandlerTest
         Mockito.when(this.jdbcConnectionFactory.getConnection(nullable(CredentialsProvider.class))).thenReturn(this.connection);
         this.secretsManager = Mockito.mock(SecretsManagerClient.class);
         this.athena = Mockito.mock(AthenaClient.class);
-        Mockito.when(this.secretsManager.getSecretValue(Mockito.eq(GetSecretValueRequest.builder().secretId("testSecret").build()))).thenReturn(GetSecretValueResponse.builder().secretString("{\"username\": \"testUser\", \"password\": \"testPassword\"}").build());
+        Mockito.when(this.secretsManager.getSecretValue(Mockito.eq(GetSecretValueRequest.builder().secretId(TEST_SECRET).build()))).thenReturn(GetSecretValueResponse.builder().secretString(TEST_SECRET_STRING).build());
         this.metadataHandler = new ClickHouseMetadataHandler(databaseConnectionConfig, this.secretsManager, this.athena, this.jdbcConnectionFactory, com.google.common.collect.ImmutableMap.of());
         this.federatedIdentity = Mockito.mock(FederatedIdentity.class);
         this.blockAllocator = Mockito.mock(BlockAllocator.class);
+    }
+
+    @After
+    public void tearDown()
+    {
+        if (this.blockAllocator != null) {
+            this.blockAllocator.close();
+        }
     }
 
     @Test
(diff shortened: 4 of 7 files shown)


athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
```diff
@@ -27,31 +27,85 @@
...
+        String expectedConnectionString = "hive://jdbc:hive2://localhost:49100/default;${testSecret}";
+        assertEquals(expectedConnectionString, clouderaHiveConnectionProperties.get(DEFAULT));
+    }
+
+    @Test
+    public void testClouderaHiveConnectionProperties_withSessionConfsOnly()
+    {
+        connectionProperties.put(SESSION_CONFS, "hive.execution.engine=mr");
+        Map<String, String> clouderaHiveConnectionProperties = clouderaHiveEnvironmentProperties.connectionPropertiesToEnvironment(connectionProperties);
+        String expectedConnectionString = "hive://jdbc:hive2://localhost:49100/default;hive.execution.engine=mr${testSecret}";
```
AbdulR3hman: this string `hive://jdbc:hive2://localhost:49100` used multiple times, can we make it a static variable?

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
```diff
@@ -27,31 +27,85 @@
...
+        assertEquals(expectedConnectionString, clouderaHiveConnectionProperties.get(DEFAULT));
+    }
+
+    @Test
+    public void testClouderaHiveConnectionProperties_withSessionAndHiveConfs()
+    {
+        connectionProperties.put(SESSION_CONFS, "hive.execution.engine=mr");
```
AbdulR3hman, line 73: same here; `hive.execution.engine=mr`

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveRecordHandlerTest.java
```diff
@@ -167,10 +170,50 @@ public void buildSplitSqlTimestamp()
...
+                    .put("schemaFunctionName", "system.query")
+                    .put("enableQueryPassthrough", "true")
+                    .put("name", "query")
+                    .put("schema", "system")
+                    .build();
+
+            Mockito.when(constraints.getQueryPassthroughArguments()).thenReturn(queryPassthroughArgs);
+
+            PreparedStatement expectedPreparedStatement = Mockito.mock(PreparedStatement.class);
+            Mockito.when(this.connection.prepareStatement(nullable(String.class))).thenReturn(expectedPreparedStatement);
```
AbdulR3hman: I would use `Assert.assertSame` to make sure it returns the same prepared statement you've passed to it in the mockito 

you can also do 

```
Mockito.verify(constraints).isQueryPassThrough();
Mockito.verifyNoMoreInteractions(this.connection);
```

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveRecordHandlerTest.java
```diff
@@ -167,10 +170,50 @@ public void buildSplitSqlTimestamp()
...
         Timestamp expectedTimestamp = new Timestamp(timestamp.toInstant(ZoneOffset.UTC).toEpochMilli());
         Assert.assertEquals(expectedPreparedStatement, preparedStatement);
         Mockito.verify(preparedStatement, Mockito.times(1))
                 .setTimestamp(1, expectedTimestamp);
     }
+
+    @Test
+    public void testBuildSplitSql_withQueryPassthrough()
+    {
+        try {
```
AbdulR3hman: try/catch in unit test is not a good practice; just let the exception bubble up; 

```
public void testBuildSplitSql_withQueryPassthrough() throws Exception {
```

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveRecordHandlerTest.java
```diff
@@ -167,10 +170,50 @@ public void buildSplitSqlTimestamp()
...
         Assert.assertEquals(expectedPreparedStatement, preparedStatement);
-        LocalDateTime timestampExp = LocalDateTime.parse("2024-10-03 12:34:56", formatter);
         Timestamp expectedTimestamp = new Timestamp(timestamp.toInstant(ZoneOffset.UTC).toEpochMilli());
         Assert.assertEquals(expectedPreparedStatement, preparedStatement);
         Mockito.verify(preparedStatement, Mockito.times(1))
                 .setTimestamp(1, expectedTimestamp);
     }
+
+    @Test
+    public void testBuildSplitSql_withQueryPassthrough()
```
AbdulR3hman: Missing case: 

1. Passthrough disabled
* isQueryPassThrough() returns false → verify the non-passthrough SQL path (likely uses split properties/constraints).

2. Passthrough enabled but missing/invalid args
* Missing QUERY → expect an exception or fallback behavior.
* Wrong schemaFunctionName / enableQueryPassthrough not "true" → ensure handler fails or ignores passthrough as intended.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveRecordHandlerTest.java
```diff
@@ -167,10 +170,50 @@ public void buildSplitSqlTimestamp()
...
+            SchemaBuilder schemaBuilder = SchemaBuilder.newBuilder();
+            schemaBuilder.addField(FieldBuilder.newBuilder("testCol1", Types.MinorType.INT.getType()).build());
+            schemaBuilder.addField(FieldBuilder.newBuilder("partition", Types.MinorType.VARCHAR.getType()).build());
+            Schema schema = schemaBuilder.build();
+
+            Split split = Mockito.mock(Split.class);
+            Mockito.when(split.getProperties()).thenReturn(Collections.singletonMap("partition", "p0"));
+            Mockito.when(split.getProperty(Mockito.eq("partition"))).thenReturn("p0");
+
+            Constraints constraints = Mockito.mock(Constraints.class);
```
AbdulR3hman: nit: constraints is a pretty straightforward class; I don't think you need to mock it. you can just instantiate it.

AbdulR3hman (changes requested):

chngpe (approved):

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
Jithendar12: Hi @AbdulR3hman , thank you for reviewing.
I’ve added static constants for the repeated strings.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
Jithendar12: Created constant.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveRecordHandlerTest.java
Jithendar12: Used assertSame instead of assertEquals for mock verifications.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveRecordHandlerTest.java
Jithendar12: Removed try/catch in favor of throws Exception. Initially, we had used throws Exception, but based on the previous review comment in [SDK unit test PR](https://github.com/awslabs/aws-athena-query-federation/pull/2715#discussion_r2104713586), we have updated it to use try/catch:

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveRecordHandlerTest.java
Jithendar12: Added missing test cases for query passthrough: 

-  Disabled passthrough 
-  Missing QUERY argument 
-  Wrong schemaFunctionName 
- For the last condition (enableQueryPassthrough not "true"), we did not add a test case because this is not currently validated in the code. The logic in buildSplitSql method only checks [constraints.isQueryPassThrough()](https://github.com/awslabs/aws-athena-query-federation/blob/3a17ba1d74e2badb30d72cdd958b90140bae182b/athena-cloudera-hive/src/main/java/com/amazonaws/athena/connectors/cloudera/HiveRecordHandler.java#L76C9-L78C10), which simply verifies that the queryPassthroughArguments map is non-empty. Since there’s no explicit validation for the value of enableQueryPassthrough, the handler will never enter a failure path for this case. If we want the connector to reject invalid values for enableQueryPassthrough, that would require changes to the main code (e.g., adding explicit validation).

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveRecordHandlerTest.java
Jithendar12: Used actual Constraints instances.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
```diff
@@ -27,31 +27,93 @@
...
+        String expectedConnectionString = BASE_CONNECTION_STRING + TEST_SECRET;
+        assertEquals(expectedConnectionString, clouderaHiveConnectionProperties.get(DEFAULT));
+    }
+
+    @Test
+    public void testClouderaHiveConnectionProperties_withSessionConfsOnly()
+    {
+        connectionProperties.put(SESSION_CONFS, HIVE_EXECUTION_ENGINE_MR);
+        Map<String, String> clouderaHiveConnectionProperties = clouderaHiveEnvironmentProperties.connectionPropertiesToEnvironment(connectionProperties);
+        String expectedConnectionString = BASE_CONNECTION_STRING + HIVE_EXECUTION_ENGINE_MR + TEST_SECRET;
```
ejeffrli: Is there no '&' needed between HIVE_EXECUTION_ENGINE_MR and TEST_SECRET? Can we test query behavior with a real connection?

ejeffrli: I think the actual function getJdbcParameters is problematic. Can we verify a connection can actually be made with session confs + secret?

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
```diff
@@ -27,31 +27,93 @@
...
+        assertEquals(expectedConnectionString, clouderaHiveConnectionProperties.get(DEFAULT));
+    }
+
+    @Test
+    public void testClouderaHiveConnectionProperties_withSessionAndHiveConfs()
```
ejeffrli: Verify as well

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
```diff
@@ -27,31 +27,93 @@
...
+    {
+        connectionProperties.put(SESSION_CONFS, HIVE_EXECUTION_ENGINE_MR);
+        connectionProperties.put(HIVE_CONFS, MAPREDUCE_TESTING);
+        Map<String, String> clouderaHiveConnectionProperties = clouderaHiveEnvironmentProperties.connectionPropertiesToEnvironment(connectionProperties);
+        String expectedConnectionString = BASE_CONNECTION_STRING + HIVE_EXECUTION_ENGINE_MR + ";" + MAPREDUCE_TESTING + TEST_SECRET;
+        assertEquals(expectedConnectionString, clouderaHiveConnectionProperties.get(DEFAULT));
+    }
+
+    @Test
+    public void testClouderaHiveConnectionProperties_withAllConfigs()
```
ejeffrli: Verify as well

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
```diff
@@ -27,31 +27,93 @@
...
+        connectionProperties.put(SESSION_CONFS, HIVE_EXECUTION_ENGINE_MR);
+        connectionProperties.put(HIVE_CONFS, MAPREDUCE_TESTING);
+        connectionProperties.put(HIVE_VARS, ENV_TEST);
+        Map<String, String> clouderaHiveConnectionProperties = clouderaHiveEnvironmentProperties.connectionPropertiesToEnvironment(connectionProperties);
+        String expectedConnectionString = BASE_CONNECTION_STRING + HIVE_EXECUTION_ENGINE_MR + ";" + MAPREDUCE_TESTING + ";" + ENV_TEST + ";" + TEST_SECRET;
+        assertEquals(expectedConnectionString, clouderaHiveConnectionProperties.get(DEFAULT));
+    }
 
+    @Test
+    public void testClouderaHiveConnectionProperties_withHiveVarsOnly()
```
ejeffrli: Verify as well

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
```diff
@@ -27,31 +27,93 @@
...
+    {
+        connectionProperties.put(HIVE_VARS, ENV_TEST);
         Map<String, String> clouderaHiveConnectionProperties = clouderaHiveEnvironmentProperties.connectionPropertiesToEnvironment(connectionProperties);
+        String expectedConnectionString = BASE_CONNECTION_STRING + ENV_TEST + ";" + TEST_SECRET;
+        assertEquals(expectedConnectionString, clouderaHiveConnectionProperties.get(DEFAULT));
+    }
 
-        String expectedConnectionString = "hive://jdbc:hive2://50.100.00.10:49100/default;${testSecret}";
+    @Test
+    public void testClouderaHiveConnectionProperties_withHiveConfsAndVars()
```
ejeffrli: Verify as well

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveJdbcConnectionFactoryTest.java
```diff
@@ -32,19 +33,61 @@
...
+        );
+
+        // Setup other required parameters
+        HiveJdbcConnectionFactory connectionFactory = getHiveJdbcConnectionFactory(databaseConnectionConfig);
+
+        // Attempt to get connection with null credentials - this should use the original JDBC string
+        Connection connection = connectionFactory.getConnection(null);
+
+        // Verify the connection URL matches the original JDBC string
+        assertEquals(testJdbcString, connection.getMetaData().getURL());
```
ejeffrli: does this line get hit if SQLException is thrown?

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveJdbcConnectionFactoryTest.java
```diff
@@ -32,19 +33,61 @@
...
+    }
+
+    private static HiveJdbcConnectionFactory getHiveJdbcConnectionFactory(DatabaseConnectionConfig databaseConnectionConfig)
```
ejeffrli: does this need to be static?

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveMetadataHandlerTest.java
```diff
@@ -237,37 +277,111 @@ public void doGetSplits()
...
-        GetSplitsResponse getSplitsResponse = this.hiveMetadataHandler.doGetSplits(splitBlockAllocator, getSplitsRequest);
-        Assert.assertEquals(2, getSplitsResponse.getSplits().size());
+        GetSplitsRequest getSplitsRequest = new GetSplitsRequest(this.federatedIdentity, TEST_QUERY_ID, TEST_CATALOG_NAME,
+                tempTableName, getTableLayoutResponse.getPartitions(), new ArrayList<>(partitionCols), constraints, null);
+        GetSplitsResponse getSplitsResponse = this.hiveMetadataHandler.doGetSplits(blockAllocator, getSplitsRequest);
+        assertEquals(2, getSplitsResponse.getSplits().size());
     }
 
+    @Test
+    public void testDoGetSplits_withQueryPassthrough() {
```
ejeffrli: what is the expected difference for doGetSplits when using query pass through and when not using query pass through? Is this test here to ensure the split is size 1?

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
Trianz-Akshay: You're right — & is not required. That’s why we override getJdbcParameters in the Cloudera Hive implementation. There was one scenario where we were missing a ;, which is now fixed. The changes have been tested with a real connection as well.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
Trianz-Akshay: verified and updated.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
Trianz-Akshay: verified and updated.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
Trianz-Akshay: verified and updated.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
Trianz-Akshay: verified and updated.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveJdbcConnectionFactoryTest.java
Trianz-Akshay: You're right — that line doesn’t get hit if an SQLException is thrown. The assertions weren’t being executed, so I removed them. Thanks for pointing this out.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveJdbcConnectionFactoryTest.java
Trianz-Akshay: Removed. Thanks for pointing out.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/HiveMetadataHandlerTest.java
Trianz-Akshay: Yes, one of the differences is that the split size is 1 when using query passthrough. Additionally, the split contains other QPT-specific parameters. I’ve now added assertions to verify these as well.

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
ejeffrli: did the real connection have all the types of CONFS? SESSION_CONFS, HIVE_CONFS, etc.

Trianz-Akshay: Yes, the real connection does support all the configuration types, including SESSION_CONFS, HIVE_CONFS, and others.
For more details, please refer to the attached documentation:
https://docs.cloudera.com/cdw-runtime/1.5.4/securing-hive/topics/hive_connection_string_url_syntax.html

athena-cloudera-hive/src/test/java/com/amazonaws/athena/connectors/cloudera/ClouderaHiveEnvironmentPropertiesTest.java
```diff
@@ -27,31 +27,93 @@
...
         clouderaHiveEnvironmentProperties = new ClouderaHiveEnvironmentProperties();
     }
 
     @Test
-    public void clouderaHiveConnectionPropertiesTest() {
+    public void connectionPropertiesToEnvironment_WithValidProperties_ReturnsCorrectConnectionString()
+    {
+        Map<String, String> clouderaHiveConnectionProperties = clouderaHiveEnvironmentProperties.connectionPropertiesToEnvironment(connectionProperties);
+
+        String expectedConnectionString = BASE_CONNECTION_STRING + TEST_SECRET;
```
ejeffrli: can we show the actual expected string here instead of constructing for all these assertions? That way the expected string is easier to read

VenkatasivareddyTR: Hi @ejeffrli, thank you for reviewing the changes.
 
Originally, we used the full expected string directly within the assertions. However, following Abdul’s recommendation, we refactored the value into a constant to avoid duplication and improve maintainability.
Reference to the suggestion: https://github.com/awslabs/aws-athena-query-federation/pull/2834#discussion_r2278151505.
 
Please let us know if you would prefer that we still replace the constant with the explicit expected string for readability. We are happy to update it if needed.

ejeffrli: I think literal strings are better for this scenario. Constants should be used if it's related to test infrastructure, but having strings plainly written will make debugging faster.

Jithendar12: Hi @ejeffrli, thank you for your feedback. We’ve removed the constants and updated the tests to use string literals for the expected connection string assertion statements.

PR #2836, opened 2025-06-20
added unit tests for athena-oracle.

ritiktrianz:
*Issue #, if available:*

*Description of changes:*
Added unit tests for athena-oracle connector.


athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
```diff
@@ -19,33 +19,140 @@
...
         String driverClass = drv.getClass().getName();
-        Assert.assertEquals("oracle.jdbc.OracleDriver", driverClass);
+        assertEquals("oracle.jdbc.OracleDriver", driverClass);
     }
 
+    @Test
+    public void testGetConnection_withSsl() throws Exception {
+        Driver mockDriver = mock(Driver.class);
+
+            withEnvironmentVariable("is_fips_enabled", "true").execute(() -> {
```
aimethed: Looked in the code and this is the correct check. But what is the behavior for connectors using glue connections? Do customers need to set the lambda environment variable as well?

ritiktrianz: Hi @aimethed ,
Thank you for reviewing. Yes, If customers want to use FIPS mode, they should set the Lambda environment variable "is_fips_enabled" to true. Please find the official documentation link below for your reference.
https://docs.aws.amazon.com/athena/latest/ug/connectors-oracle.html#connectors-oracle-parameters.

aimethed (dismissed):

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
```diff
@@ -19,33 +19,148 @@
...
-        Connection connection =  new OracleJdbcConnectionFactory(databaseConnectionConfig, DatabaseConnectionInfo).getConnection(credentialsProvider);
+        DatabaseConnectionConfig databaseConnectionConfig = new DatabaseConnectionConfig(TEST_CATALOG, OracleConstants.ORACLE_NAME,
+                CONNECTION_STRING, TEST_SECRET);
+        DatabaseConnectionInfo databaseConnectionInfo = new DatabaseConnectionInfo(OracleConstants.ORACLE_DRIVER_CLASS, OracleConstants.ORACLE_DEFAULT_PORT);
+        Connection connection =  new OracleJdbcConnectionFactory(databaseConnectionConfig, databaseConnectionInfo).getConnection(credentialsProvider);
         String originalURL = connection.getMetaData().getURL();
         Driver drv = DriverManager.getDriver(originalURL);
         String driverClass = drv.getClass().getName();
-        Assert.assertEquals("oracle.jdbc.OracleDriver", driverClass);
+        assertEquals("oracle.jdbc.OracleDriver", driverClass);
```
christov: If I am reading this test correctly, it is expected to throw an exception. If that is the case, why is there an `assert` statement?

Also, which statement is expected to throw the exception and why?

A couple of suggestions if my understanding of this test is correct:
- Remove the `assert` statement and any irrelevant code and keep only the code necessary to trigger the exception.
- Update the test name to reflect the situation that this test targets. For example, `testGetConnection_throwsAnExceptionWhen...`

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
```diff
@@ -19,33 +19,148 @@
...
     }
 
+    @Test
+    public void testGetConnection_withSsl() throws Exception {
```
christov: Please, specify what this test tests in the test method name (see previous [comment](https://github.com/awslabs/aws-athena-query-federation/pull/2836/files#r2467012310) for naming convention suggestion. Also, the existing test on line 154 follows a similar convention).

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
```diff
@@ -19,33 +19,148 @@
...
+        Properties sslProps = capturedProps[0];
+        assertEquals("JKS", sslProps.getProperty("javax.net.ssl.trustStoreType"));
+        assertEquals("changeit", sslProps.getProperty("javax.net.ssl.trustStorePassword"));
+        assertEquals("true", sslProps.getProperty("oracle.net.ssl_server_dn_match"));
+        assertEquals(
+                "(TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384, TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256, TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA, TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA)",
+                sslProps.getProperty("oracle.net.ssl_cipher_suites")
+        );
+
+        DriverManager.deregisterDriver(mockDriver);
```
christov: This clean-up code will not be executed if one of the assertions fails. It will be safer if a `try`/`finally` statement is used.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
```diff
@@ -19,33 +19,148 @@
...
+        DatabaseConnectionInfo info = new DatabaseConnectionInfo(OracleConstants.ORACLE_DRIVER_CLASS, OracleConstants.ORACLE_DEFAULT_PORT);
+
+        Driver mockDriver = mock(Driver.class);
+        when(mockDriver.acceptsURL(anyString())).thenReturn(true);
+        when(mockDriver.connect(anyString(), any(Properties.class)))
+                .thenThrow(new SQLException("DB down", "08001", 1234));
+
+        DriverManager.registerDriver(mockDriver);
+
+            new OracleJdbcConnectionFactory(config, info).getConnection(credentialsProvider);
```
christov, line 138: Incorrect indentation of this and the next line.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
```diff
@@ -19,33 +19,148 @@
...
+            DriverManager.deregisterDriver(mockDriver);
```
christov, line 141: I don't think this line will run given that an exception is expected to be thrown earlier. See a previous comment about using `try`/`finally`.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
```diff
@@ -19,33 +19,148 @@
...
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void getConnection_whenDriverClassNotFound_throwsRuntimeException()
```
christov: Inconsistent naming: this method should also start with `test` like the rest of the methods in this class. Alternatively, each test method can start with the name of the method under test, but we need to be consistent.

I am leaning toward the latter--skip `test` since the method is already annotated with `@Test` and it is inside of a class that has `Test` in its name.

Let's do the following for all Oracle unit tests:
- For each class with production code, have one test class that has the same name but it ends with `Test`
- For each method we're testing (usually these are the methods externally visible from a class), have one or more test methods that follow the naming convention `<methodUnderTest>_<doesXyzWhenAbc>`.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
Jithendar12: Hi @christov  thank you for reviewing.
This was an existing test, which was incorrect. A similar (and correct) test case has already been added:
[getConnection_whenDriverClassNotFound_throwsRuntimeException()](https://github.com/Jithendar12/aws-athena-query-federation/blob/e80e31ae6a25f2fd8af140d8d840177ae5a826c7/athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java#L154).
Hence, this test has been removed.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
Jithendar12: The test method has been renamed accordingly.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
Jithendar12: This statement has been moved into a finally block.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
Jithendar12: Fixed.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
Jithendar12: All other test methods have been renamed to align with this naming convention.

Jithendar12:
> Please, address the comments in this review.
> 
> For this [comment](https://github.com/awslabs/aws-athena-query-federation/pull/2836/files#r2467115837), please apply its test naming suggestions across all Oracle unit tests.

All test methods across the Oracle connector have been renamed to follow the convention:
MethodName_StateUnderTest_ExpectedBehavior

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleJdbcConnectionFactoryTest.java
Jithendar12: This statement has been moved into a finally block.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/resolver/OracleJDBCCaseResolverTest.java
```diff
@@ -0,0 +1,132 @@
...
+    @Before
+    public void setup() throws SQLException
+    {
+        mockConnection = Mockito.mock(Connection.class);
+        preparedStatement = Mockito.mock(PreparedStatement.class);
+        when(mockConnection.prepareStatement(any())).thenReturn(preparedStatement);
+    }
+
+    @Test
+    public void getAdjustedSchemaNameString_whenCaseInsensitiveMode_schemaNameCasingAdjusted() throws SQLException
```
christov: Nitpick: The test name will be more informative if it said how the schema name was adjusted. E.g., "schema name is lower cased" or something along those lines. This applies to other tests in this class too.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleMetadataHandlerTest.java
```diff
@@ -91,24 +114,29 @@ public void setup()
...
 
+    @After
+    public void tearDown()
+    {
+        blockAllocator.close();
+    }
+
     @Test
-    public void getPartitionSchema()
+    public void getPartitionSchema_whenCalled_returnsPartitionSchema()
```
christov: Nitpick: "whenCalled" is not necessary--it is clear that the method will need to be called to return something. In this situation, it's fine to just skip the "when" part from the method name--that's the vanilla case/happy path. It is useful to add "when ..." when there is a specific condition/situation that is being tested (which is usually all other cases except for the "vanilla case").

"whenCalled" can be removed from other test methods too.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleMetadataHandlerTest.java
```diff
@@ -335,8 +404,76 @@ public void doGetTable()
...
-        Assert.assertEquals(expected, getTableResponse.getSchema());
-        Assert.assertEquals(inputTableName, getTableResponse.getTableName());
-        Assert.assertEquals("testCatalog", getTableResponse.getCatalogName());
+        assertEquals(expected, getTableResponse.getSchema());
+        assertEquals(inputTableName, getTableResponse.getTableName());
+        assertEquals("testCatalog", getTableResponse.getCatalogName());
+    }
+
+    @Test
+    public void doGetTable_whenUnsupportedColumnTypes_fallbackToVarchar() throws Exception
```
christov: Nitpick: "fallsBackToVarchar" (i.e., use a verb since this is about what the method does, and to be consistent with the naming of the other tests).

christov (changes requested):
Thank you for addressing the previous comments. The test names are much clearer now! I had just a few minor nitpicks on the latest version.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/resolver/OracleJDBCCaseResolverTest.java
Jithendar12: Updated test names to clarify how the schema/table name is adjusted.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleMetadataHandlerTest.java
Jithendar12: Removed unnecessary "whenCalled" from test names.

athena-oracle/src/test/java/com/amazonaws/athena/connectors/oracle/OracleMetadataHandlerTest.java
Jithendar12: Renamed with fallsBackToVarchar.

Jithendar12:
> Thank you for addressing the previous comments. The test names are much clearer now! I had just a few minor nitpicks on the latest version.

Hi @christov , thanks for the suggestions! I’ve addressed these minor nitpicks as well.

aimethed (approved):

PR #2846, opened 2025-06-23
Adding support to use custom SecretManagerClient for Google Big Query

JawedRakib:
*Issue #, if available:* 
N/A

*Description of changes:*
This PR we have added to support custom object of SecretsManagerClient.

Added constructor in BigQueryMetadataHandler
Updated the constructor in BigQueryRecordHandler
Overloaded two methods getBigQueryClient and getCredentialsFromSecretsManager of BigQueryUtils which will now accept object of SecretsManagerClient as parameter.

### DIFF
--- a/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
+++ b/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
@@ -46,6 +46,7 @@ import com.amazonaws.athena.connector.lambda.metadata.optimizations.pushdown.Com
 import com.amazonaws.athena.connector.lambda.metadata.optimizations.pushdown.FilterPushdownSubType;
 import com.amazonaws.athena.connector.lambda.metadata.optimizations.pushdown.LimitPushdownSubType;
 import com.amazonaws.athena.connector.lambda.metadata.optimizations.pushdown.TopNPushdownSubType;
+import com.amazonaws.athena.connector.lambda.security.EncryptionKeyFactory;
 import com.amazonaws.athena.connectors.google.bigquery.qpt.BigQueryQueryPassthrough;
 import com.google.api.gax.paging.Page;
 import com.google.cloud.bigquery.BigQuery;
@@ -67,11 +68,14 @@ import org.apache.arrow.vector.types.pojo.FieldType;
 import org.apache.arrow.vector.types.pojo.Schema;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.services.athena.AthenaClient;
+import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 
 import static com.amazonaws.athena.connector.lambda.metadata.ListTablesRequest.UNLIMITED_PAGE_SIZE_VALUE;
 import static com.amazonaws.athena.connectors.google.bigquery.BigQueryUtils.fixCaseForDatasetName;
@@ -83,6 +87,7 @@ public class BigQueryMetadataHandler
     extends MetadataHandler
 {
     private static final Logger logger = LoggerFactory.getLogger(BigQueryMetadataHandler.class);
+    private final SecretsManagerClient secretsManager;
     private final String projectName = configOptions.get(BigQueryConstants.GCP_PROJECT_ID) != null ?
             configOptions.get(BigQueryConstants.GCP_PROJECT_ID).toLowerCase() : null;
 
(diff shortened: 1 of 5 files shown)


athena-elasticsearch/src/main/java/com/amazonaws/athena/connectors/elasticsearch/ElasticsearchRecordHandler.java
```diff
@@ -102,7 +102,7 @@ public ElasticsearchRecordHandler(Map<String, String> configOptions)
     }
 
     @VisibleForTesting
```
AbdulR3hman, line 104: if this is public now; this is no longer needed. it is visible by definition.

JawedRakib: We have now removed above changes

athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryUtils.java
```diff
@@ -89,6 +89,17 @@ public static Credentials getCredentialsFromSecretsManager(java.util.Map<String,
                         "https://www.googleapis.com/auth/drive"));
```
burhan94: Please take the strings in line 88, 89, 99, 100 and make them into constants either in this file or in a util file.

burhan94 (changes requested):

athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryUtils.java
JawedRakib: As per suggestion, We have created scope constants for Big Query in BigQueryUtils class

athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
```diff
@@ -91,6 +96,19 @@ public class BigQueryMetadataHandler
     BigQueryMetadataHandler(java.util.Map<String, String> configOptions)
     {
         super(BigQueryConstants.SOURCE_TYPE, configOptions);
+        secretsManager = SecretsManagerClient.create();
```
burhan94: the parent class already has access to secrets manager. Why do we need to create another client? The client we create in the parent class has caching logic and a ton of additional addons

burhan94 (changes requested):

athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
pearlswan: @JawedRakib the parent class has a constructor that takes secret manager client, can we see if we can just use this one in glue connector https://github.com/awslabs/aws-athena-query-federation/blob/b8ceb736f6301686fe04424c93d9670664111057/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java#L175C5-L183C6

JawedRakib: We updated the code and now using secret manager of parent class instead of initializing and creating new secret manager client for BigQuery

athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
```diff
@@ -93,6 +98,17 @@ public class BigQueryMetadataHandler
         super(BigQueryConstants.SOURCE_TYPE, configOptions);
     }
 
+    public BigQueryMetadataHandler(EncryptionKeyFactory encryptionKeyFactory,
```
burhan94, line 101: quick question, why is this new constructor needed?

burhan94 (changes requested):
3char

athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
JawedRakib: We need this constructor in **BigQueryMetadataHandler** to use custom secretManager other wise we are getting error '**java.util.NoSuchElementException: No value present at java.base/java.util.Optional.get(Optional.java:143)**', We are using below code to create object of BigQueryMetadataHandler

```
.....
createPrivateObject(BigQueryMetadataHandler.class, null,
secretsManager, athena, null, null, null, configurationOptions)
.....
public static <T> T createPrivateObject(Class<T> clazz, Object... args) {
        Constructor<?>[] constructors = clazz.getDeclaredConstructors();
        Constructor<?> constructor = (Constructor)Arrays.stream(constructors).filter((c) -> {
            return c.getParameterCount() == args.length;
        }).findFirst().get();
        constructor.setAccessible(true);

        try {
            return constructor.newInstance(args);
        } catch (IllegalAccessException | InvocationTargetException | InstantiationException var5) {
            ReflectiveOperationException e = var5;
            throw new RuntimeException(e);
        }
    }
```

athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
```diff
@@ -123,7 +139,7 @@ public ListSchemasResponse doListSchemaNames(BlockAllocator blockAllocator, List
         logger.info("doListSchemaNames called with Catalog: {}", listSchemasRequest.getCatalogName());
 
         final List<String> schemas = new ArrayList<>();
-        BigQuery bigQuery = BigQueryUtils.getBigQueryClient(configOptions);
+        BigQuery bigQuery = BigQueryUtils.getBigQueryClient(configOptions, getSecret(getEnvBigQueryCredsSmId(configOptions)));
```
ejeffrli, line 142: This makes BigQueryConstants.ENV_BIG_QUERY_CREDS_SM_ID a required field right? Do we know if this field will always exist?

 Is this backwards compatible?

ejeffrli (approved):

PR #2868, opened 2025-07-07
Add support for _meta: { 'field': 'json' } to parse object fields as varchar

### LINKED ISSUE #2867: [FEATURE] Add support for _meta: { 'field': 'json' } to parse object fields as varchar - Elasticsearch/Opensearch
**Is your feature request related to a problem? If yes, please describe.**  
The Athena Elasticsearch/Opensearch connector currently fails to infer fields of type `"object"` with `"enabled": false` when no properties are explicitly defined. This makes it impossible to query such fields via Athena, even though they exist in the source index.

The connector already supports a workaround for similar scenarios using `_meta: { "field": "list" }`, which allows those fields to be properly parsed as arrays.

---

**Describe the solution you'd like**  
I'd like to extend the existing `_meta` mechanism by introducing support for `_meta: { "field": "json" }`. When this meta is present, the connector will parse the field's content as a JSON string and store it as a `varchar` in Athena.

---

**Describe alternatives you've considered**  
- Manually defining properties inside the `"object"` field mapping, which isn't always feasible, especially when the structure is dynamic or unknown at index creation time.  

---

**Additional context**  
This proposed solution follows the same idea already implemented for `_meta: { "field": "list" }`, ensuring minimal code change and consistent behavior.  

I already have a working implementation of this feature and would be happy to submit a pull request after opening this issue.

fksegundo:
## Summary

This pull request adds support for parsing Elasticsearch/Opensearch fields of type `"object"` with `"enabled": false` using a `_meta: { "field": "json" }` flag.

## Motivation

Currently, fields with `"enabled": false` and no defined properties cannot be inferred by the Athena connector, unless `_meta: { "field": "list" }` is provided.

This PR extends that logic by introducing support for `_meta: { "field": "json" }`. When this flag is present, the connector parses the field's content as a JSON string and exposes it as a `varchar` in Athena, making the field queryable.

## Implementation Notes

The logic follows the same pattern already used for handling the `"list"` meta flag, ensuring minimal changes and consistent behavior.

## Related Issue

Fixes #2867


burhan94 (commented):
## Code Review

### 🔴 Critical Issues

1. **NPE risk** — `field.getMetadata().get("json")` will NPE if `getMetadata()` returns null.

2. **Jackson usage** — Consider using Gson instead of Jackson for consistency with the rest of the codebase.

### 🟡 Medium

- Error handling silently sets `dst.isSet=0` without propagating serialization failures — could mask data issues.
- Missing class-level Javadoc update for the behavior change.

Good test coverage and backward-compatible approach. Please fix the NPE risk.

fksegundo:
Hi @burhan94, thanks for the review! I've updated the PR with the following changes:

### Improvements & Fixes
* **NPE Protection:** Made metadata access null-safe in the `VARCHAR`/`json` path and added a regression test for null field metadata.
* **Documentation:** Updated class-level Javadocs to properly document the `_meta`/`json` behavior.

### Dependency Decisions
* **Jackson vs Gson:** I checked the `athena-elasticsearch` module and didn't find any existing Gson usage. Since **Jackson** is already a project dependency and sufficient for this serialization, I decided to stick with it to avoid introducing a new dependency pattern in this module.

### Note on Scope
* I noticed a similar null-safety pattern in the `BIGINT` `scaling_factor` path. I've left it unchanged for now to keep this PR focused, but let me know if you'd like me to address that here as well!

PR #2903, opened 2025-07-28
Adding support for Substrait query plan and update DDb connector implementation to use the plan

DilipReddyGaddam:
*Issue #, if available:*

*Description of changes:*
As part of query federation feature we are adding support for Substrait query plan in the athena federation SDK. We have also updated the DDB connector impl to use the plan and execute the queries.

We have already demoed the testing of these changes and We have made sure that these are back wards compatible.

### DIFF
--- a/athena-oracle/pom.xml
+++ b/athena-oracle/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <name>Amazon Athena Oracle Connector</name>
     <parent>
         <artifactId>aws-athena-query-federation</artifactId>
         <groupId>com.amazonaws</groupId>
--- a/athena-postgresql/pom.xml
+++ b/athena-postgresql/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <name>Amazon Athena PostgreSQL Connector</name>
     <parent>
         <artifactId>aws-athena-query-federation</artifactId>
         <groupId>com.amazonaws</groupId>
--- a/athena-redshift/pom.xml
+++ b/athena-redshift/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <name>Amazon Athena Redshift Connector</name>
     <parent>
         <artifactId>aws-athena-query-federation</artifactId>
         <groupId>com.amazonaws</groupId>


burhan94:
Is there a reason why the subtstrait classes are in federation-sdk-tools?

DilipReddyGaddam:
> Is there a reason why the subtstrait classes are in federation-sdk-tools?

I thought these substrait classes can be used across multiple connectors and sdk-tools might be correct place to do that. Let me know if its not correct package to do that.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/MetadataHandler.java
```diff
@@ -230,20 +229,21 @@ protected EncryptionKey makeEncryptionKey()
      * @param request 
      * @return A unique spill location.
      */
-    protected SpillLocation makeSpillLocation(MetadataRequest request)
+    public SpillLocation makeSpillLocation(MetadataRequest request)
```
burhan94: why did this need to become public?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/QueryPlan.java
```diff
@@ -30,23 +30,19 @@ public class QueryPlan
...
     /**
      *
      * @param substraitVersion
      * @param substraitPlan
-     * @param sqlConformance
      */
     @JsonCreator
     public QueryPlan(@JsonProperty("substraitVersion") String substraitVersion,
-                     @JsonProperty("substraitPlan") String substraitPlan,
-                     @JsonProperty("sqlConformance") String sqlConformance)
```
burhan94: is sqlConformance removed from the API now?

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBPredicateUtils.java
```diff
@@ -67,48 +77,19 @@ private DDBPredicateUtils() {}
     public static DynamoDBIndex getBestIndexForPredicates(DynamoDBTable table, List<String> requestedCols, Map<String, ValueSet> predicates)
```
burhan94: Can you pls use Q to generate as many tests for this class that cover the various predicates we want to support for DDB?. We've seen issues in the past with between predicates for example not working as intended due to inclusive/exclusive values at the start and end of the range.

burhan94:
> > Is there a reason why the subtstrait classes are in federation-sdk-tools?
> 
> I thought these substrait classes can be used across multiple connectors and sdk-tools might be correct place to do that. Let me know if its not correct package to do that.

Lets put this in the Federation SDK. We already publish the federation SDK to maven and it'll keep all of the frameworks together in 1 package.

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
```diff
@@ -185,6 +196,24 @@ public GetDataSourceCapabilitiesResponse doGetDataSourceCapabilities(BlockAlloca
...
+        List<StandardFunctions> supportedFunctions = new ArrayList<>();
+        supportedFunctions.add(StandardFunctions.AND_FUNCTION_NAME);
+        supportedFunctions.add(StandardFunctions.IS_NULL_FUNCTION_NAME);
+        supportedFunctions.add(StandardFunctions.EQUAL_OPERATOR_FUNCTION_NAME);
+        supportedFunctions.add(StandardFunctions.GREATER_THAN_OPERATOR_FUNCTION_NAME);
+        supportedFunctions.add(StandardFunctions.LESS_THAN_OPERATOR_FUNCTION_NAME);
+        supportedFunctions.add(StandardFunctions.GREATER_THAN_OR_EQUAL_OPERATOR_FUNCTION_NAME);
+        supportedFunctions.add(StandardFunctions.LESS_THAN_OR_EQUAL_OPERATOR_FUNCTION_NAME);
+        capabilities.put(DataSourceOptimizations.SUPPORTS_COMPLEX_EXPRESSION_PUSHDOWN.withSupportedSubTypes(
+                ComplexExpressionPushdownSubType.SUPPORTED_FUNCTION_EXPRESSION_TYPES
```
burhan94: DDB does not support complex expression pushdown nor does it support functions.

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBRecordHandler.java
```diff
@@ -271,18 +290,33 @@ private void writeItemsToBlock(
             }
             spiller.writeRows((Block block, int rowNum) -> rowWriter.writeRow(block, rowNum, item) ? 1 : 0);
             numRows++;
-            if (canApplyLimit && numRows >= recordsRequest.getConstraints().getLimit()) {
+            if (limitPair.getLeft() && numRows >= limitPair.getRight()) {
```
burhan94: can you add a comment here explaining that we're checking if limit is enabled? The variable before was descriptive and it was easy to tell what its purpose was. The limitPair hides the purpose

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/domain/predicate/QueryPlan.java
DilipReddyGaddam: We have not removed from API will do it.

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
DilipReddyGaddam: Yeah DDB doesn't support all complex expression pushdown. It only supports standard functions listed above.

burhan94: hmm, I don't think DDB supports:
$is_distinct_from
$add
$multiply
$divide
$modulus
$negate
$like_pattern

burhan94:
As discussed offline, pls create a utility class that can be used to convert arrow -> substrait  for literals and vice versa. This can then be used in the engine and in connectors.

burhan94 (approved):

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitFunctionParser.java
```diff
@@ -0,0 +1,244 @@
...
+     */
+    private static boolean isLogicalOperator(String functionName)
+    {
+        return "and:bool".equals(functionName) || "or:bool".equals(functionName);
+    }
+
+    /**
+     * Helper class to hold scalar function information.
+     */
+    private record ScalarFunctionInfo(String functionName, List<FunctionArgument> arguments)
```
AbdulR3hman: this won't work; we are now using JDK 11 -- this was introduced later. JDK 17; need to change this.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitLiteralConverter.java
```diff
@@ -0,0 +1,167 @@
...
+                Expression.Literal.Decimal decimal = literal.getDecimal();
+                arrowType = new ArrowType.Decimal(decimal.getPrecision(), decimal.getScale(), 16);
+                return Pair.of(DecimalUtil.getBigDecimalFromBytes(decimal.getValue().toByteArray(),
+                        decimal.getScale(), 16), arrowType);
+            case DATE:
+                arrowType = new ArrowType.Date(DateUnit.DAY);
+                return Pair.of((long) literal.getDate(), arrowType);
+            case TIMESTAMP:
+                if (literal.hasTimestampTz()) {
+                    arrowType = new ArrowType.Timestamp(TimeUnit.MICROSECOND, "UTC");
```
AbdulR3hman: are we always going to default this to UTC? does this literal support specific timestamp zone? and if not set in UTC don't we need to convert it to UTC first?

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBRecordHandler.java
```diff
@@ -271,18 +291,34 @@ private void writeItemsToBlock(
...
+            // If limit is enabled and records fetched is greater than limit, We can stop execution.
+            if (limitPair.getLeft() && numRows >= limitPair.getRight()) {
                 return;
             }
         }
         logger.info("readWithConstraint: numRows[{}]", numRows);
     }
 
-    private boolean canApplyLimit(Constraints constraints)
+    private boolean canApplyLimit(Constraints constraints,
```
AbdulR3hman: I think `canApplyLimit` and `getLimit` can be simplified into one method that returns an optional; that way we don't do the same logic twice.

pom.xml
```diff
@@ -234,6 +234,31 @@
             <version>${testng.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.substrait</groupId>
+            <artifactId>isthmus</artifactId>
+            <version>0.52.0</version>
```
AbdulR3hman: can we extract this version as a top level pom variable; so that when we upgrade the version for substrait dependencies; they are all updated together; very useful/helpful as well when reviewing this from the Dependency Bot PR.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/model/Operator.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package com.amazonaws.athena.connector.substrait.model;
+
+/**
+ * Represents a subset of Substrait-supported operators for query federation.
+ * This enum contains only the commonly used operators that are supported
+ * by the Athena Query Federation framework. The full Substrait specification
+ * includes many more operators, We will extend more operators as we need.
+ */
+public enum Operator
```
AbdulR3hman: can we rename this to specifically say that it is for Substrait.

DilipReddyGaddam: Yes Renamed it

pom.xml
DilipReddyGaddam: Moved this to top level pom variable

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBRecordHandler.java
DilipReddyGaddam: created a common method for this

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitLiteralConverter.java
DilipReddyGaddam: I can check and update if we have Timezone information in this literal.

PR #2905, opened 2025-07-30
Enable Code Guru Security

AbdulR3hman:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/code-guru-security.yml
+++ b/.github/workflows/code-guru-security.yml
@@ -1,0 +1,47 @@
+name: CodeGuru Security Example
+on:
+  push:
+    branches:
+      - 'master'
+
+permissions:
+  id-token: write
+  # for writing security events.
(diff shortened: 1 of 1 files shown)


christov (approved):

.github/workflows/code-guru-security.yml
```diff
@@ -0,0 +1,47 @@
+name: CodeGuru Security Example
+on:
+  push:
```
fal-bharadwaj, line 3: looks good, a minor suggestion maybe to add a trigger on creation of PRs to master, can do as a follow up later too? something like -
```
pull_request:
    branches:
      - 'master'
```

AbdulR3hman: this is a trigger on pull request against Master. so I think it should cover it. if not I'll make sure it does in next round.

PR #2922, opened 2025-08-08
Include linked accounts option when querying metric_samples table.

perkerssens:
Cross-account metrics were only fetched for the metrics table in PR2881. This change includes the cross-account metrics to metric_samples table queries.


athena-cloudwatch-metrics/src/main/java/com/amazonaws/athena/connectors/cloudwatch/metrics/MetricsMetadataHandler.java
```diff
@@ -242,40 +243,53 @@ public GetSplitsResponse doGetSplits(BlockAllocator blockAllocator, GetSplitsReq
             Set<Split> splits = new HashSet<>();
             ListMetricsRequest listMetricsRequest = listMetricsRequestBuilder.build();
             ListMetricsResponse result = invoker.invoke(() -> metrics.listMetrics(listMetricsRequest));
+            
+            logger.debug(result.toString());
```
AbdulR3hman: can we remove this please.

AbdulR3hman (approved):

athena-cloudwatch-metrics/src/test/java/com/amazonaws/athena/connectors/cloudwatch/metrics/MetricDataQuerySerDeTest.java
```diff
@@ -24,47 +24,48 @@
...
 
 import java.util.ArrayList;
 import java.util.List;
 
 import static org.junit.Assert.*;
 
-public class MetricStatSerDeTest
+public class MetricDataQuerySerDeTest
 {
-    private static final Logger logger = LoggerFactory.getLogger(MetricStatSerDeTest.class);
```
burhan94, line 36: any way we can keep the expected serialization static? The reasoning why is that the serializer/deserializer has some additional custom logic and only comparing the SerDe roundtrip could hide some issues if there ever was a bad commit. I do think the serDe roundtrip should still exist though as a separate test

burhan94 (changes requested):

athena-cloudwatch-metrics/src/test/java/com/amazonaws/athena/connectors/cloudwatch/metrics/MetricDataQuerySerDeTest.java
perkerssens: Agree, I am adding back in the check for the expected serialization string.

AbdulR3hman (approved):

PR #2932, opened 2025-08-13
Abstract common OAuth handling and add OAuth support to Athena DataLake Gen2 Connector

Jithendar12:
Issue #, if available:

Description of changes:
Abstracted common OAuth handling into a new OAuthCredentialsProvider base class within the SDK module to enable reuse across multiple connectors.
Added support for OAuth-based authentication in the DataLake Gen2 connector using Microsoft Entra ID with the client_credentials grant type. The connector now obtains Bearer tokens directly from Entra ID and securely manages them via AWS Secrets Manager.

Users must store the following credentials in AWS Secrets Manager:
{
"client_id": "&lt;your-entra-id-app-client-id>",
"client_secret": "&lt;your-entra-id-app-client-secret>",
"tenant_id": "&lt;your-entra-id-app-tenant-id>"
}

Please find attached reference and test documents.
[gen2-oauth-testing.odt](https://github.com/user-attachments/files/21749132/gen2-oauth-testing.odt)
[OAuth Integration in datalakegen2.docx](https://github.com/user-attachments/files/21749134/OAuth.Integration.in.datalakegen2.docx)
[SYNAPSE_FUNCTIONAL_TEST_2025-08-29.xlsx](https://github.com/user-attachments/files/22047681/SYNAPSE_FUNCTIONAL_TEST_2025-08-29.xlsx)
[SQLSERVER_FUNCTIONAL_TEST_2025-08-29.xlsx](https://github.com/user-attachments/files/22047682/SQLSERVER_FUNCTIONAL_TEST_2025-08-29.xlsx)
[DATALAKEGEN2_FUNCTIONAL_TEST_2025-08-29.xlsx](https://github.com/user-attachments/files/22047683/DATALAKEGEN2_FUNCTIONAL_TEST_2025-08-29.xlsx)
[SNOWFLAKE_FUNCTIONAL_TEST_2025-08-29.xlsx](https://github.com/user-attachments/files/22047684/SNOWFLAKE_FUNCTIONAL_TEST_2025-08-29.xlsx)
[ORACLE_FUNCTIONAL_TEST_2025-08-29.xlsx](https://github.com/user-attachments/files/22047685/ORACLE_FUNCTIONAL_TEST_2025-08-29.xlsx)
[CLOUDERAHIVE_FUNCTIONAL_TEST_2025-08-29.xlsx](https://github.com/user-attachments/files/22047686/CLOUDERAHIVE_FUNCTIONAL_TEST_2025-08-29.xlsx)

### DIFF
--- a/athena-cloudwatch/pom.xml
+++ b/athena-cloudwatch/pom.xml
@@ -32,7 +32,7 @@
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>cloudwatchlogs</artifactId>
-            <version>2.32.29</version>
+            <version>2.33.8</version>
             <exclusions>
                 <!-- replaced with jcl-over-slf4j -->
                 <exclusion>
--- a/athena-elasticsearch/pom.xml
+++ b/athena-elasticsearch/pom.xml
@@ -73,7 +73,7 @@
         <dependency>
             <groupId>org.elasticsearch.client</groupId>
             <artifactId>elasticsearch-rest-client</artifactId>
-            <version>9.1.2</version>
+            <version>9.1.3</version>
         </dependency>
         <!-- https://mvnrepository.com/artifact/org.elasticsearch.client/elasticsearch-rest-high-level-client -->
         <dependency>
--- a/athena-hbase/pom.xml
+++ b/athena-hbase/pom.xml
@@ -33,6 +33,22 @@
             <classifier>withdep</classifier>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4j-log4j.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+            <version>${slf4j-log4j.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-slf4j2-impl</artifactId>
+            <version>${log4j2Version}</version>
+            <scope>runtime</scope>
+        </dependency>
         <!-- https://mvnrepository.com/artifact/software.amazon.awscdk/emr -->
         <dependency>
             <groupId>software.amazon.awscdk</groupId>
--- a/athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseTableNameUtils.java
+++ b/athena-hbase/src/main/java/com/amazonaws/athena/connectors/hbase/HbaseTableNameUtils.java
@@ -60,7 +60,8 @@ public final class HbaseTableNameUtils
      */
    public static String getQualifiedTableName(String schema, String table)
    {
-       return schema + NAMESPACE_QUALIFIER + table;
+       String namespacePrefix = schema + NAMESPACE_QUALIFIER;
+       return table.startsWith(namespacePrefix) ? table : namespacePrefix + table;
    }
 
     /**
--- a/athena-hbase/src/test/java/com/amazonaws/athena/connectors/hbase/HbaseTableNameUtilsTest.java
+++ b/athena-hbase/src/test/java/com/amazonaws/athena/connectors/hbase/HbaseTableNameUtilsTest.java
@@ -42,6 +42,17 @@ public class HbaseTableNameUtilsTest
     public void getQualifiedTableName()
     {
         String table = "table";
+        testGetQualifiedTableName(table);
+    }
+
+    @Test
+    public void getQualifiedTableNameWithNamespace()
+    {
+        String table = "schema:table";
+        testGetQualifiedTableName(table);
+    }
+
+    private void testGetQualifiedTableName(String table) {
         String schema = "schema";
         String expected = "schema:table";
         String actualWithTable = HbaseTableNameUtils.getQualifiedTableName(new TableName(schema, table));
--- a/athena-redis/pom.xml
+++ b/athena-redis/pom.xml
@@ -25,7 +25,7 @@
         <dependency>
             <groupId>io.lettuce</groupId>
             <artifactId>lettuce-core</artifactId>
-            <version>6.8.0.RELEASE</version>
+            <version>6.8.1.RELEASE</version>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>


athena-datalakegen2/src/main/java/com/amazonaws/athena/connectors/datalakegen2/DataLakeGen2RecordHandler.java
```diff
@@ -52,7 +53,7 @@ public DataLakeGen2RecordHandler(java.util.Map<String, String> configOptions)
     public DataLakeGen2RecordHandler(DatabaseConnectionConfig databaseConnectionConfig, java.util.Map<String, String> configOptions)
     {
         this(databaseConnectionConfig, S3Client.create(), SecretsManagerClient.create(), AthenaClient.create(),
-                new GenericJdbcConnectionFactory(databaseConnectionConfig, DataLakeGen2MetadataHandler.JDBC_PROPERTIES,
+                new DataLakeGen2JdbcConnectionFactory(databaseConnectionConfig, DataLakeGen2MetadataHandler.JDBC_PROPERTIES,
```
ejeffrli: How come DataLakeGen2RecordHandler was using GenericJdbcConnectionFactory when there is a DataLakeGen2JdbcConnectionFactory class being used by MetadataHandler before?

athena-datalakegen2/src/main/java/com/amazonaws/athena/connectors/datalakegen2/DataLakeGen2Constants.java
```diff
@@ -26,5 +26,8 @@ public class DataLakeGen2Constants
     public static final int DEFAULT_PORT = 1433;
     public static final String QUOTE_CHARACTER = "\"";
 
+    //Property name used to set OAuth access token in JDBC connection properties
+    public static final String ACCESS_TOKEN_PROPERTY = "accessToken";
```
ejeffrli: will this property be used by other connectors that can use OAuth? Can we move this constant variable somewhere else so connectors like Synapse can use this (if the names are the same)?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/OAuthCredentialsProvider.java
```diff
@@ -0,0 +1,279 @@
...
+
+    protected OAuthCredentialsProvider(String secretName)
+    {
+        this(secretName, SecretsManagerClient.create(), HttpClient.newBuilder().build());
+    }
+
+    protected OAuthCredentialsProvider(String secretName, SecretsManagerClient secretsClient, HttpClient httpClient)
+    {
+        this.secretName = secretName;
+        this.secretsManager = new CachableSecretsManager(secretsClient);
```
ejeffrli: Can we re-use the secretsManager object from MetadataHandler and RecordHandler?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/OAuthCredentialsProvider.java
```diff
@@ -0,0 +1,279 @@
...
+    /**
+     * Maps OAuth access token to JDBC connection properties.
+     * Each provider can specify how the token should be used (e.g. as password, as property).
+     */
+    protected abstract Map<String, String> mapOAuthCredentials(String accessToken);
+
+    /**
+     * Checks if OAuth is configured by verifying required fields exist.
+     */
+    protected abstract boolean isOAuthConfigured(Map<String, String> secretMap);
```
ejeffrli, line 122: does this need to be abstract? Do two connectors check if oauth is configured in different ways?

athena-datalakegen2/src/main/java/com/amazonaws/athena/connectors/datalakegen2/DataLakeGen2Constants.java
burhan94: +1, lets think about moving this to the federation SDK since we're already extracting common OAuth code.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/OAuthCredentialsProvider.java
```diff
@@ -0,0 +1,279 @@
...
+    public static final String ACCESS_TOKEN = "access_token";
+    public static final String EXPIRES_IN = "expires_in";
+    public static final String FETCHED_AT = "fetched_at";
+
+    // Constants for OAuth configuration fields
+    public static final String CLIENT_ID = "client_id";
+    public static final String CLIENT_SECRET = "client_secret";
+
+    // Constants for basic authentication fields
+    public static final String USERNAME = "username";
```
burhan94: CredentialsProvider already exposes username, password

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/OAuthCredentialsProvider.java
```diff
@@ -0,0 +1,279 @@
...
+
+    protected OAuthCredentialsProvider(String secretName, SecretsManagerClient secretsClient, HttpClient httpClient)
+    {
+        this.secretName = secretName;
+        this.secretsManager = new CachableSecretsManager(secretsClient);
+        this.httpClient = httpClient;
+    }
+
+    @Override
+    public DefaultCredentials getCredential()
```
burhan94: We should improve the abstraction here. I suggest the following:

1. Credential interface -> UserNamePasswordCredential and OAuthAccessTokenCredential
2. OAuthCredentialsProvider returns a Credential object, and subclasses for OAuthCredentialsProvider will return the actual shape of the credential. A separate UserNamePasswordCredentialProvider should handle username/password separately. CredentialProvider can then be simplified

athena-datalakegen2/src/main/java/com/amazonaws/athena/connectors/datalakegen2/DataLakeGen2RecordHandler.java
Jithendar12: Hi @ejeffrli , this might be an oversight from the initial connector implementation. For Gen2, these changes weren’t strictly required because username/password were already being set in the JDBC properties ([ref](https://github.com/awslabs/aws-athena-query-federation/blob/e302bc0a378472299a0e8f8918fab32651cc2d2d/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/connection/GenericJdbcConnectionFactory.java#L86C1-L86C75)). So, deleted DataLakeGen2JdbcConnectionFactory file as it is no more required. Thanks!

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/OAuthCredentialsProvider.java
Jithendar12: Yes, this has to be abstract. For example SAP HANA uses client_id, client_secret and **token_url**.

athena-datalakegen2/src/main/java/com/amazonaws/athena/connectors/datalakegen2/DataLakeGen2Constants.java
Jithendar12: @ejeffrli / @burhan94, Created new CredentialsConstants class in SDK for different Authentication constants and moved to there. Thanks!

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/OAuthCredentialsProvider.java
Jithendar12: Yes, added getCachableSecretsManager() method to return same secretManger instance of MetadataHandler and RecordHandler.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/OAuthCredentialsProvider.java
Jithendar12: Updated code. Thank you!

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/OAuthCredentialsProvider.java
Jithendar12: @burhan94 ,
We have created a Credentials interface, which is implemented by both OAuthAccessTokenCredentials and the existing DefaultCredentials (for username/password).
We also created an OAuthCredentialsProvider that returns an OAuthAccessTokenCredentials object.
Instead of creating a new class for UserNamePasswordCredentialProvider, we have used the existing DefaultCredentialsProvider. Thank you!

burhan94:
Can you add some more tests? looks like code coverage check is failing

Jithendar12:
> Can you add some more tests? looks like code coverage check is failing

Hi @burhan94, thank you for reviewing the PR. Initially, we added unit tests to cover the major OAuth code. However, as the design was further updated, it resulted in multiple changes across classes. Unit Tests have been added to improve the coverage of these changed classes, including the scenarios for error handling.

burhan94 (approved):

athena-datalakegen2/src/main/java/com/amazonaws/athena/connectors/datalakegen2/DataLakeGen2CredentialProviderUtils.java
```diff
@@ -0,0 +1,73 @@
...
+import software.amazon.awssdk.services.glue.model.ErrorDetails;
+import software.amazon.awssdk.services.glue.model.FederationSourceErrorCode;
+
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * Utility class for handling credential provider functionality.
+ */
+public final class DataLakeGen2CredentialProviderUtils
```
ejeffrli: This class seems pretty generic. Seems like it returns oauthcredentialsprovider or returns defaultcredentialprovider. Any way to make this re-usable by synapse and other connectors who support oauth?

ejeffrli: I'm thinking this part can be abstracted out, and all connectors can eventually call this utils class to get default/oauth credentials
```
                if (DataLakeGen2OAuthCredentialsProvider.isOAuthConfigured(secretMap)) {
                    return new DataLakeGen2OAuthCredentialsProvider(secretName, secretMap, secretsManager);
                }
                ```

**Comment by ejeffrli** [comment]
Can you also provide an up-to-date testing document of 
```
1. Datalakegen2 with oauth
2. Datalakegen2 without oauth
3. Snowflake connector (as well as sqlserver, hive, etc. the more the better) (this is original non-oauth flow because it seems like DefaultCredentialsProvider was slightly touched)
```

File: athena-datalakegen2/src/main/java/com/amazonaws/athena/connectors/datalakegen2/DataLakeGen2CredentialProviderUtils.java (hunk 7 above)
Comment (Jithendar12) [reply athena-datalakegen2/src/main/java/com/amazonaws/athena/connectors/datalakegen2/DataLakeGen2CredentialProviderUtils.java]: Hi @ejeffrli , I've now moved this logic into the CredentialsProviderFactory class in the federation SDK module, making it reusable across all connectors that need OAuth support. Thank you!

**Comment by Jithendar12** [comment]
> Can you also provide an up-to-date testing document of
> 
> ```
> 1. Datalakegen2 with oauth
> 2. Datalakegen2 without oauth
> 3. Snowflake connector (as well as sqlserver, hive, etc. the more the better) (this is original non-oauth flow because it seems like DefaultCredentialsProvider was slightly touched)
> ```

Hi @ejeffrli , attached test documents to the PR. Thank you!

File: athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/CredentialsProviderFactory.java [hunk 8]
```diff
@@ -0,0 +1,101 @@
...
+     * @param secretName The name of the secret in AWS Secrets Manager
+     * @param secretsManager The secrets manager instance
+     * @param oAuthProviderClass The class of the OAuth provider to instantiate (must extend OAuthCredentialsProvider)
+     * @return A new CredentialsProvider instance based on the secret configuration
+     * @throws AthenaConnectorException if there are errors deserializing the secret or creating the provider
+     */
+    public static <T extends OAuthCredentialsProvider> CredentialsProvider createCredentialProvider(
+            String secretName,
+            CachableSecretsManager secretsManager,
+            Class<T> oAuthProviderClass)
```
AbdulR3hman: I really think we shouldn't use reflection when we have control over the lifecycle of these objects. here are my thoughts:

1. Current Issue:
   - The code unnecessarily uses reflection when we have full control over all intended classes
   - Passing class objects adds complexity without benefits in this controlled environment

2. Proposed Solution:
   - Replace reflection with an interface-based approach
   - Create an interface (e.g., `AthenaFederationCredentialProvider`) that all credential providers implement (if you find a better name go for it)
   - The interface should define the setup/initialization behavior

3. Implementation Suggestion:
   - Instead of:
     ```java
     Class<T> oAuthProviderClass // as parameter
     ```
   - Use:
     ```java
     AthenaFederationCredentialProvider provider // as parameter
     ```
   - Each provider (like DataLakeGen2CredentialsProvider) would implement this interface
   - The CredentialProviderFactory would work with the interface directly, eliminating the need for reflection

4. Benefits:
   - More straightforward and type-safe implementation
   - Better compile-time checking
   - Easier to maintain and extend
   - More explicit contract between components

ejeffrli: +1

Jithendar12: Hi @AbdulR3hman,Thank you for your suggestion. we’ve removed reflection and refactored to an interface-based approach (InitializableCredentialsProvider). All advanced providers (e.g., OAuthCredentialsProvider) will implement this interface, leveraging properties from Secrets Manager for initialization. Thank you!

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/CredentialsProviderFactory.java
```diff
@@ -50,40 +49,37 @@ private CredentialsProviderFactory()
...
+     * @param provider The InitializableCredentialsProvider instance to check and initialize
      * @return A new CredentialsProvider instance based on the secret configuration
      * @throws AthenaConnectorException if there are errors deserializing the secret or creating the provider
      */
-    public static <T extends OAuthCredentialsProvider> CredentialsProvider createCredentialProvider(
+    public static CredentialsProvider createCredentialProvider(
             String secretName,
             CachableSecretsManager secretsManager,
-            Class<T> oAuthProviderClass)
+            InitializableCredentialsProvider provider)
```
ejeffrli, line 59: is it necessary to create this new interface? Why not use OauthCredentialsProvider directly here and pass that in? Let, OauthCredentialsProvider class have a contract of isOauthConfigured and initialize.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/credentials/CredentialsProviderFactory.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Factory class for handling credentials provider creation.
+ * This class can be used by any connector that needs to support both
+ * OAuth and username/password authentication.
+ */
+public final class CredentialsProviderFactory
+{
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    private CredentialsProviderFactory()
```
ejeffrli, line 41: Could we let other classes create CredentialsProviderFactory objects? I think other classes should be creating CredentialsProviderFactory object. You should be able to create this class with your own providers.

For example,
factory = new CredentialsProviderFactory(secret, secretManager, OauthProvider);
factory.createCredentialProvider() 

maybe in future there are other constructor shapes such as 
factory = new CredentialsProviderFactory(secret, secretManager, OauthProvider, credProvider1, null);
factory.createCredentialProvider()  -> goes thru each provider provided when provider provided is != null.

checks within createCredentialProvider would look like
```
if this.oauthProvider != null:
  if this.oauthProvider.isOauthConfigured:
    this.oauthProvider.initialize
```

My idea is to make CredentialsProviderFactory.createCredentialsProvider be generic enough so all connectors can use this to get credentials. Having OauthCredentialsProvider (IntializableCredentialsProvider in your case) requirement in parameters makes this only work with oauth supported connectors.

ejeffrli, line 41: more of a dependency injection approach than a static factory function.

Jithendar12:
> more of a dependency injection approach than a static factory function.

Hi @ejeffrli, it seems you have deleted the two review comments above which are related to this comment. We just want to confirm is this still valid?

burhan94:
Met with Abdul and Jeff, we are good to approve.

PR #2948, opened 2025-08-20
update cloudera driver to address zookeeper CVE

aimethed:
*Issue #, if available:*

*Description of changes:*

## Issue Summary

Problem: Amazon Inspector detected a vulnerable version of Apache ZooKeeper (3.7.0) in your container images that use the AWS Athena Query Federation JAR files.

Root Cause: The vulnerable ZooKeeper 3.7.0 was embedded within the Cloudera HiveJDBC42 driver (version 2.6.23.1027) used by two Athena connectors:
• athena-cloudera-hive
• athena-hortonworks-hive

Why This Was Critical: ZooKeeper 3.7.0 contains known security vulnerabilities that could be exploited in production environments, making it a compliance and security risk.

Challenge: The ZooKeeper classes were shaded/embedded directly inside the third-party Cloudera JDBC driver JAR, so standard Maven dependency exclusions weren't sufficient.

Solution Applied:
1. Added Maven dependency exclusions for ZooKeeper from the HiveJDBC42 driver
2. Added explicit dependency on secure ZooKeeper 3.9.1 (version recommended by inspector)
3. Configured Maven shade plugin to exclude vulnerable ZooKeeper classes during JAR packaging
4. Rebuilt both connectors to replace 3.7.0 with 3.9.1

Result: Eliminated the security vulnerability while maintaining full functionality of the Hive connectors. Amazon Inspector will no longer flag this vulnerability in your images.

### DIFF
--- a/athena-cloudera-hive/src/main/java/com/amazonaws/athena/connectors/cloudera/HiveConstants.java
+++ b/athena-cloudera-hive/src/main/java/com/amazonaws/athena/connectors/cloudera/HiveConstants.java
@@ -37,7 +37,7 @@ public class HiveConstants
     static final int FETCH_SIZE = 1000;
     static final String ALL_PARTITIONS = "*";
     public static final String HIVE_NAME = "hive";
-    public static final String HIVE_DRIVER_CLASS = "org.apache.hive.jdbc.HiveDriver";
+    public static final String HIVE_DRIVER_CLASS = "com.cloudera.hive.jdbc.HS2Driver";
     public static final int HIVE_DEFAULT_PORT = 10000;
     public static final String METADATA_COLUMN_NAME = "col_name";
     public static final String METADATA_COLUMN_TYPE = "data_type";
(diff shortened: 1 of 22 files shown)


athena-cloudera-hive/src/main/java/com/amazonaws/athena/connectors/cloudera/HiveConstants.java
```diff
@@ -37,7 +37,7 @@ private HiveConstants()
     static final int FETCH_SIZE = 1000;
     static final String ALL_PARTITIONS = "*";
     public static final String HIVE_NAME = "hive";
-    public static final String HIVE_DRIVER_CLASS = "com.cloudera.hive.jdbc.HS2Driver";
+    public static final String HIVE_DRIVER_CLASS = "org.apache.hive.jdbc.HiveDriver";
```
burhan94: are there any backwards compatibility concerns here?

Trianz-Akshay:
We have tested PR#2948 for Cloudera Hive, and it is working as expected. Please find attached the test results document for reference.
As communicated earlier, we are unable to test on Hortonworks Hive since the required environment is not available.
[Cloudera_Hive_test_zookeeper_cve 1.docx](https://github.com/user-attachments/files/22041906/Cloudera_Hive_test_zookeeper_cve.1.docx)

athena-cloudera-hive/src/main/java/com/amazonaws/athena/connectors/cloudera/HiveConstants.java
aimethed: Approach was changed to continue using the same driver as before so shouldn't be relevant anymore

samarsar (approved):

burhan94 (approved):

PR #3007, opened 2025-09-17
[Fix] Include default truststore path when passing JAVA_TOOL_OPTIONS for Java 17 image

Jithendar12:
Issue #, if available:
**javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target**

The above error occurred while querying the DocumentDB database on Athena because passing JAVA_TOOL_OPTIONS (**--add-opens=java.base/java.nio=ALL-UNNAMED -XX:-TieredCompilation**) as a build argument in Java 17 was overriding the default truststore path (**-Djavax.net.ssl.trustStore=/var/lang/lib/security/cacerts**).

Description of changes:

Updated Dockerfiles to always prepend **-Djavax.net.ssl.trustStore=/var/lang/lib/security/cacerts** in JAVA_TOOL_OPTIONS instead of replacing it. Please find attached document.
[fix-java-17-ssl-handshake.odt](https://github.com/user-attachments/files/22388297/fix-java-17-ssl-handshake.odt)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of your choice.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,10 @@ ARG JAVA_VERSION=11
 # Use the specified version of Java
 FROM public.ecr.aws/lambda/java:${JAVA_VERSION}
 
-# Argument for Java tool options for ssl
-ARG JAVA_TOOL_OPTIONS="-Djavax.net.ssl.trustStore=/var/lang/lib/security/cacerts"
-# Set the JAVA_TOOL_OPTIONS environment variable for Java 17
-ENV JAVA_TOOL_OPTIONS=${JAVA_TOOL_OPTIONS}
+# Argument for Java tool options
+ARG JAVA_TOOL_OPTIONS=""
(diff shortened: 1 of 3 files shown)


Jithendar12:
Hi @fal-bharadwaj , thank you for reviewing.

1) Is this an issue only on DocDB and that's why other connectors do not need an update?
**This issue was reproduced while testing the DocDB connector built using Java 17. The Oracle connector and the root Dockerfiles also require this change, as they include the `-Djavax.net.ssl.trustStore=/var/lang/lib/security/cacerts` option, which has been updated in this PR. For other connectors, this change is not applicable.**

2) Can we update the unit test for this change?
**No, this is a Dockerfile-level change and does not involve main code that can be tested via unit tests.**

AbdulR3hman (approved):

PR #3008, opened 2025-09-17
Fixing error messages to not leak sensitive info

burhan94:
Addressing appsec review where error messages were potentially leaking info


macohen:
It's interesting that inspector doesn't pick this up. probably could use an OWASP detection action on PRs to catch stuff like this.

PR #3009, opened 2025-09-18
Optimize Snowflake jar size

Trianz-Akshay:
*Issue #, if available:*

*Description of changes:*
This PR the changes made to `athena-snowflake/pom.xml` to optimize the JAR size while maintaining functionality. The optimization process involved replacing the Maven Assembly Plugin with the Maven Shade Plugin and adding extensive dependency exclusions.
[Snowflake-jar-reduce.docx](https://github.com/user-attachments/files/22409201/Snowflake-jar-reduce.docx)
[SNOWFLAKE_FUNCTIONAL_TEST.csv](https://github.com/user-attachments/files/22409205/SNOWFLAKE_FUNCTIONAL_TEST.csv)

### DIFF
--- a/athena-snowflake/Dockerfile
+++ b/athena-snowflake/Dockerfile
@@ -8,9 +8,13 @@ ARG JAVA_TOOL_OPTIONS=""
 # Set the JAVA_TOOL_OPTIONS environment variable for Java 17
 ENV JAVA_TOOL_OPTIONS=${JAVA_TOOL_OPTIONS}
 
-# Copy function code and runtime dependencies from Maven layout
-COPY target/athena-snowflake-2022.47.1.jar ${LAMBDA_TASK_ROOT}
+# Copy pom.xml and JAR file
+COPY pom.xml target/athena-snowflake-*.jar /tmp/
 
-RUN jar xf athena-snowflake-2022.47.1.jar
+# Extract version and setup (remove JAR after extraction)
+RUN VERSION=$(grep -oP '(?<=<version>)[^<]+' /tmp/pom.xml | head -1) && \
+    echo "Extracted version: $VERSION" && \
+    jar xf /tmp/athena-snowflake-${VERSION}.jar && \
+    rm -f /tmp/pom.xml
 # Command can be overwritten by providing a different command in the template directly.
 # No need to specify here (already defined in .yaml file because legacy and connections use different)


athena-snowflake/Dockerfile
```diff
@@ -9,9 +9,8 @@ ARG JAVA_TOOL_OPTIONS=""
 ENV JAVA_TOOL_OPTIONS=${JAVA_TOOL_OPTIONS}
 
 # Copy function code and runtime dependencies from Maven layout
-COPY target/athena-snowflake.zip ${LAMBDA_TASK_ROOT}
-# Unpack the jar
-RUN jar xf athena-snowflake.zip
+COPY target/athena-snowflake-2022.47.1.jar ${LAMBDA_TASK_ROOT}
```
burhan94: this might break since the JAR version will match the release version AFAIK

Trianz-Akshay: Thanks for pointing this out. This is the same approach we’ve followed for other connectors where the JAR is consumed — the versioned artifact name matches the release version. For Snowflake, we’ve updated the logic to read the version directly from the Snowflake POM, so it should remain consistent and not cause issues.

athena-snowflake/pom.xml
```diff
@@ -19,22 +19,202 @@
...
+                    <artifactId>avatica-core</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>io.substrait</groupId>
+                    <artifactId>isthmus</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>io.substrait</groupId>
+                    <artifactId>core</artifactId>
+                </exclusion>
```
ZhitongYan, line 62: we might need subsrait dependencies for query federation

Trianz-Akshay: @ZhitongYan  Thanks for the review!
We have tested the Snowflake connector after these changes. Please find the test document attached in the PR description — everything is working as expected.
Kindly let us know if there are any specific scenarios we might have missed, so we can verify those cases where a Substrait dependency may be required.

ZhitongYan: https://github.com/awslabs/aws-athena-query-federation/pull/3044

Thanks for replying. The functionalities you tested should only within Athena Scope. There are more features supported in the current implementation. Looping @burhan94 in, as we cannot remove substrait / calcite dependencies due to this reason. 

Could you please help to test with PR #3044 changes to make sure there is no regression introduced? I have tested the changes, just want to double confirm.

Trianz-Akshay: Thanks for the clarification. @burhan94, could you please confirm if we should go ahead and test PR #3044 on our side as well?

Trianz-Akshay: Hi @burhan94,
We have tested PR [#3044](https://github.com/awslabs/aws-athena-query-federation/pull/3044) and found working as expected.
Please find attached test result.
[SNOWFLAKE_FUNCTIONAL_TEST.xlsx](https://github.com/user-attachments/files/22970525/SNOWFLAKE_FUNCTIONAL_TEST.xlsx)

burhan94:
closing in favor of #3044

PR #3016, opened 2025-09-26
DocumentDB substrait query plan usage and SubstraitFunctionParser updates

MitraAnanya-Dev:
*Description of changes:*

- As part of query federation feature we have added support for Substrait query plan in the DocumentDB connector to use the plan and execute the queries.

- We have also made updates in Athena Federation SDK in the SubstraitFunctionParser in order to add additional support for the following:
    - OR semantics was not preserved for the connector to consume and differentiate between AND and OR functions. Support has been added to maintain tree hierarchy and not return flattened lists.
    - Support for NOT operations such as NOR (NOT + OR) and NAND (NOT + AND) has been added for connector consumption.

- **Existing methods unchanged**: All current functionality preserved
- **Additive enhancement**: New method doesn't affect existing code paths

We have made sure that these are backwards compatible.


athena-docdb/pom.xml
```diff
@@ -34,6 +34,12 @@
             <artifactId>docdb</artifactId>
             <version>${aws-sdk-v2.version}</version>
             <scope>test</scope>
+            <exclusions>
```
DilipReddyGaddam, line 30: why are we excluding this ?

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBRecordHandler.java
```diff
@@ -127,74 +141,147 @@ private static Map<String, Object> documentAsMap(Document document, boolean case
...
+            logger.info("Using Substrait query plan for optimization");
+        }
+        else {
+            hasQueryPlan = false;
+            plan = null;
+            logger.info("No Substrait query plan available, using constraint-based filtering");
+        }
+
+        // ---------------------- LIMIT pushdown support ----------------------
+        final Pair<Boolean, Integer> limitPair = getLimit(plan, recordsRequest.getConstraints());
```
DilipReddyGaddam: getLimit logic and getSort logic looks common for most of the connectors. Can we move this logic to MetadataHandler.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
```diff
@@ -248,6 +268,343 @@ public static Document parseFilter(String filter)
...
+        List<String> tableColumns = SubstraitMetadataParser.getTableColumns(substraitRelModel);
+
+        return SubstraitFunctionParser.getColumnPredicatesMap(
+                extensionDeclarations,
+                substraitRelModel.getFilterRel().getCondition(),
+                tableColumns);
+    }
+
+    /**
+     * Enhanced query builder that tries tree-based approach first, then falls back to flattened approach
```
DilipReddyGaddam: Can you please add an example of how flattened approach query looks like. Does both queries result in consistent results ?

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
```diff
@@ -248,6 +268,343 @@ public static Document parseFilter(String filter)
...
+        }
+
+        // Apply the logical operator to combine child documents
+        // Example: AND → {"$and": [child1, child2]}, OR → {"$or": [child1, child2]}
+        return switch (expression.getOperator()) {
+            case AND -> new Document(AND_OP, childDocuments); // {"$and": [{"col1": "val1"}, {"col2": "val2"}]}
+            case OR -> new Document(OR_OP, childDocuments);   // {"$or": [{"col1": "val1"}, {"col2": "val2"}]}
+            default ->
+                // For unknown operators, default to OR for safety
+                    new Document(OR_OP, childDocuments);
```
DilipReddyGaddam: should we throw exception when an unknown logic operator is used ? query results from using OR operator might not be correct.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
```diff
@@ -248,6 +268,343 @@ public static Document parseFilter(String filter)
...
+                    Document nandCondition = new Document(NOR_OP, Collections.singletonList(new Document(AND_OP, andConditions)));
+                    List<Document> nandFinalConditions = new ArrayList<>();
+                    
+                    // Add null exclusion for each column involved in NAND operation
+                    for (String col : nandColumns) {
+                        nandFinalConditions.add(new Document(col, isNotNullPredicate()));
+                    }
+                    nandFinalConditions.add(nandCondition);
+                    return new Document(AND_OP, nandFinalConditions);
+                case NOR:
```
DilipReddyGaddam, line 463: Can a NAND oeprator or NOR operator combined with any other operators ?

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
```diff
@@ -248,6 +268,343 @@ public static Document parseFilter(String filter)
...
+        else if (equalValues.size() == 1) {
+            Object eqValue = equalValues.get(0);
+            Document equalPredicate;
+            if (column.equals(COLUMN_NAME_ID)) {
+                equalPredicate = new Document(EQ_OP, new ObjectId(eqValue.toString()));
+            }
+            else {
+                equalPredicate = new Document(EQ_OP, eqValue);
+            }
+            if (!otherPredicates.isEmpty()) {
```
DilipReddyGaddam, line 511: This common logic is used in multiple places refactor it. one more question in the case of equal to condition on a column can we have different filter operators on the same column

```
List<Document> andConditions = new ArrayList<>();
                andConditions.add(new Document(column, equalPredicate));
                for (Document otherPred : otherPredicates) {
                    andConditions.add(new Document(column, otherPred));
                }
                return new Document(AND_OP, andConditions);
```

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
```diff
@@ -248,6 +268,343 @@ public static Document parseFilter(String filter)
...
+        // Handle non-EQUAL predicates with special null exclusion for NOT_EQUAL operations
+        // NOT_EQUAL operations should exclude records where the field is null to match SQL semantics
+        else if (!otherPredicates.isEmpty()) {
+            // Check if any predicate is NOT_EQUAL - these need null exclusion
+            // Example: "column <> 'value'" should not match records where column is null
+            boolean hasNotEqual = otherPredicates.stream()
+                    .anyMatch(doc -> doc.containsKey(NOT_EQ_OP));
+
+            if (hasNotEqual && otherPredicates.size() == 1) {
+                // Single NOT_EQUAL case - wrap with null exclusion
```
DilipReddyGaddam, line 526: Is this functionality consistent with existing connector functionality ?

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
```diff
@@ -248,6 +268,343 @@ public static Document parseFilter(String filter)
...
+                    }
+                    else {
+                        // Keep other predicates unchanged (GREATER_THAN, LESS_THAN, etc.)
+                        // Example: {"column": {"$gt": 5}} remains as-is
+                        orConditions.add(new Document(column, predicate));
+                    }
+                }
+                // Combine all conditions with OR
+                // Example: {"$or": [{"$and": [null_check, not_equal]}, {"column": {"$gt": 5}}]}
+                return new Document(OR_OP, orConditions);
```
DilipReddyGaddam: why did we choose OR operation here ?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/model/Operator.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package com.amazonaws.athena.connector.substrait.model;
+
+/**
+ * Represents a subset of Substrait-supported operators for query federation.
+ * This enum contains only the commonly used operators that are supported
+ * by the Athena Query Federation framework. The full Substrait specification
+ * includes many more operators, We will extend more operators as we need.
+ */
+public enum Operator
```
DilipReddyGaddam: why do you need new enum for this. Can add new functionality to athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/model/SubstraitOperator.java

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitFunctionParser.java
```diff
@@ -201,38 +280,29 @@ private static boolean isLogicalOperator(String functionName)
...
-     * This method is mapping only small set of operators, and we will extend this as we need.
+     * This method supports comparison operators, logical operators, null checks, and the NOT operator.
+     * The mapping will be extended as additional operators are needed.
      * 
-     * @param functionName The Substrait function name (e.g., "gt:any_any", "equal:any_any")
+     * @param functionName The Substrait function name (e.g., "gt:any_any", "equal:any_any", "not:bool")
      * @return The corresponding Operator enum value
      * @throws UnsupportedOperationException if the function name is not supported
      */
     private static SubstraitOperator mapToOperator(String functionName)
```
DilipReddyGaddam, line 306: I don't think federation supports java17 to use this switch case. Please make sure your all the PR checks and build are successful

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandler.java
```diff
@@ -157,6 +161,30 @@ public GetDataSourceCapabilitiesResponse doGetDataSourceCapabilities(BlockAlloca
...
+        supportedFunctions.add(StandardFunctions.EQUAL_OPERATOR_FUNCTION_NAME);
+        supportedFunctions.add(StandardFunctions.GREATER_THAN_OPERATOR_FUNCTION_NAME);
+        supportedFunctions.add(StandardFunctions.LESS_THAN_OPERATOR_FUNCTION_NAME);
+        supportedFunctions.add(StandardFunctions.GREATER_THAN_OR_EQUAL_OPERATOR_FUNCTION_NAME);
+        supportedFunctions.add(StandardFunctions.LESS_THAN_OR_EQUAL_OPERATOR_FUNCTION_NAME);
+        supportedFunctions.add(StandardFunctions.NOT_EQUAL_OPERATOR_FUNCTION_NAME);
+
+        // To check for $nin and $nor
+
+        capabilities.put(DataSourceOptimizations.SUPPORTS_COMPLEX_EXPRESSION_PUSHDOWN.withSupportedSubTypes(
```
burhan94, line 211: have you validated that documentDB supports pushing down all of the standard functions defined above?

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBRecordHandler.java
```diff
@@ -127,74 +141,147 @@ private static Map<String, Object> documentAsMap(Document document, boolean case
...
 
-        MongoClient client = getOrCreateConn(recordsRequest.getSplit());
-        MongoDatabase db;
-        MongoCollection<Document> table;
+        final MongoClient client = getOrCreateConn(recordsRequest.getSplit());
+        final MongoDatabase db;
+        final MongoCollection<Document> table;
         Document query;
 
+        // ---------------------- Substrait Plan extraction ----------------------
```
burhan94, line 165: can we move this logic to a util function or a new class that returns the desired predicates with the constraints object as the input?

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
burhan94: we cannot fallback to a default here. If we ever get to a point where an operator is pushed down to this layer that is not applicable to the source, then we messed up in communicating the capabilities to the engine.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
```diff
@@ -248,6 +268,343 @@ public static Document parseFilter(String filter)
...
+     */
+    private static Document convertColumnPredicatesToDoc(String column, List<ColumnPredicate> colPreds)
+    {
+        if (colPreds == null || colPreds.isEmpty()) {
+            return new Document();
+        }
+        // Special handling for universal constraints (none / all)
+        for (ColumnPredicate pred : colPreds) {
+            if (pred.getOperator() == SubstraitOperator.IS_NULL) {
+                return documentOf(column, isNullPredicate());
```
burhan94: this doesn't make sense. This code suggests that isNull and isNotNull can't have any other predicates defined on that column. I don't think that is true. If it is true, then you need to validate that List<ColumnPredicate> colPreds is a single value list, otherwise you will be dropping predicates.

athena-docdb/src/test/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandlerTest.java
```diff
@@ -64,11 +57,7 @@
 import software.amazon.awssdk.services.glue.GlueClient;
 import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
+import java.util.*;
```
burhan94: no * imports.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/Block.java
```diff
@@ -179,6 +179,34 @@ public boolean offerValue(String fieldName, int row, Object value)
...
+     * @param fieldName The name of the field you wish to write to.
+     * @param row The row number to write to. Note that Apache Arrow Blocks begin with row 0 just like a typical array.
+     * @param value The value you wish to write.
+     * @param hasQueryPlan Whether the operation is running under a query plan, if true, bypasses constraint checks.
+     * @return True if the value was written to the Block (even if the field is missing from the Block),
+     * False if the value was not written due to failing a constraint or if query plan is present.
+     * @note This method will take no action if the provided fieldName is not a valid field in this Block's Schema.
+     * In such cases the method will return true.
+     */
+    public boolean offerValue(String fieldName, int row, Object value, boolean hasQueryPlan)
```
burhan94, line 198: why is hasQueryPlan needed here and why are we bypassing the constraint check?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitFunctionParser.java
```diff
@@ -201,38 +280,29 @@ private static boolean isLogicalOperator(String functionName)
...
-                return SubstraitOperator.IS_NOT_NULL;
-            case "and:bool":
-                return SubstraitOperator.AND;
-            case "or:bool":
-                return SubstraitOperator.OR;
-            default:
-                throw new UnsupportedOperationException("Unsupported operator function: " + functionName);
-        }
+        return switch (functionName) {
+            case "gt:any_any", "gt:pts_pts" -> SubstraitOperator.GREATER_THAN;
```
burhan94: what is pts_pts?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitFunctionParser.java
```diff
@@ -85,6 +86,15 @@ public static List<ColumnPredicate> parseColumnPredicates(List<SimpleExtensionDe
             return columnPredicates;
         }
 
+        // Handle NOT unary operator
+        if ("not:bool".equals(functionInfo.getFunctionName())) {
```
burhan94: please make not:bool and other function names into static vars, or even better yet, extract them into a common class.

athena-docdb/pom.xml
MitraAnanya-Dev: Post previous Substrait merges we were facing a conflict while testing the connector for the documentDB default URL not being recognized and was attached to an error related to this netty package. Applying the exclusion that other connectors like Snowflake, SAP that also have this netty-nio-client excluded had helped resolve. Retested without the exclusion, seems functional. Hence removed it.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBRecordHandler.java
MitraAnanya-Dev: I suppose it won't be accessible from MetadataHandler. I shifted it to RecordHandler and put the business logic in a utility class.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
MitraAnanya-Dev: Flattened path gives results aligned to tree based except for OR, NOR, etc. We had applied a fix for this in the later phases for this in the SDK and put the connector updates via the tree based approach and retained the flattened approach just as a safety net only. It's highly unlikely we'll ever have to fall back, hence, removing it. Now if for any reason the filter fails, I'm returning an empty document to match Constraint based approach.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
MitraAnanya-Dev: Yes it can be combined. Example, a query like this works: SELECT * FROM table WHERE NOT (job_title='Lead Engineer' OR job_title='Project Lead') AND employee_name !='Null'

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
MitraAnanya-Dev: - Optimized. 
- Yes we can have it. It's tested.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
MitraAnanya-Dev: Yes this Null check for != operation had to be specifically introduced to match the result against Constraints.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
MitraAnanya-Dev: - This code path handles the fallback flattened approach where Substrait expressions like column = 'A' OR column > 10 get parsed into a flat list of predicates: [EQUAL('A'), GREATER_THAN(10)] for the same column.
- Original SQL semantics: When multiple predicates on the same column come from the flattened approach, they originated from OR expressions in the SQL Mixed operator handling: Unlike EQUAL predicates (which get optimized to $in), mixed operators like EQUAL + GREATER_THAN must use explicit $or.
- The tree-based approach (preferred path) handles OR correctly at the logical level which is now, the only route to build queries.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/model/Operator.java
MitraAnanya-Dev: Remained as an unused class from previous merges. Deleted.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitFunctionParser.java
MitraAnanya-Dev: PR checks and builds are successful. However, still converted them to Java 11 compatible syntax.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandler.java
MitraAnanya-Dev: Yes we've only included what's present from the supportedFunctions list.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBRecordHandler.java
MitraAnanya-Dev: - Both Substrait and Constraints (existing connector code) processing are part of the same input handling flow:
- recordsRequest.getConstraints().getQueryPlan() (Substrait)
- recordsRequest.getConstraints().getSummary() (Constraints) Since they serve the same purpose, if we extract Substrait logic for reusability, we would also need to extract the Constraints logic for consistency, i suppose?
- Both are:
1. Input processing from the same source object
2. Connector-specific in their usage
3. Part of the same decision tree (Substrait → fallback to Constraints)
- And we followed the design other Athena connectors are following for consuming Query Plan such as DynamoDB.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
MitraAnanya-Dev: Throwing UnsupportedOperationException now.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/QueryUtils.java
MitraAnanya-Dev: - For Substrait, we're using the tree-Based approach first via makeQueryFromLogicalExpression(logicalExpr).
- If this fails only then do we fall back to a flattened approach.
- It follows this path:
1. parseLogicalExpression() creates a tree: OR(IS_NULL(rate), EQUAL(rate, 404458.840))
2. makeQueryFromLogicalExpression() processes each leaf individually.
- So,
1. Substrait creates a logical OR tree
2. Tree approach processes each predicate individually
3. convertColumnPredicatesToDoc() never receives mixed predicates

- Example: For a query like select * from table where rate is null or rate = '404458.840' it will generate the final Mongo query as {$or=[{rate={$exists=true, $eq=null}}, {rate={$eq=404458.840}}]}. So it isn't being dropped. 

- Have removed the fallback flattened approach

athena-docdb/src/test/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandlerTest.java
MitraAnanya-Dev: Refactored

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/Block.java
MitraAnanya-Dev: We identified a scenario that when both Constraints and Substrait objects are present in a request, they may contain conflicting table, column, or filter information. Although Substrait takes precedence over Constraints, the existing constraintEvaluator validation still runs and fails due to these mismatches. To prevent this issue, we are bypassing Constraint validation entirely when a Substrait (query plan) is present, since the Constraint validation becomes unnecessary.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitFunctionParser.java
MitraAnanya-Dev: this is the function name that substrait generates when comparative operations are applied on timestamp fields

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitFunctionParser.java
MitraAnanya-Dev: Created SubstraitFunctionNames to hold the function names.

AbdulR3hman:
can we please move out all the changes to the SDK from this pr and PR-3134 into its own PR only for the SDK changes?

MitraAnanya-Dev:
> can we please move out all the changes to the SDK from this pr and PR-3134 into its own PR only for the SDK changes?

@AbdulR3hman 
Created https://github.com/awslabs/aws-athena-query-federation/pull/3137 to split SDK changes from this PR.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandler.java
AbdulR3hman: Looking at this again; I think these are only supported in case of managed connector and not in the normal constraint code path. we need to rethink about how we are approaching this.

AbdulR3hman:
closing this for the same of; https://github.com/awslabs/aws-athena-query-federation/pull/3229 | It rebases it properly as well as refactor the code so it doesn't include redundant work.

PR #3064, opened 2025-10-13
Handle KMS and DDB NotFoundExceptions by throwing AthenaConnectorException

suyinlee:
*Issue #, if available:*

*Description of changes:*
Handle KMS and DDB NotFoundExceptions by throwing AthenaConnectorException


burhan94:
for context, this issue was discovered when DNA team was getting runtime exceptions from the lambda. There was some confusion about what types of exceptions were thrown from the lambda, and if we could rewrap the runtime exceptions into something else to help with the DNA team's operations. I spent some time with the team diving into this and showed them that everything should be wrapped already with an AthenaConnectorException, but looks like there were some missing cases that fell through. This PR fixes some of those missing cases.

ejeffrli (approved):

PR #3065, opened 2025-10-13
[Fix] Casing mode ANNOTATION error in Snowflake and Saphana.

ritiktrianz:
*Issue #, if available:*
When configuring the Snowflake/SAP HANA connectors from the SAR, tables cannot be listed, and an error is thrown.

*Description of changes:*
Updated non-glue default casing mode to NONE from ANNOTATION in constructor. PFA analysis and testing document for reference.
[saphana-snowflake-casing-mode-issue.odt](https://github.com/user-attachments/files/22883622/saphana-snowflake-casing-mode-issue.odt)
[fix-annotation-casing.docx](https://github.com/user-attachments/files/22969982/fix-annotation-casing.docx)


athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/resolver/SnowflakeJDBCCaseResolver.java
```diff
@@ -38,7 +38,7 @@ public class SnowflakeJDBCCaseResolver
 
     public SnowflakeJDBCCaseResolver(String sourceType)
     {
-        super(sourceType, FederationSDKCasingMode.ANNOTATION, FederationSDKCasingMode.NONE);
+        super(sourceType, FederationSDKCasingMode.CASE_INSENSITIVE_SEARCH, FederationSDKCasingMode.NONE);
```
burhan94: default should be NONE.

ritiktrianz: Hi @burhan94 , thank you for your review. As suggested, we have updated the default casing mode to NONE.

athena-saphana/src/main/java/com/amazonaws/athena/connectors/saphana/resolver/SaphanaJDBCCaseResolver.java
```diff
@@ -38,7 +38,7 @@ public class SaphanaJDBCCaseResolver
 
     public SaphanaJDBCCaseResolver(String sourceType)
     {
-        super(sourceType, FederationSDKCasingMode.ANNOTATION, FederationSDKCasingMode.NONE);
+        super(sourceType, FederationSDKCasingMode.NONE, FederationSDKCasingMode.NONE);
```
fal-bharadwaj, line 41: I have a couple of overall high level questions as i may be missing some context - 

The testing documents attached have the proposed solution for CASE_INSENSITIVE_SEARCH, have we done similar testing for NONE? I see that "UPPER CASE SCHEMA/TABLE" testing for "NONE" failed from the attachment, is that expected?

And are their any guidance for customers for backward compatibility (or I guess any cases where this might break for existing customers?)

chngpe: >I see that "UPPER CASE SCHEMA/TABLE" testing for "NONE" failed from the attachment, is that expected?

Yes, we did similar testing, "UPPER CASE SCHEMA/TABLE" testing for "NONE"  will failed as expected as if data source is default/configured UPPER CASED, the connector will failed to resolve the entity name dude to not match. Hence suggest customer to either configure federated source with lower case only or using the casing mode to specify the desired behavior for connector.


>And are their any guidance for customers for backward compatibility (or I guess any cases where this might break for existing customers?)

We are in the process of updating doc. We are not removing annotation support for existing.... The annotation has its limitation hence you see the exception from the attachment(existing behavior just clear error message from last update)

PR #3103, opened 2025-11-07
Updating EncryptionKeyFactory to add overirde AWS request configuration in KMS calls made generate encryption key

DilipReddyGaddam:
… calls

*Issue #, if available:*

*Description of changes:*
Added new methods in EncryptionKey Factory to override default credentials used in KMS request. This will help managed connector to use customer credentials while generating encryption key.

Validated the changes by deploying the DDB connector locally.

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/KmsKeyFactory.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/security/KmsKeyFactory.java
@@ -79,13 +79,9 @@ public class KmsKeyFactory
             throw new AthenaConnectorException(e.getMessage(), ErrorDetails.builder().errorCode(FederationSourceErrorCode.ENTITY_NOT_FOUND_EXCEPTION.toString()).build());
         }
 
-        GenerateRandomRequest.Builder randomBuilder = GenerateRandomRequest.builder()
-                .numberOfBytes(AesGcmBlockCrypto.NONCE_BYTES);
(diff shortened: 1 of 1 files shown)


burhan94 (approved):

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/DynamoDBMetadataHandler.java
```diff
@@ -559,6 +559,8 @@ public GetSplitsResponse doGetSplits(BlockAllocator allocator, GetSplitsRequest
             logger.info("QPT Split Requested");
             return setupQueryPassthroughSplit(request);
         }
+        FederatedIdentity federatedIdentity = request.getIdentity();
+        AwsRequestOverrideConfiguration overrideConfig = getRequestOverrideConfig(federatedIdentity.getConfigOptions());
```
ejeffrli, line 563: Will overrideConfig always be populated? How does this function work from athena view where we dont have FAS_TOKEN and SERVICE_KMS defined in configOptions

DilipReddyGaddam: If the required parameters are not present getRequestOverrideConfig() will return null and makeEncryptionKey method handles null values. This is backwards compatible.

ejeffrli (approved):

burhan94 (approved):

PR #3108, opened 2025-11-19
Fix epoch date conversion correctness issue when machine time zone is not in UTC

chngpe:
*Issue #, if available:*
For BigQuery, snowflake connector when predicate date is `19358` (2023-01-01) and the TZ at GMT -5(or any negative), the logic will parse it as (2022-12-31).  Or if date is (2023-12-31) and the TZ at GMT +5 (or any positive), it will result in 2024-01-01.

*Description of changes:*
Treat epoch date received as UTC, do offset on TZ.


fal-bharadwaj (approved):

chrlso:
Add some description of the issue and what we are fixing.

PR #3122, opened 2025-11-25
Adding change for big query substrait plan

JawedRakib:
Description of changes:

As part of query federation feature we have added support for Substrait query plan in the GoogleBigQuery connector to use the plan and execute the queries.

We have also made updates in Athena Federation SDK in the SubstraitFunctionParser in order to add additional support for the following:

Existing methods unchanged: All current functionality preserved

Additive enhancement: New method doesn't affect existing code paths

We have made sure that these are backwards compatible.

### DIFF
--- a/athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandler.java
+++ b/athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandler.java
@@ -24,6 +24,7 @@ import com.amazonaws.athena.connector.lambda.data.BlockAllocator;
 import com.amazonaws.athena.connector.lambda.data.BlockWriter;
 import com.amazonaws.athena.connector.lambda.domain.Split;
 import com.amazonaws.athena.connector.lambda.domain.TableName;
+import com.amazonaws.athena.connector.lambda.domain.predicate.functions.StandardFunctions;
 import com.amazonaws.athena.connector.lambda.domain.spill.SpillLocation;
 import com.amazonaws.athena.connector.lambda.handlers.GlueMetadataHandler;
 import com.amazonaws.athena.connector.lambda.metadata.GetDataSourceCapabilitiesRequest;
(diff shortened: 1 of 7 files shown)


AbdulR3hman:
hi; this change is touching too many packages at the same time; reverting it in case of a bug will be really problematic; can you please split them out to three if possible; or at least the changes to the two different plugins to be done separately. thanks. 

In the mean time I'll do a quick review here.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/Block.java
```diff
@@ -179,6 +179,34 @@ public boolean offerValue(String fieldName, int row, Object value)
...
+     * @param value The value you wish to write.
+     * @param hasQueryPlan Whether the operation is running under a query plan, if true, bypasses constraint checks.
+     * @return True if the value was written to the Block (even if the field is missing from the Block),
+     * False if the value was not written due to failing a constraint or if query plan is present.
+     * @note This method will take no action if the provided fieldName is not a valid field in this Block's Schema.
+     * In such cases the method will return true.
+     */
+    public boolean offerValue(String fieldName, int row, Object value, boolean hasQueryPlan)
+    {
+        if (!hasQueryPlan && !constraintEvaluator.apply(fieldName, value)) {
```
AbdulR3hman, line 200: should this be an `||`?

AbdulR3hman: the reason I'm asking is that all other methods here check only for `constraintEvaluator.apply` being true; I guess why the new condition is needed?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandler.java
```diff
@@ -100,6 +101,7 @@ public final void handleRequest(InputStream inputStream, OutputStream outputStre
         try (BlockAllocatorImpl allocator = new BlockAllocatorImpl()) {
             int resolvedSerDeVersion = SerDeVersion.SERDE_VERSION;
             byte[] allInputBytes = com.google.common.io.ByteStreams.toByteArray(inputStream);
+            logger.info("allInputBytes: '{}'", new String(allInputBytes, StandardCharsets.UTF_8));
```
AbdulR3hman, line 104: should this be debug to reduce noise?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/Block.java
AbdulR3hman: just finished going through the unit test; it seems you only want this method to work with a plan and not constraint object. my concern is that we are expecting other methods to work with both except for this one method,

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandler.java
```diff
@@ -103,6 +116,14 @@ public class DocDBMetadataHandler
     // used to filter out Glue databases which lack the docdb-metadata-flag in the URI.
     private static final DatabaseFilter DB_FILTER = (Database database) -> (database.locationUri() != null && database.locationUri().contains(DOCDB_METADATA_FLAG));
 
+    private static final String SECRET_ARN_KEY = "secret_arn";
+    private static final String AUTH_DB_KEY = "AUTHENTICATION_DATABASE";
+    
+    // JSON credential field names
```
AbdulR3hman, line 122: should all these be here; `athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBEnvironmentProperties.java` ?

PR #3124, opened 2025-11-26
Snowflake export update

chngpe:
-add support for specify storage integration
-update partition logic for S3Export path
-Introduce sql string method in SQL builder, remove duplicate logic in snowflake SQL string builder.
-fix query failed when contains predicate
-fix query return empty result when multi sources within same query
-fix support for datetime, timestamp and datetimemilli predicate on s3Export path

*Description of changes:*

### DIFF
--- a/tools/bump_versions/bump_connectors_version.py
+++ b/tools/bump_versions/bump_connectors_version.py
@@ -13,7 +13,8 @@ import common
 EXCLUDE_DIRS={
     "athena-federation-sdk",
     "athena-federation-integ-test",
-    "athena-federation-sdk-tools"
+    "athena-federation-sdk-tools",
+    "athena_federation_testing"
 }
 
 
--- a/tools/bump_versions/bump_sdk_version.py
+++ b/tools/bump_versions/bump_sdk_version.py
@@ -35,7 +35,7 @@ if __name__ == "__main__":
         common.output_xml(soup, f"{sdk_dir}/pom.xml")
 
     # This incluldes the root project
-    all_projects = glob.glob("athena*") + ["."]
+    all_projects = glob.glob("athena-*") + ["."]
 
     sdk_project_artifact_ids = common.get_projects_artifact_ids_map(SDK_DIRS)
 


athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeQueryStringBuilder.java
```diff
@@ -86,221 +76,97 @@ protected List<String> getPartitionWhereClauses(Split split)
...
-        List<String> conjuncts = new ArrayList<>();
-        for (Field column : columns) {
-            ArrowType type = column.getType();
-            if (constraints.getSummary() != null && !constraints.getSummary().isEmpty()) {
-                ValueSet valueSet = constraints.getSummary().get(column.getName());
-                if (valueSet != null) {
-                    conjuncts.add(toPredicate(column.getName(), valueSet, type, accumulator));
-                }
-            }
+        if (sql == null) {
```
AbdulR3hman: nit; shouldn't this be null or empty?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeQueryStringBuilder.java
```diff
@@ -63,6 +52,7 @@ public class SnowflakeQueryStringBuilder
     private static final Logger LOGGER = LoggerFactory.getLogger(SnowflakeQueryStringBuilder.class);
     private static final String quoteCharacters = "\"";
```
AbdulR3hman: nit: I know those originally from previous code; but since they are static and final; can you capitalize them please?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeQueryStringBuilder.java
```diff
@@ -63,6 +52,7 @@ public class SnowflakeQueryStringBuilder
...
     private static final String singleQuoteCharacters = "\'";
```
AbdulR3hman: same here.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/SupportedTypes.java
```diff
@@ -52,6 +52,7 @@ public enum SupportedTypes
     VARCHAR(Types.MinorType.VARCHAR),
     STRUCT(Types.MinorType.STRUCT),
     LIST(Types.MinorType.LIST),
+    TIMESTAMPNTZ(Types.MinorType.TIMESTAMPMILLI),
```
AbdulR3hman: q: is this going to be true across all connectors? can we test other connectors to make sure before/after is reflecting correctly?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/BlockAllocatorImpl.java
```diff
@@ -128,7 +128,9 @@ public synchronized Block createBlock(Schema schema)
         List<FieldVector> vectors = new ArrayList();
         try {
             for (Field next : schema.getFields()) {
-                vectors.add(next.createVector(rootAllocator));
+                FieldVector vector = next.createVector(rootAllocator);
+                vector.allocateNew();
```
AbdulR3hman, line 132: discussing this offline; but just for the sake of keeping this PR in the loop; is `allocateNew()` required now in the newer version? Since we never actually had to do this before? and is the default allocation good enough initially?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/S3BlockSpiller.java
```diff
@@ -202,10 +202,6 @@ public void writeRows(RowWriter rowWriter)
             throw (ex instanceof RuntimeException) ? (RuntimeException) ex : new RuntimeException(ex);
         }
 
-        if (rows > maxRowsPerCall) {
```
AbdulR3hman, line 207: why is this no longer needed? or better question; why did we need `maxRowsPerCall` initially and now we don't?

AbdulR3hman: and it seems to be used only within this method; if not needed should we clean up the rest of the class and the constructor? but only after we answer the first question; i.e. why we no longer need max rows per call

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/utils/SnowflakeArrowTypeConverter.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Coverts Jdbc data type to Arrow data type.
+     *
+     * @param jdbcType Jdbc integer type. See {@link java.sql.Types}.
+     * @param precision Decimal precision.
+     * @param scale Decimal scale.
+     * @return Arrow type. See {@link ArrowType}.
+     */
+    public static Optional<ArrowType> toArrowType(String name, final int jdbcType, final int precision, final int scale, java.util.Map<String, String> configOptions)
```
AbdulR3hman: this seems to be more or less the exact copy of `athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcArrowTypeConverter.java` 

I know we are trying to handle special snow flake type; can't we just add it in the original JDBC class instead? that way we don't have to maintain two?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -115,13 +114,18 @@
...
 public class SnowflakeMetadataHandler extends JdbcMetadataHandler
 {
-    static final Map<String, String> JDBC_PROPERTIES = ImmutableMap.of("databaseTerm", "SCHEMA", "CLIENT_RESULT_COLUMN_CASE_INSENSITIVE", "true");
     private static final Logger LOGGER = LoggerFactory.getLogger(SnowflakeMetadataHandler.class);
     private static final String COLUMN_NAME = "COLUMN_NAME";
     private static final String EMPTY_STRING = StringUtils.EMPTY;
-    public static final String SEPARATOR = "/";
-    static final String BLOCK_PARTITION_COLUMN_NAME = "partition";
     private static final int MAX_SPLITS_PER_REQUEST = 1000_000;
+    private static final String STORAGE_INTEGRATION_CONFIG_KEY = "snowflake_storage_integration_name";
```
AbdulR3hman: should all these be moved to the Constants for the connector?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -798,36 +704,119 @@ private boolean checkForView(TableName tableName) throws Exception
...
+        }
+
+        return bucketPath;
+    }
+
+    /**
+     * Get Snowflake storage integration name from config
+     * @return
+     */
+    private String getStorageIntegrationName()
```
AbdulR3hman, line 763: qq; what exactly is this used for? *(still going through the PR so it might be answered in this class);

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -798,36 +704,119 @@ private boolean checkForView(TableName tableName) throws Exception
...
-            GetFunctionRequest request = GetFunctionRequest.builder()
-                    .functionName(functionName)
-                    .build();
+        return null;
+    }
 
-            GetFunctionResponse response = lambdaClient.getFunction(request);
-            return response.configuration().role();
+    @VisibleForTesting
+    Map<String, String> getStorageIntegrationProperties(Connection connection, String integrationName) throws SQLException
```
AbdulR3hman: Just couple of thoughts on `getStorageIntegrationProperties` and `isSFStorageIntegrationExistAndValid`;

I think we can make `isSFStorageIntegrationExistAndValid` return an optional; that we we don't have to try to get the same environment properties twice calling this in both methods;

```
Optional.ofNullable(properties.get(STORAGE_INTEGRATION_BUCKET_KEY))
```

I'd write it something like this; 

```
private String requireProperty(Map<String, String> properties, String key)
{
    return Optional.ofNullable(properties.get(key))
            .orElseThrow(() -> new IllegalArgumentException(
                    String.format("Snowflake Storage Integration, field:%s cannot be null", key)));
}
```

this should check any missing field, and the following call should make the two methods simpler to read:

```
private String resolveAndValidateS3BucketPath(Connection connection, String integrationName) throws SQLException
{
    Map<String, String> properties = getStorageIntegrationProperties(connection, integrationName);
    if (properties.isEmpty()) {
        throw new IllegalArgumentException(
                String.format("Snowflake Storage Integration: name:%s not found", integrationName));
    }

    String bucketPath = requireProperty(properties, STORAGE_INTEGRATION_BUCKET_KEY);
    String provider = requireProperty(properties, STORAGE_INTEGRATION_STORAGE_PROVIDER_KEY);

    if (!"S3".equalsIgnoreCase(provider)) {
        throw new IllegalArgumentException(
                String.format("Snowflake Storage Integration, field:%s must be S3",
                        STORAGE_INTEGRATION_STORAGE_PROVIDER_KEY));
    }

    // Single path only
    if (bucketPath.split(",").length != 1) {
        throw new IllegalArgumentException(
                String.format("Snowflake Storage Integration, field:%s must be a single S3 path",
                        STORAGE_INTEGRATION_BUCKET_KEY));
    }

    // Validate it's an S3 path
    if (!bucketPath.startsWith("s3://")) {
        throw new IllegalArgumentException(
                String.format("Storage integration bucket path must be an S3 path: %s", bucketPath));
    }

    // Normalize trailing slash
    if (bucketPath.endsWith("/")) {
        bucketPath = bucketPath.substring(0, bucketPath.length() - 1);
    }

    return bucketPath;
}
```

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -333,132 +333,16 @@ private void handleDirectQueryPartitions(BlockWriter blockWriter, GetTableLayout
...
-                }
-            }
-            LOGGER.debug("Integration {} not found", integrationName);
-            return false;
-        }
-        catch (SQLException e) {
-            LOGGER.error("Error checking for integration {}: {}", integrationName, e.getMessage());
-            throw new SQLException("Failed to check for integration existence: " + e.getMessage(), e);
-        }
+        return constraints.getQueryPassthroughArguments().get(JdbcQueryPassthrough.QUERY);
```
AbdulR3hman, line 338: should we add a check to see if its QPT; and S3 integration enabled; we should throw an error? meaning users shouldn't be able to use S3 integration directly with QPT, given that we need to do good amount of work prior.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/SupportedTypes.java
chngpe: no, we don't have extractor support for this, it was coming as separate CR. but i agree, let me separate this out

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/data/S3BlockSpiller.java
chngpe: > why did we need maxRowsPerCall initially and now we don't?

The initial was to prevent `blocks that exceed the max block size`. We should not have this to begin with, from engine we don't care about the block size (s3blockSpillReader). I think this was use to protect the block size being too large hence create issue on s3 PutObject. However, with sdkv2 we can use high level Multi-part upload. This is less a concern now and Multi-part upload is coming next

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/utils/SnowflakeArrowTypeConverter.java
chngpe: I can try to deduplicate this, but the main idea is we are treating the decimal completely different as Snowflake tread number as decimal(38,x,128) hence doing the slight adjustment here with UTC timestamp

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeRecordHandler.java
```diff
@@ -418,7 +206,14 @@ protected ArrowReader constructArrowReader(String uri)
                 FileFormat.PARQUET,
                 uri);
         Dataset dataset = datasetFactory.finish();
-        ScanOptions options = new ScanOptions(/*batchSize*/ 32768);
+
+        // do a scan projection, only getting the column we want
+        ScanOptions options = new ScanOptions(/*batchSize*/ 32768,
```
AbdulR3hman: where is this `32768` comes from? and can we make it static?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeRecordHandler.java
```diff
@@ -167,60 +144,49 @@ public void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recordsR
...
-            //get column name and type from the Schema
-            HashMap<String, Types.MinorType> mapOfNamesAndTypes = new HashMap<>();
-            HashMap<String, Object> mapOfCols = new HashMap<>();
-
-            for (Field field : schemaName.getFields()) {
-                Types.MinorType minorTypeForArrowType = Types.getMinorTypeForArrowType(field.getType());
-                mapOfNamesAndTypes.put(field.getName(), minorTypeForArrowType);
-                mapOfCols.put(field.getName(), null);
-            }
+        if (s3ObjectKey.isEmpty()) {
```
AbdulR3hman, line 156: maybe a warning is warranted here; just in case Cx want to look at their logs. however it might be confusing; so maybe a debug?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/utils/SnowflakeArrowTypeConverter.java
AbdulR3hman: I meant to say that that one line that treats Snowflake's differently; we can just move it to the original file; it should still work no? I can talk to you offline to explain what I mean.

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
chngpe: updated

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeEnvironmentProperties.java
```diff
@@ -43,14 +43,6 @@ public class SnowflakeEnvironmentProperties extends JdbcEnvironmentProperties
...
-
-    private final boolean enableS3Export;
-
-    public SnowflakeEnvironmentProperties(Map<String, String> properties)
-    {
-        this.enableS3Export = Boolean.parseBoolean(properties.getOrDefault(ENABLE_S3_EXPORT, "false"));
-    }
 
     @Override
     public Map<String, String> connectionPropertiesToEnvironment(Map<String, String> connectionProperties)
```
aimethed, line 48: iirc this is how glue connection properties are read in, by not updating this for the S3 export for a customer to enable this, they have to do through lambda environment properties right?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
aimethed: Adding to this, line 215 of this file makes it seem like QPT works with S3 Integration?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -798,36 +678,113 @@ private boolean checkForView(TableName tableName) throws Exception
...
-        final String secretName = getDatabaseConnectionConfig().getSecret();
-        if (StringUtils.isNotBlank(secretName)) {
-            return new SnowflakeCredentialsProvider(secretName);
+        String s3ExportPath = Optional.ofNullable(storageIntegrationMap.get(STORAGE_INTEGRATION_BUCKET_KEY))
+                .orElseThrow(() -> new IllegalArgumentException(String.format("Snowflake Storage Integration, field:%s cannot be null", STORAGE_INTEGRATION_BUCKET_KEY)));
+
+        String provider = Optional.ofNullable(storageIntegrationMap.get(STORAGE_INTEGRATION_STORAGE_PROVIDER_KEY))
+                .orElseThrow(() -> new IllegalArgumentException(String.format("Snowflake Storage Integration, field:%s cannot be null", STORAGE_INTEGRATION_STORAGE_PROVIDER_KEY)));
+
+        if (!"S3".equalsIgnoreCase(provider)) {
```
aimethed, line 727: If the provider has to be S3, why ask for it?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
```diff
@@ -798,36 +678,113 @@ private boolean checkForView(TableName tableName) throws Exception
...
+
+    /**
+     * Get Snowflake storage integration name from config
+     * @return
+     */
+    private String getStorageIntegrationName()
+    {
+        // Check if integration name is provided in the config.
+        return this.getSFStorageIntegrationNameFromConfig().orElseThrow(() -> {
+            return new AthenaConnectorException("Snowflake storage integration name is missing from properties",
```
aimethed: nit: would it be better to explicitly say the property it is missing from? Instead of "from properties", like "integration name property <integration name property> was missing". Similar to how 722 and 725 say the property that cannot be null.

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeQueryStringBuilder.java
```diff
@@ -86,221 +77,97 @@ protected List<String> getPartitionWhereClauses(Split split)
...
-        name = name.replace(quoteCharacters, quoteCharacters + quoteCharacters);
-        return quoteCharacters + name + quoteCharacters;
-    }
-
-    protected String singleQuote(String name)
-    {
-        name = name.replace(singleQuoteCharacters, singleQuoteCharacters + singleQuoteCharacters);
-        return singleQuoteCharacters + name + singleQuoteCharacters;
+        sqlBaseString = expandSql(sqlBaseString, accumulator);
+        LOGGER.info("Expanded Generated SQL : {}", sqlBaseString);
```
aimethed, line 103: q: this log would have customer data right? When logs are in a customer account this is likely fine but is it different if this changes?

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeRecordHandler.java
aimethed: +1 The only way to get here is that s3Export is enabled, so the cx would be trying to use it.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/exceptions/AthenaConnectorException.java
```diff
@@ -81,6 +81,17 @@ public AthenaConnectorException(@Nonnull final Object response,
         requireNonNull(e);
     }
 
+    public AthenaConnectorException(@Nonnull final String message,
+                                    @Nonnull final Exception e,
+                                    @Nonnull final ErrorDetails errorDetails)
+    {
+        super(message, e);
+        this.errorDetails = requireNonNull(errorDetails);
+        this.response = null;
```
ejeffrli, line 90: why do we want a null response? How does this look like as an exception

ejeffrli (approved):

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeEnvironmentProperties.java
chngpe: >iirc this is how glue connection properties are read in, by not updating this for the S3 export for a customer to enable this, they have to do through lambda environment properties right?


yes, there is a separate task for that

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
chngpe: It is supported actually, i enable it

athena-snowflake/src/main/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandler.java
chngpe: they can also export it to Azure which we currently do not support, hence checking

Issue #3102, opened 2025-11-05
[BUG] statistics flooding postgres with errors with connector athena-postgresql


jmartinez-outrider:
**Describe the bug**
The federated lambda query keeps trying to get max and min of uuids by calling max(uuid) and min(uuid). These functions don't exist in postgres so you get thousands of errors

**To Reproduce**
1. setup a federated postgres connection to postgres rds instance
2. check the error logs in your rds instance

**Expected behavior**
Being able to deactivate statistics for postgres connector

**Screenshots / Exceptions / Errors**
Example error
```
2025-11-05 17:36:37 UTC:10.255.1.235(17628):role_name@db_name:[24943]:HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
2025-11-05 17:36:37 UTC:10.255.1.235(17628):role_name@db_name:[24943]:STATEMENT:  select min(id), max(id) from public.table_name
2025-11-05 17:36:37 UTC:10.255.1.235(55814):role_name@db_name:[24950]:ERROR:  function min(uuid) does not exist at character 8
```

**Connector Details (please complete the following information):**
 - Version: not in tags. Deployed on 2025-05-13T11:33:31.342+01:00
 - Name: postgres

jmartinez-outrider:
For people running in to this, I'm using this workaround right now (creating the missing functions)
```sql
-- functions to get rid of athena errors

CREATE OR REPLACE FUNCTION uuid_max(uuid, uuid)
RETURNS uuid AS $$
BEGIN
    RETURN GREATEST($1, $2);
END;
$$ LANGUAGE plpgsql IMMUTABLE STRICT;

-- Create a function to compare MIN uuid
CREATE OR REPLACE FUNCTION uuid_min(uuid, uuid)
RETURNS uuid AS $$
BEGIN
    RETURN LEAST($1, $2);
END;
$$ LANGUAGE plpgsql IMMUTABLE STRICT;

-- Create optimized binary MAX aggregate
CREATE OR REPLACE AGGREGATE max(uuid) (
    SFUNC = uuid_max,
    STYPE = uuid,
    COMBINEFUNC = uuid_max,
    PARALLEL = SAFE,
    SORTOP = operator (>) -- Essential for index optimization
);

-- Create optimized binary MIN aggregate
CREATE OR REPLACE AGGREGATE min(uuid) (
    SFUNC = uuid_min,
    STYPE = uuid,
    COMBINEFUNC = uuid_min,
[...]

PR #3166, opened 2025-12-16
Postgresql uuid type issue fix.

VenkatasivareddyTR:
*Issue #, if available:*
https://github.com/awslabs/aws-athena-query-federation/issues/3102
*Description of changes:*
Moved getSplitClauses() to Postgres since it is used only by the PostgreSQL connector. Also added a condition to skip split generation for UUID primary key columns, as PostgreSQL does not support MIN/MAX on UUIDs.

### DIFF
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
@@ -45,8 +45,6 @@ import com.amazonaws.athena.connectors.jdbc.connection.JdbcConnectionFactory;
 import com.amazonaws.athena.connectors.jdbc.qpt.JdbcQueryPassthrough;
 import com.amazonaws.athena.connectors.jdbc.resolver.DefaultJDBCCaseResolver;
 import com.amazonaws.athena.connectors.jdbc.resolver.JDBCCaseResolver;
-import com.amazonaws.athena.connectors.jdbc.splits.Splitter;
-import com.amazonaws.athena.connectors.jdbc.splits.SplitterFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
@@ -69,8 +67,6 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
-import java.sql.Statement;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
```diff
@@ -467,40 +460,6 @@ public abstract void getPartitions(
     @Override
     public abstract GetSplitsResponse doGetSplits(BlockAllocator blockAllocator, GetSplitsRequest getSplitsRequest);
 
-    protected List<String> getSplitClauses(final TableName tableName)
```
AbdulR3hman, line 480: todo; @AbdulR3hman double check the usage of this method.

athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
```diff
@@ -421,6 +430,91 @@ public static List<String> getCharColumns(Connection connection, String schema,
...
+     * Checks if a column is of UUID type in PostgreSQL.
+     *
+     * @param connection JDBC connection
+     * @param tableName Table containing the column
+     * @param columnName Column to check
+     * @return true if column is UUID type, false otherwise
+     */
+    private boolean isUuidColumn(Connection connection, TableName tableName, String columnName)
+    {
+        try (PreparedStatement stmt = connection.prepareStatement(GET_DATA_TYPE_QUERY)) {
```
AbdulR3hman, line 514: this approach will add extra latency for all RDS calls; instead; if we are sure RDS doesn't support this at all; we should override the `getSplitClauses` method in the Redshift metadata handler and not have to make a metadata call.

athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
```diff
@@ -421,6 +430,91 @@ public static List<String> getCharColumns(Connection connection, String schema,
...
 
+    /**
+     * This method automatically detects and skips UUID columns to prevent flooding PostgreSQL logs
+     * with "function min(uuid) does not exist" errors.
+     *
+     * @param tableName The table to generate split clauses for
+     * @return List of split clauses, empty if splits cannot be generated or if disabled
+     */
+
+    private List<String> getSplitClauses(final TableName tableName)
```
AbdulR3hman: see my other comment; I think we make this `protected`and override it for redshift; that way we don't have to make an extra metadata call.

AbdulR3hman (changes requested):

athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
VenkatasivareddyTR: Thank you @AbdulR3hman for the review. 
Overriden the `getSplitClauses` method in the Redshift metadata handler to skip the UUID type check, avoiding the extra metadata call.

athena-postgresql/src/main/java/com/amazonaws/athena/connectors/postgresql/PostGreSqlMetadataHandler.java
VenkatasivareddyTR: Thank you for the suggestion @AbdulR3hman. Made getSplitClauses `protected` and override it in the Redshift metadata handler.

burhan94 (commented):
## Code Review

### ✅ Looks Good

Well-implemented fix preventing PostgreSQL log flooding from UUID MIN/MAX errors. Good test coverage and proper error handling. No issues.

AbdulR3hman (dismissed):

AbdulR3hman (approved):

hareeshkumar-t (approved):

PR #3199, opened 2025-12-29
Add Query Federation Support with Substrait Integration and Credential Federation for JDBC Based Connectors

ZhitongYan:
# Add Query Federation Support with Substrait Integration and Credential Federation

## Overview
This PR adds support for query federation capabilities to AWS Athena Query Federation, enabling cross-catalog queries with proper credential management and Substrait query plan integration. The changes primarily focus on the JDBC connector framework and propagate necessary updates across all JDBC-based connectors.

### 1. Substrait Query Plan Support
- **New Classes:**
  - `SubstraitTypeAndValue`: Encapsulates SQL type and value information from Substrait plans
  - `SubstraitAccumulatorVisitor`: Visitor pattern implementation for extracting literals and parameters from Calcite SQL nodes

- **Enhanced `JdbcSplitQueryBuilder`:**
  - Added `prepareStatementWithCalciteSql()` method to translate Substrait plans to SQL
  - Improved handling of data types in prepared statements (DATE, TIMESTAMP, DECIMAL, etc.)
  - Support for timezone-aware date/timestamp conversions
  - Added `getSqlDialect()` method for database-specific SQL generation
  - Added `appendLimitOffsetWithValue()` for pagination support

### 2. Credential Federation Support
- **Updated `JdbcMetadataHandler` and `JdbcRecordHandler`:**
  - Added `AwsRequestOverrideConfiguration` parameter to credential provider methods
  - Implemented `getDatabaseConnectionSecret()` and `createCredentialsProvider()` methods
  - Support for FAS (Federated Access Service) token-based authentication
  - Proper credential propagation through query execution lifecycle

- **Modified `getSchema()` Signature:**
  - Added `AwsRequestOverrideConfiguration` parameter to enable credential federation
  - Updated across all JDBC-based connectors for consistency

### 3. Connector Updates
All JDBC-based connectors updated to support credential federation:
- Cloudera Hive
- Cloudera Impala
- DataLakeGen2
- DB2 and DB2 AS400
- Hortonworks Hive
- Oracle
- SAP HANA
- Snowflake
- SQL Server
- Synapse
- Teradata

## Breaking Changes
⚠️ **API Change:** The `getSchema()` method signature now includes an additional `AwsRequestOverrideConfiguration` parameter. Custom connector implementations will need to update their method signatures accordingly.

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
@@ -33,7 +33,6 @@ import org.apache.calcite.util.NlsString;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
@@ -70,14 +69,8 @@ public class SubstraitAccumulatorVisitor extends SqlShuttle
             LOGGER.info("literal value {} doesn't have an associated column. skipping", literal.toValue());
             return literal;
         }
-        Field arrowField = null;
-        try {
-            arrowField = schema.findField(currentColumn);
-        }
-        catch (final IllegalArgumentException e) {
-            LOGGER.warn("exact match for column {} not found.", currentColumn);
-            arrowField = findMatchingField(currentColumn);
-        }
+
+        Field arrowField = schema.findField(currentColumn);
 
         if (arrowField == null) {
             throw new IllegalArgumentException("field " + currentColumn + " not found in " + schema.getFields());
@@ -91,47 +84,6 @@ public class SubstraitAccumulatorVisitor extends SqlShuttle
             accumulator.add(new SubstraitTypeAndValue(typeName, literal.getValue(), currentColumn));
         }
         return new SqlDynamicParam(0, literal.getParserPosition());
-    }
-
-    /**
-     * Finds a matching field using case-insensitive exact match first, then case-sensitive prefix match (bidirectional).
-     * If multiple prefix matches are found, throws an exception to avoid ambiguity.
-     *
-     * @param columnName the column name to match
-     * @return the matching Field, or null if no match found
-     * @throws IllegalArgumentException if multiple prefix matches are found
-     */
-    private Field findMatchingField(final String columnName)
-    {
-        if (columnName == null || columnName.isEmpty()) {
-            return null;
-        }
-
-        // Strategy 1: Case-insensitive exact match
-        for (final Field field : schema.getFields()) {
-            if (field.getName().toLowerCase().equals(columnName.toLowerCase())) {
-                return field;
-            }
-        }
-
-        // Strategy 2: Case-sensitive prefix match
-        final List<String> prefixMatches = new ArrayList<>();
-        for (final Field field : schema.getFields()) {
-            if (field.getName().startsWith(columnName)) {
-                prefixMatches.add(field.getName());
-            }
-        }
-
-        if (prefixMatches.isEmpty()) {
-            return null;
-        }
-
-        if (prefixMatches.size() > 1) {
-            throw new IllegalArgumentException(String.format("Ambiguous column name '%s'. Multiple prefix matches found: %s",
-                columnName, prefixMatches));
-        }
-
-        return schema.findField(prefixMatches.get(0));
     }
 
     private SqlTypeName mapArrowTypeToSqlTypeName(final ArrowType arrowType)


athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
```diff
@@ -175,20 +176,30 @@ protected DatabaseConnectionConfig getDatabaseConnectionConfig()
...
+    {
+        DatabaseConnectionConfig databaseConnectionConfig = getDatabaseConnectionConfig();
+        if (Objects.nonNull(databaseConnectionConfig)) {
+            return databaseConnectionConfig.getSecret();
+        }
         return null;
     }
 
+    @Override
+    public CredentialsProvider createCredentialsProvider(String secretName, AwsRequestOverrideConfiguration requestOverrideConfiguration)
```
ejeffrli: Does this need to be overriden? I saw FederationRequestHandler.java already has this as the original behavior.

ZhitongYan: nice catch we can remove them in JDBC metadata and record handlers

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcSplitQueryBuilder.java
```diff
@@ -367,4 +382,170 @@ protected String appendLimitOffset(Split split, Constraints constraints)
     {
         return " LIMIT " + constraints.getLimit();
     }
+
+    protected String appendLimitOffsetWithValue(String limit, String offset)
+    {
+        if (offset == null) {
+            return " LIMIT " + limit;
+        }
+        return " LIMIT " + limit + " OFFSET " + offset;
```
ejeffrli, line 392: Is this syntax used by all jdbc sources? I know there can be slight differences sometimes. Does this work with oracle for example?

ZhitongYan: its a basic / common impl for JDBC connectors, and individual ones could override it to have its own logic.

ejeffrli: okay. is this the same as appendLimitOffset? I see that being used in OracleQueryStringBuilder now

ZhitongYan: appendLimitOffset and appendLimitOffsetWithValue are slightly different. 

- appendLimitOffset is like `LIMIT 10`
- appendLimitOffsetWithValue is like `LIMIT 10 OFFSET 0` with a paginated behavior.

ZhitongYan: For Oracle for example:
`
SELECT * FROM table
ORDER BY some_column
OFFSET 10 ROWS
FETCH NEXT 10 ROWS ONLY;
`

athena-snowflake/src/test/java/com/amazonaws/athena/connectors/snowflake/SnowflakeMetadataHandlerTest.java
```diff
@@ -106,13 +106,13 @@ public class SnowflakeMetadataHandlerTest
         extends TestBase
 {
     private static final Logger logger = LoggerFactory.getLogger(SnowflakeMetadataHandlerTest.class);
-    
+
```
ejeffrli, line 109: nit: what are all these blank changes? I see this new line in this file already

ejeffrli (approved):

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
```diff
@@ -391,7 +397,17 @@ protected Optional<ArrowType> convertDatasourceTypeToArrow(int columnIndex, int
...
+     * Gets the schema for a table with optional request override configuration.
+     *
+     * @param jdbcConnection the JDBC connection
+     * @param tableName the table name
+     * @param partitionSchema the partition schema
+     * @param requestOverrideConfiguration optional AWS request override configuration for credential federation
+     * @return the schema
+     * @throws Exception if an error occurs
+     */
+    protected Schema getSchema(Connection jdbcConnection, TableName tableName, Schema partitionSchema, AwsRequestOverrideConfiguration requestOverrideConfiguration)
```
AbdulR3hman, line 410: going through the changes; I have few comments; why are we adding this as a parameter to this method? which in turn forces us to override every single implementation? 

The `AwsRequestOverrideConfiguration` already available to class through inheritance. `getRequestOverrideConfig` ; there is no need to change the method signature to thread an already existing information that can be provided by the method. This should also cut down on so many changes

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
```diff
@@ -327,7 +333,7 @@ public GetTableResponse doGetQueryPassthroughSchema(final BlockAllocator blockAl
         jdbcQueryPassthrough.verify(getTableRequest.getQueryPassthroughArguments());
         String customerPassedQuery = getTableRequest.getQueryPassthroughArguments().get(JdbcQueryPassthrough.QUERY);
 
-        try (Connection connection = getJdbcConnectionFactory().getConnection(getCredentialProvider())) {
+        try (Connection connection = getJdbcConnectionFactory().getConnection(getCredentialProvider(getRequestOverrideConfig(getTableRequest)))) {
```
AbdulR3hman, line 336: See my comment below before this; but this is an example; where you didn't need to thread the the override config through this method.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcSplitQueryBuilder.java
```diff
@@ -26,15 +26,25 @@
 import com.amazonaws.athena.connector.lambda.domain.predicate.SortedRangeSet;
 import com.amazonaws.athena.connector.lambda.domain.predicate.ValueSet;
 import com.amazonaws.athena.connector.lambda.exceptions.AthenaConnectorException;
+import com.amazonaws.athena.connector.substrait.SubstraitSqlUtils;
```
AbdulR3hman, line 30: we going to have a problem with all these changes as they conflict with this PR: https://github.com/awslabs/aws-athena-query-federation/pull/3206/files

We are moving away from this to String Templates.

AbdulR3hman: we need to meet and discuss this with @burhan94

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.util.SqlShuttle;
+import org.apache.calcite.util.NlsString;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Map;
+
+public class SubstraitAccumulatorVisitor extends SqlShuttle
```
AbdulR3hman, line 39: qq; wouldn't this visitor be needed on non-jdbc related connectors? might worth moving it to the sdk package?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
```diff
@@ -0,0 +1,145 @@
...
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(SubstraitAccumulatorVisitor.class);
+
+    private List<SubstraitTypeAndValue> accumulator;
+    private Map<String, String> splitProperties;
+    private final Schema schema;
+    private String currentColumn;
```
AbdulR3hman, line 46: nit; maybe optional is a better choice?

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/visitor/SubstraitAccumulatorVisitor.java
```diff
@@ -0,0 +1,145 @@
...
+            // such as LIMIT
+            LOGGER.info("literal value {} doesn't have an associated column. skipping", literal.toValue());
+            return literal;
+        }
+        Field arrowField = null;
+        try {
+            arrowField = schema.findField(currentColumn);
+        }
+        catch (IllegalArgumentException e) {
+            LOGGER.warn("exact match for column {} not found. searching with prefixes", currentColumn);
```
AbdulR3hman: why are we using prefixes? shouldn't the match be 1:1?

AbdulR3hman: I guess another way to ask this; in which case a prefix is acceptable?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
AbdulR3hman: nvm, I think this is specific to JDBC.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
ZhitongYan: Thanks for the feedback Abdul. As you also noticed the method signature does not need to be updated at all as `getRequestOverrideConfig` is available, but there is one special case which is `getSchema` method.

The reason for updating method signature for `getSchema` is due to: input for `getSchema` [`(Connection jdbcConnection, TableName tableName, Schema partitionSchema)`] does not have information for input for `getRequestOverrideConfig` [either `(Map<String, String> configOptions)` or `(FederationRequest request)`]

So I tried several workarounds and chose the current approach. 
1. update getSchema to have configOptions information (not preferred as more updates needed due to all JDBC based connectors are based on the method signature)
2. update getSchema to have AwsRequestOverrideConfiguration information (current change)
3. method overload: having below methods at the same time. One con for this approach is sub classes need to be very clear to know which method it wants to override and which method it want to call. This con adds risk for clean and easy backwards compatible changes. To avoid such ambiguity we need to mark 3 parameter method final to avoid allowing others to override, which again forces the method signature update in sub classes. 
3.1. `protected Schema getSchema(Connection jdbcConnection, TableName tableName, Schema partitionSchema)`
3.2. `protected Schema getSchema(Connection jdbcConnection, TableName tableName, Schema partitionSchema, AwsRequestOverrideConfiguration requestOverrideConfiguration)`

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcSplitQueryBuilder.java
ZhitongYan: briefly checked the PR #3206 and I think we don't have any overlaps from both logic and code change stand point of view. In this CR it goes to a Substrait based implementation which leverages AST internally in query federation case. PR #3206 updates current non-query federation case from String builder to String templates.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
ZhitongYan: yeah I thought it might be more easier to follow if we put it in the JDBC context.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/visitor/SubstraitAccumulatorVisitor.java
ZhitongYan: Calcite is used as an intermediate representation layer to convert between Substrait plans (query execution plans) and SQL dialects. Column index renaming occurs at `RelToSqlConverter` implementation. It helps to remove the ambiguity due to possible duplicate names in the select lists / aggregate function result / projections and joins etc. 

Prefix matching here is a fallback behavior when there is no exact matches between substrait table schema and SqlNode column identifiers. This is a imperfect pragmatic workaround for column discrepancy issues, but we have tested with significant amount of queries using TPCH benchmarks on the correctness.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
ZhitongYan: String vs Null is quite straightforward so I kept it as is, maybe worth to update the optional for more complex structures

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java
AbdulR3hman: got it; I looked at it again; it seems this method is mainly called from the JDBC base metadata handler; the intention behind my comment is to try to not change the method signature which requires a change in all places that are begin overridden; 12 places as far as I can tell. Alternatively was thinking to make it a protected class member; but that's Nit at this point; I'll concede haha.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/visitor/SubstraitAccumulatorVisitor.java
AbdulR3hman: got it; I think I have a small fellow up on this; I would recommend two things:

1. Prefix match only if it resolves uniquely | i.e. if we find more than one match; we shouldn't match. 
2. Should we be worried about the prefix matching trivial-alias? as in `EXPR$N` ? I assume not give nwe are using the field from the schema directly. but thought of asking.

ZhitongYan: Thanks for the advice, I just enhanced the matching logic based on your suggestion. Also for cases like `EXPR$N` we have coverage on unit tests here: https://github.com/ZhitongYan/aws-athena-query-federation/blob/pr-jdbc/athena-federation-sdk/src/test/java/com/amazonaws/athena/connector/substrait/SubstraitSqlUtilsTest.java#L401. 

Since we're using fields from the schema directly, those should be the actual data source field names, not Calcite-generated trivial aliases.

AbdulR3hman: thank you

ZhitongYan:
ideally, 1:1 match should be followed between substrait and SQL nodes, even with case insensitive exact match. Team found it out due to they were trying to invoke the lambda directly with manually generated substrait plan strings. Even technically valid but not a case for end to end workflow. Removing it to completely eliminate ambiguity on column matching

ejeffrli (approved):

PR #3202, opened 2026-01-06
Enhance exception handling

suyinlee:
*Issue #, if available:* In the managed connector workflow, all exceptions are now classified as client exceptions on the Glue side due to the difficulty of catching connector-thrown exceptions at multiple points. We need to add SDK functionality that enables connectors to handle their own connector-specific exceptions.

*Description of changes:* Added extensible exception handling framework to improve error management across Athena connectors. Each connector can use this new handler to handle client exceptions.


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandler.java
```diff
@@ -154,10 +165,22 @@ public final void handleRequest(BlockAllocator allocator, FederationRequest rawR
...
+                        metadataHandler.doHandleRequest(allocator, objectMapper, (MetadataRequest) rawReq, outputStream));
+            }
+            else {
+                metadataHandler.doHandleRequest(allocator, objectMapper, (MetadataRequest) rawReq, outputStream);
+            }
         }
         else if (rawReq instanceof RecordRequest) {
-            recordHandler.doHandleRequest(allocator, objectMapper, (RecordRequest) rawReq, outputStream);
+            if (exceptionHandler != null) {
+                exceptionHandler.handle(() ->
```
ZhitongYan: this might lost some of the thread local information. instead of having a brand new class ExceptionHandler, maybe we can just add a abstract method in FedertaionReqeustHandler interface and make other handlers implements their own.

suyinlee: Makes sense, will refactor the code.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandler.java
```diff
@@ -176,4 +182,13 @@ private void assertNotNull(FederationResponse response)
             throw new AthenaConnectorException("Response was null", ErrorDetails.builder().errorCode(FederationSourceErrorCode.INVALID_RESPONSE_EXCEPTION.toString()).build());
         }
     }
+
+    public Exception handleException(Exception e)
+    {
+        if (e instanceof KmsException) {
+            return new AthenaConnectorException(e.getMessage(),
```
ZhitongYan, line 189: 1. where in original code handles KmsException? 
2. can we sanitize a little bit e.getMessage() to avoid internal impl leaks?

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBRecordMetadata.java
```diff
@@ -222,4 +226,24 @@ public boolean getGlueTableContainedPreviouslyUnsupportedTypes()
...
     }
+
+    /**
+     * Transforms DynamoDB exceptions into AthenaConnectorException with appropriate error code.
+     * Returns other exceptions unchanged.
+     *
+     * @param e The exception to handle
+     * @return The transformed exception or original exception
+     */
+    public static Exception handleDynamoDBException(Exception e)
```
ZhitongYan: why not directly add this into ddb composite handler

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandler.java
suyinlee: The message comes from KmsKeyFactory. Previously, we only converted KMS key not found errors to AthenaConnectorException in KmsKeyFactory, but we’ve seen additional cases such as AccessDenied. These messages are already customer-visible and don’t expose internal implementation details; this change simply wraps them as AthenaConnectorException for consistency.

athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/util/DDBRecordMetadata.java
suyinlee: sg. I can move this to ddb composite handler.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/CompositeHandler.java
```diff
@@ -176,4 +182,13 @@ private void assertNotNull(FederationResponse response)
...
+                    ErrorDetails.builder().errorCode(FederationSourceErrorCode.INVALID_INPUT_EXCEPTION.toString()).build());
+        }
```
DilipReddyGaddam, line 191: why are we using  FederationSourceErrorCode.INVALID_INPUT_EXCEPTION ?

suyinlee: Synced with Dilip offline. We are setting the error code to INVALID_INPUT_EXCEPTION for all KMS exceptions as a general default, unless we want to triage each type of KMS exception and map it to a specific Glue error code. We can discuss this with @hamburha and decide if we want to make it more specific.

AbdulR3hman: what was the outcome of this?

suyinlee: This is wrapping KMS exception message in the AthenaConnectorException with error code INVALID_INPUT_EXCEPTION.

Issue #14, opened 2019-11-15
[FEATURE] Refactor JDBC & BigQuery Connectors to use StringTemplate


avirtuos:
Refactor JDBC & BigQuery Connectors to use StringTemplate since that will allow us to make dialect specific templates for the basic logic of translating constraints to SQL fragments we can push down into other SQL engines we federate to.

macohen:
#2870 closes this issue for BigQuery

PR #3206, opened 2026-01-09
Feature to support string templates in JDBC Connectors.

ritiktrianz:
*Issue #, if available:*
#14 

*Description of changes:*
This PR introduces the base implementation for string templates in JDBC connectors. The changes establish a standardized approach to SQL query generation using StringTemplate, a departure from the prior use of StringBuilder and string concatenation.


AbdulR3hman:
This is no longer needed; discussed internally the reasoning behind it.

PR #3214, opened 2026-01-15
Adding validation on DocumentDB ListTables response to handle nextpage token

DilipReddyGaddam:
*Issue #, if available:*

*Description of changes:*
Adding validation on DocumentDB ListTables response to handle nextpage token. Nextpage token will be null if there no more tables to fetch.

### DIFF
--- a/athena-db2/src/main/java/com/amazonaws/athena/connectors/db2/Db2Constants.java
+++ b/athena-db2/src/main/java/com/amazonaws/athena/connectors/db2/Db2Constants.java
@@ -39,7 +39,7 @@ public class Db2Constants
             "from syscat.tables where type in ('T', 'U', 'V', 'W') and tabschema = ? " +
             "ORDER BY tabname OFFSET ? ROWS FETCH NEXT ? ROWS ONLY";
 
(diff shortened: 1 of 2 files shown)


athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandler.java
```diff
@@ -230,6 +230,9 @@ public ListTablesResponse doListTables(BlockAllocator blockAllocator, ListTables
 
         List<TableName> paginatedTables = tableNames.map(tableName -> new TableName(request.getSchemaName(), tableName)).collect(Collectors.toList());
         combinedTables.addAll(paginatedTables);
+        if (combinedTables.isEmpty()) {
```
burhan94: why not check if the combinedTables.size < pageSize?

if page size for first call is 10 and there are 11 tables, we would make a 2nd call for 1 table, and a 3rd call since combinedTables.size == 1 in the 2nd call.

burhan94 (commented):
.

PR #3215, opened 2026-01-15
Postgresql &  MySQL Substrait Implementation

stelanof88:
*Description of changes:*
MySql & PostgreSql substrait implementation

### DIFF
--- a/athena-datalakegen2/src/test/java/com/amazonaws/athena/connectors/datalakegen2/DataLakeGen2EnvironmentPropertiesTest.java
+++ b/athena-datalakegen2/src/test/java/com/amazonaws/athena/connectors/datalakegen2/DataLakeGen2EnvironmentPropertiesTest.java
@@ -1,0 +1,123 @@
+/*-
+ * #%L
+ * athena-datalakegen2
+ * %%
+ * Copyright (C) 2019 - 2025 Amazon Web Services
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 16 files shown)


athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcSplitQueryBuilder.java
```diff
@@ -422,7 +423,15 @@ protected PreparedStatement prepareStatementWithCalciteSql(
 
             PreparedStatement statement = jdbcConnection.prepareStatement(root.toSqlString(sqlDialect).getSql());
 
-            handleDataTypesForPreparedStatement(statement, accumulator, tableSchema);
+            ParameterMetaData metaData = statement.getParameterMetaData();
+            if (metaData != null && metaData.getParameterCount() != accumulator.size()) {
```
burhan94, line 439: why is this check needed and does this skip predicate pushdown if this code path is hit?

stelanof88: The check is necessary because MySQL's Calcite dialect embeds filter values directly in SQL. ie `WHERE age > 25` instead of using placeholders `WHERE age > ?`, which causes a mismatch between the statement's parameter count (0) and the accumulator's collected parameters (1). Predicate pushdown still works because the filter is already embedded in the SQL string sent to the database. The check only skips the unnecessary parameter binding step that would otherwise fail with the following error `Parameter index out of range (1 > number of parameters, which is 0)`

AbdulR3hman:
heads up; I've update the PR title.

burhan94:
Is there any way we can create some unit tests for this that validate that the SQL is being translated correctly?

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcSplitQueryBuilder.java
```diff
@@ -422,7 +423,15 @@ protected PreparedStatement prepareStatementWithCalciteSql(
...
+                LOGGER.error("Parameter count mismatch: SQL has {} parameters, accumulator has {}. Skipping parameter binding.",
```
ejeffrli: Is this an actual error? Maybe this should be LOGGER.warn(...) if it is just skipping for some scenarios

PR #3217, opened 2026-01-16
Refactor CrossAccountCredentialsProviderV2 to use StsAssumeRoleCreden…

joshualiao:
*Issue #, if available:* N/A

*Description of changes:* `CrossAccountCredentialsProviderV2` uses `StaticCredentialsProvider` which does not refresh credentials. When Lambda containers using cross account credentials stay warm longer than the STS session duration (default 1 hour), cross-account queries will fail. PR refactors to `StsAssumeRoleCredentialsProvider`. 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. [Yes]

### DIFF
--- a/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/credentials/CrossAccountCredentialsProviderV2.java
+++ b/athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/credentials/CrossAccountCredentialsProviderV2.java
@@ -7,9 +7,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


athena-dynamodb/src/main/java/com/amazonaws/athena/connectors/dynamodb/credentials/CrossAccountCredentialsProviderV2.java
```diff
@@ -48,10 +45,10 @@ public static AwsCredentialsProvider getCrossAccountCredentialsIfPresent(Map<Str
...
                     .build();
-            AssumeRoleResponse assumeRoleResponse = stsClient.assumeRole(assumeRoleRequest);
-            Credentials credentials = assumeRoleResponse.credentials();
-            AwsSessionCredentials sessionCredentials = AwsSessionCredentials.create(credentials.accessKeyId(), credentials.secretAccessKey(), credentials.sessionToken());
-            return StaticCredentialsProvider.create(sessionCredentials);
+            return StsAssumeRoleCredentialsProvider.builder()
+                    .stsClient(stsClient)
+                    .refreshRequest(assumeRoleRequest)
+                    .build();
         }
```
burhan94, line 62: I recommend adding a check here to resolve credentials using resolveCredentials() API. 

eg:


```
try {
    provider.resolveCredentials();
} catch (Exception e) {
    // handle or rethrow so failures are detected early
    throw new AthenaConnectorException("Failed to assume role for " + assumeRoleRequest.roleArn() + e, ErrorDetails.builder().errorCode(FederationSourceErrorCode.INVALID_CREDENTIALS_EXCEPTION.toString()).build());
}
```

joshualiao: Thanks for the catch, this will keep the original fail-fast behavior. Added exception logic

burhan94 (approved):

PR #3229, opened 2026-01-20
DocDB Substrait Implementation

AbdulR3hman:
Enabling DocDB Substrait implementation. 

This is a refactored and rebased (for easier review) version of [PR-3016](https://github.com/awslabs/aws-athena-query-federation/pull/3016)


athena-docdb/src/test/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandlerTest.java
```diff
@@ -635,14 +636,7 @@ public void doListTables_withPagination_returnsPaginatedResults() throws Excepti
         assertEquals("Should return 1 table for final page", 1, finalResponse.getTables().size());
         List<TableName> finalPageTables = new ArrayList<>(finalResponse.getTables());
         assertEquals(new TableName(DEFAULT_SCHEMA, TABLE5), finalPageTables.get(0));
-        assertEquals("6", finalResponse.getNextToken());
-
-        // Verify empty page after all data is consumed
-        ListTablesRequest emptyRequest = new ListTablesRequest(IDENTITY, QUERY_ID, DEFAULT_CATALOG, DEFAULT_SCHEMA,
```
AbdulR3hman, line 641: this part of the test expects that the paging happens infinitely; and that the server pagination is stateful; which it isn't. by setting the next token to null; we are expecting to restart from the start which what happens here.

AbdulR3hman: therefore this part of the test isn't correct hence removed.

athena-docdb/src/main/java/com/amazonaws/athena/connectors/docdb/DocDBMetadataHandler.java
```diff
@@ -365,4 +429,96 @@ protected Field convertField(String name, String glueType)
...
+     *   "port": 27017
+     * }
+     * </pre>
+     * 
+     * @param configOptions Map containing federated identity configuration including:
+     *                     HOST, PORT, secret_arn, JDBC_PARAMS, ENFORCE_SSL, AUTHENTICATION_DATABASE
+     * @return Fully constructed MongoDB connection string in format: mongodb://username:password@host:port/?jdbcParams
+     * @throws RuntimeException if JSON credential parsing fails or required parameters are missing
+     */
+    private String getConfigOptionsFromFederatedIdentity(Map<String, String> configOptions)
```
ejeffrli, line 461: Is there a reason to use this function instead of DocDbEnvironmentProperties? That should already contain the connection string, secrets, etc.

DilipReddyGaddam: In Managed connectors query federation use case DocDbEnvironmentProperties will not have required connection properties. They will be present in the createconnection API properties(which will be populated in configOptions).

PR #3233, opened 2026-01-23
Uppercase ReadRecordsRequest tableName and schema fields for UPPERCASE_ONLY catalog casing

skyyiiii:
This PR adds support for uppercasing ReadRecordsRequest tableName and schema fields when the connection CATALOG_CASING_FILTER is set to UPPERCASE_ONLY.

**Changes:**
- Modified ReadRecordsRequest to uppercase tableName and schema fields based on catalog casing filter
- Ensures consistency with catalog casing requirements for uppercase-only configurations

**Testing:**
- Verified functionality with UPPERCASE_ONLY catalog casing filter
- Existing tests continue to pass

### DIFF
--- a/.github/workflows/run_release_tests.yml
+++ b/.github/workflows/run_release_tests.yml
@@ -41,7 +41,7 @@ jobs:
           echo PATH=$PATH >> $GITHUB_ENV
 
       - name: Setup AWS Cred
-        uses: aws-actions/configure-aws-credentials@v4
+        uses: aws-actions/configure-aws-credentials@v5
         with:
           aws-region: us-west-2
           role-duration-seconds: 18000
--- a/athena-cloudwatch/pom.xml
+++ b/athena-cloudwatch/pom.xml
@@ -32,7 +32,7 @@
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>cloudwatchlogs</artifactId>
-            <version>2.40.13</version>
+            <version>2.41.17</version>
             <exclusions>
                 <!-- replaced with jcl-over-slf4j -->
                 <exclusion>
--- a/athena-dynamodb/Dockerfile
+++ b/athena-dynamodb/Dockerfile
@@ -3,6 +3,12 @@ ARG JAVA_VERSION=11
 # Use the specified version of Java
 FROM public.ecr.aws/lambda/java:${JAVA_VERSION}
 
+# Fix CVE-2025-14087: glib2 buffer underflow
+# Fix CVE-2025-64720 and CVE-2025-64505: libpng vulnerabilities
+RUN yum update -y glib2 libpng && \
+    yum clean all && \
+    rm -rf /var/cache/yum
+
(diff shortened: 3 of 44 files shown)


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/RecordHandler.java
```diff
@@ -276,4 +287,37 @@ private void assertNotNull(FederationResponse response)
...
+        // Check if uppercase casing filter is enabled
+        String casingFilter = request.getIdentity().getConfigOptions().getOrDefault(EnvironmentConstants.CATALOG_CASING_FILTER, null);
+        if (!EnvironmentConstants.UPPERCASE_ONLY.equals(casingFilter)) {
+            return request; // No transformation needed
+        }
+        
+        // Convert table and schema names to uppercase
+        TableName originalTableName = request.getTableName();
+        TableName uppercaseTableName = new TableName(
+                originalTableName.getSchemaName() != null ? originalTableName.getSchemaName().toUpperCase() : null,
```
ejeffrli, line 302: can schema/table ever be null here?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/RecordHandler.java
```diff
@@ -276,4 +287,37 @@ private void assertNotNull(FederationResponse response)
...
+                originalTableName.getTableName() != null ? originalTableName.getTableName().toUpperCase() : null);
+        
+        // Convert schema field names to uppercase, except if field name is in partition columns
```
ejeffrli, line 305: why are partition columns excluded from uppercasing?

ejeffrli:
We already have casing_mode to handle cases. https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/resolver/CaseResolver.java

Is it possible we continue to use that environment property and then add another casing variable or extend casing_mode to support uppercasing schema? I think having two properties is better than extending

- Scenario: UPPERCASE_SCHEMA.UPPERCASE_TABLE with lowercase columns requires two casing properties

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/RecordHandler.java
DilipReddyGaddam: This Table name is built in DNA service. Ideally it shouldn't be null but I don't think there is any harm in checking it.

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/handlers/RecordHandler.java
DilipReddyGaddam: partition columns are federation sdk generated columns they are always lower case. so we don't need to uppercase them

PR #3234, opened 2026-01-23
Redshift - Substrait/Managed Connector Implementation

AbdulR3hman:
depends on https://github.com/awslabs/aws-athena-query-federation/pull/3215

This change will allow Redshift to work with substrait; dependent on PR-3215

----


burhan94:
Just to confirm, these changes are backwards compatible right? We only send substrait plans via the DNA connector right now. I didn't check the base classes to validate. 

Pls don't merge until this is validated

ejeffrli:
Can we have more descriptive title/description that this is for redshift?

AbdulR3hman:
@ejeffrli @burhan94 both comments address; if you can review/approve thanks.

PR #3259, opened 2026-02-09
Add BigQuery Substrait Support

suyinlee:
*Issue #, if available:*

*Description of changes:*
1. Add Query Federation Substrait Support for Google BigQuery
2. Removes the timeStampColsList implementation. Treats all TIMESTAMP columns uniformly as DATETIME(MILLISECOND)


athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
```diff
@@ -221,9 +221,8 @@ public GetTableResponse doGetQueryPassthroughSchema(BlockAllocator allocator, Ge
         }
         com.google.cloud.bigquery.Schema bigQuerySchema = statistics.getSchema();
         SchemaBuilder schemaBuilder = SchemaBuilder.newBuilder();
-        List<String> timeStampColsList = new ArrayList<>();
 
-        convertBigQuerySchema(bigQuerySchema, timeStampColsList, schemaBuilder);
+        convertBigQuerySchema(bigQuerySchema, schemaBuilder);
```
ejeffrli, line 225: is this change backwards compatible? Will customer see their dates/timestamps differently?

suyinlee: Hi Jeff, this change is not backwards compatible. As discussed with Burhan last week, we've decided to remove the timeStampColsList implementation that stores timestamp information in schema metadata. Instead, we're changing the timestamp type from String to DateTime(MILLISECOND).

This change will result in losing 3 digits of precision (microseconds), but since this is how we handle timestamps in all other connectors, we've aligned on moving forward with this approach for consistency across the federation SDK.

PR #3280, opened 2026-02-13
remove arrow translation during Substrait to Calcite conversion

ZhitongYan:
Remove Arrow Translation as a intermediate step during Substrait to Calcite conversion. This change just simplifies the logic without adding / removing / updating any functionality. 

### DIFF
--- a/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
+++ b/athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
@@ -28,6 +28,7 @@ import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlLiteral;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.SqlSelect;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.util.SqlShuttle;
 import org.apache.calcite.util.NlsString;
@@ -86,6 +87,10 @@ public class SubstraitAccumulatorVisitor extends SqlShuttle
     {
         SqlKind kind = call.getOperator().getKind();
         
+        if (kind == SqlKind.SELECT) {
+            return handleSelect((SqlSelect) call);
+        }
+        
         // Handle AND/OR operators - may contain standalone boolean columns
         if (kind == SqlKind.AND || kind == SqlKind.OR) {
             return handleLogicalOperator(call);
@@ -121,12 +126,22 @@ public class SubstraitAccumulatorVisitor extends SqlShuttle
             return handleIsNull(call);
         }
         
-        return super.visit(call);
+        SqlNode result = super.visit(call);
+        return result;
     }
 
     @Override
     public SqlNode visit(SqlIdentifier id)
     {
+        if (inWhereClause && id.isSimple() && isBooleanColumn(id.getSimple())) {
+            // Transform bool_col to bool_col = TRUE
+            SqlLiteral trueLiteral = SqlLiteral.createBoolean(true, id.getParserPosition());
+            addToAccumulator(id.getSimple(), trueLiteral);
+            SqlDynamicParam param =
+                    new SqlDynamicParam(accumulator.size() - 1, trueLiteral.getParserPosition());
+            return org.apache.calcite.sql.fun.SqlStdOperatorTable.EQUALS
+                    .createCall(id.getParserPosition(), id, param);
+        }
         if (id.isSimple()) {
             columnStack.push(id.getSimple());
         }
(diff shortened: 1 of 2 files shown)


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
```diff
@@ -19,115 +19,314 @@
...
-    private final Map<String, String> splitProperties;
-    private final Schema schema;
-    private String currentColumn;
 
-    public SubstraitAccumulatorVisitor(final List<SubstraitTypeAndValue> accumulator, final Map<String, String> splitProperties, final Schema schema)
+    /** Row type schema used to resolve column types when parameterizing literals. */
+    private final RelDataType schema;
+
+    /** Tracks the current column context so literals can be associated with their column. */
+    private final Deque<String> columnStack = new ArrayDeque<>();
```
burhan94, line 74: so this is needed in order to track the index of the columns defined in the plan?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
```diff
@@ -19,115 +19,314 @@
...
     @Override
-    public SqlNode visit(final SqlIdentifier id)
+    public SqlNode visit(SqlCall call)
+    {
+        // Handle SqlSelect specially to track WHERE clause context
+        if (call instanceof SqlSelect) {
+            return handleSelect((SqlSelect) call);
+        }
+
+        SqlKind kind = call.getOperator().getKind();
```
burhan94, line 88: can we turn this into a case statement and list out the different SqlKind's that can exist here?

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcSplitQueryBuilder.java
```diff
@@ -489,23 +492,17 @@ private PreparedStatement handleDataTypesForPreparedStatement(PreparedStatement
...
                     if (typeAndValue.getValue() instanceof Number) {
+                        // Assume days since epoch for numeric date values
                         long numericValue = ((Number) typeAndValue.getValue()).longValue();
-                        if (dateType.getUnit() == DateUnit.DAY) {
-                            long utcMillis = numericValue * 24L * 60L * 60L * 1000L; // days → ms
-                            int offsetVal = TimeZone.getDefault().getOffset(utcMillis);
-                            utcMillis -= offsetVal;
-                            statement.setDate(i + 1, new Date(utcMillis));
-                        }
-                        else if (dateType.getUnit() == DateUnit.MILLISECOND) {
```
burhan94, line 503: hmm, why are we getting rid of the DateUnit.Millisecond?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
ZhitongYan: There are a lot more SqlKinds but those SqlKind values don't interfere with literal parameterization - they simply allow it to happen automatically through the visitor pattern, they automatically parameterize any literals found in child nodes.

Do you want to list all of those kinds explicitly?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
ZhitongYan: not necessarily index but keep track of current unmatched columns (identifiers) which have a literal associated.

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcSplitQueryBuilder.java
ZhitongYan: https://calcite.apache.org/docs/reference.html

Calcite DATE only have days since epoch so I removed it

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitAccumulatorVisitor.java
```diff
@@ -19,115 +19,314 @@
...
+
+    private SqlNode handleLike(SqlCall call)
+    {
+        if (!(call.operand(0) instanceof SqlIdentifier) || !(call.operand(1) instanceof SqlLiteral)) {
+            return super.visit(call);
         }
-        else if (arrowType instanceof ArrowType.Timestamp) {
-            return SqlTypeName.TIMESTAMP;
+        
+        SqlIdentifier identifier = (SqlIdentifier) call.operand(0);
```
burhan94, line 290: you have a statement above that says !(call.operand(1) instanceof SqlLiteral), but here you're casting call.operand(0). what happens if call.operand(0) is not a SqlItentifier?

burhan94: also, there's no guarantee that LIKE pattern will always be Identifier + Literal. 

for example:

```
select * from call_center where cc_call_center_id like concat(cc_call_center_id, 'a') limit 10;
```

ZhitongYan: nice catch, yeah it did not consider nested function calls. I have also updated for other SqlCall to handle this in a recursively / transparent manner. Thanks!

PR #3287, opened 2026-02-17
Added unit tests in athena-dynamodb connector

Jithendar12:
*Issue #, if available:*

*Description of changes:*
Added unit tests in athena-dynamodb connector

### DIFF
--- a/athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DDBTypeUtilsTest.java
+++ b/athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DDBTypeUtilsTest.java
@@ -48,7 +48,7 @@ import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.core.SdkBytes;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
 
-import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.time.LocalDate;
@@ -102,9 +102,7 @@ public class DDBTypeUtilsTest
     private DDBRecordMetadata ddbRecordMetadata;
 
     @Before
-    public void setUp()
-            throws IOException
-    {
+    public void setUp() {
         ddbRecordMetadata = mock(DDBRecordMetadata.class);
     }
 
@@ -202,9 +200,7 @@ public class DDBTypeUtilsTest
     }
     
     @Test
-    public void inferArrowField_withListContainingNull_infersListFieldWithVarcharChild()
-            throws Exception
-    {
+    public void inferArrowField_withListContainingNull_infersListFieldWithVarcharChild() {
         List<String> inputArray = new ArrayList<>();
         inputArray.add("value1");
         inputArray.add(null);
@@ -363,7 +359,7 @@ public class DDBTypeUtilsTest
         }
         catch (AthenaConnectorException ex) {
             assertTrue("Exception message should contain error about unknown type",
-                    ex.getMessage() != null && !ex.getMessage().isEmpty());
+                    ex.getMessage() != null && !ex.getMessage().isEmpty() && ex.getMessage().contains("Unknown type["));
         }
     }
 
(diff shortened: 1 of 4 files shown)


athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DDBTypeUtilsTest.java
```diff
@@ -185,6 +217,537 @@ public void inferArrowFieldListWithNullTest() throws Exception
...
+    @Test
+    public void getArrowFieldFromDDBType_withUnknownType_throwsAthenaConnectorException()
+    {
+        try {
+            DDBTypeUtils.getArrowFieldFromDDBType("testUnknown", UNKNOWN_TYPE);
+            fail("Expected AthenaConnectorException was not thrown");
+        }
+        catch (AthenaConnectorException ex) {
+            assertTrue("Exception message should contain error about unknown type",
+                    ex.getMessage() != null && !ex.getMessage().isEmpty());
```
fal-bharadwaj: We're saying that "Exception message should contain error about unknown type" but we're not checking for that, we're only checking that exception isn't null. Let's do an exact exception or substring match here along with the null checks.

athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DDBTypeUtilsTest.java
```diff
@@ -185,6 +217,537 @@ public void inferArrowFieldListWithNullTest() throws Exception
...
+            Field childField = new Field("child", FieldType.nullable(Types.MinorType.VARCHAR.getType()), null);
+            Field field = new Field(TEST_LIST, FieldType.nullable(Types.MinorType.LIST.getType()), Collections.singletonList(childField));
+            DDBRecordMetadata metadata = new DDBRecordMetadata(new Schema(Collections.singletonList(field)));
+            
+            DDBTypeUtils.coerceListToExpectedType(mapValue, field, metadata);
+            fail("Expected AthenaConnectorException was not thrown");
+        }
+        catch (AthenaConnectorException ex) {
+            assertTrue("Exception message should contain error about map instead of list",
+                    ex.getMessage() != null && !ex.getMessage().isEmpty());
```
fal-bharadwaj: We're saying that "Exception message should contain error about map instead of list" but we're not checking for that, we're only checking that exception isn't null. Let's do an exact exception or substring match here along with the null checks.

athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DDBTypeUtilsTest.java
```diff
@@ -185,6 +217,537 @@ public void inferArrowFieldListWithNullTest() throws Exception
...
+    @Test
+    public void toAttributeValue_withUnsupportedType_throwsAthenaConnectorException()
+    {
+        try {
+            Object unsupportedValue = new StringBuilder(UNSUPPORTED);
+            DDBTypeUtils.toAttributeValue(unsupportedValue);
+            fail("Expected AthenaConnectorException was not thrown");
+        }
+        catch (AthenaConnectorException ex) {
+            assertTrue("Exception message should contain error about unsupported type",
```
fal-bharadwaj: Please do an assertion on actual exception message along with the null checks, please also change it everywhere else in this PR where we've followed the same pattern.

athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DDBTypeUtilsTest.java
ritiktrianz: Hi @fal-bharadwaj , thanks for the review.
I’ve updated the assertions to perform exact exception or substring matching, along with null and empty checks.

athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DDBTypeUtilsTest.java
ritiktrianz: Updated the assertions to perform exact exception or substring matching, along with null and empty checks.

athena-dynamodb/src/test/java/com/amazonaws/athena/connectors/dynamodb/DDBTypeUtilsTest.java
ritiktrianz: Hi @fal-bharadwaj  ,Updated the assertions to perform exact exception or substring matching, along with null and empty checks.

PR #3294, opened 2026-02-23
Teradata - Substrait/Managed Connector Implementation

DilipReddyGaddam:
*Issue #, if available:*

*Description of changes:*
This PR enables Teradata connector to support query federation using Substrait.


ZhitongYan (dismissed):
Discussed offline that Teradata Fetch will be supported at Trino Engine level and preserve the connector capabilities. Hence we remove fetch quietly in connector.

athena-teradata/src/main/java/com/amazonaws/athena/connectors/teradata/TeradataQueryStringBuilder.java
```diff
@@ -67,4 +69,19 @@ protected String appendLimitOffset(Split split, Constraints constraints)
...
+        return TeradataSqlDialect.DEFAULT;
+    }
+
+    /**
+     *  Teradata does not support FETCH clause, so removing it from the query
+     */
+    @Override
+    protected boolean removeFetch()
+    {
+        return true;
```
burhan94: how come calcite doesn't remove the fetch itself?

burhan94:
has merge conflicts

burhan94 (commented):
## Code Review

### 🔴 Critical Issues

1. **`removeFetch()` is dead code** (JdbcSplitQueryBuilder.java, TeradataQueryStringBuilder.java) — The method is added and overridden but never called anywhere. The stated purpose ("Teradata does not support FETCH clause") is NOT achieved. Queries with FETCH will still fail on Teradata.

2. **Missing Javadoc** on base class `removeFetch()` method.

### 🟡 Medium Issues

- Unrelated whitespace change in JdbcSplitQueryBuilder.java
- Potential null safety concern in `getCredentialProvider()` chain (TeradataMetadataHandler.java)
- Test naming doesn't follow `testMethodName_Scenario_ExpectedBehavior` format
- No edge case tests for empty strings or null partition properties

### 🟢 What's Good

- `getSqlDialect()` correctly returns `TeradataSqlDialect.DEFAULT`
- Tests use real objects appropriately
- Methods are concise

The main blocker: either wire up `removeFetch()` so it actually removes FETCH clauses from generated SQL, or remove the dead code. As-is, Teradata queries with FETCH will still fail.

burhan94 (commented):
## Re-Review (April 3)

Original issues resolved — `removeFetch()` replaced with proper `TeradataSqlDialect.unparseOffsetFetch()`. Good approach.

### 🔴 New Issues (shared with #3303)

1. **NPE risk in handleSQLException** (GenericJdbcConnectionFactory.java) — `e.getMessage().contains()` will NPE if getMessage() returns null.

2. **getPooledConnection() returns null after exception** (GenericJdbcConnectionFactory.java) — If exception doesn't match patterns in handleSQLException(), method silently returns, leading to null return. Should re-throw the original exception.

These are in shared code (GenericJdbcConnectionFactory) — fixing in one PR fixes both.

AbdulR3hman (dismissed):
still fails the codecov rate. can you please look into that.

ejeffrli (approved):

ZhitongYan (approved):

DilipReddyGaddam:
Added query federation support for Teradata connector. This PR also handles Catalog casing filter connection property.

PR #3300, opened 2026-02-25
Neptune Gremlin QPT Enhancement for valueMap Traversal Queries

Trianz-Akshay:
*Issue #, if available:*

*Description of changes:*
**Overview**
This PR adds full valueMap() support for Gremlin Query Pushdown (QPT) in the Athena–Neptune connector.
Row writers now correctly handle both:

- valueMap()-style results (where property values are Lists), and
- project().by()-style results (scalars or nested Maps).

This avoids ClassCastException and enables consistent behavior across different Gremlin result shapes.

**Problem**
Gremlin QPT can return property data in different shapes depending on the traversal:
_valueMap()_ → properties are returned as Map<String, List<Object>>
Example: name -> [John]
_project().by() / by(valueMap())_ → properties can be scalars or Map values
Example: all_properties -> Map

Previously, row writers assumed a single result shape when reading field values.
This could lead to:

- ClassCastException
- Incorrect field handling when the result format differed

**Changes**
Row writers updated to use the normalizer
All extractors in the following classes now retrieve field values using:
FieldValueNormalizer.toValueList(...)
The extractors then read from the resulting list, ensuring compatibility with both valueMap() and project()-style results.
Updated classes:

1. CustomSchemaRowWriter – Custom schema vertex/edge row writing
2. VertexRowWriter – Vertex row writing
3. EdgeRowWriter – Edge row writing

Find attached document for [supported query types and limitations](https://github.com/user-attachments/files/25545370/Neptune_Gremlin_QPT_valueMap_Support_Updated.docx) as well as [functional test results](https://github.com/user-attachments/files/25545393/NEPTUNE_FUNCTIONAL_TEST.xlsx).

### DIFF
--- a/athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/CustomSchemaRowWriterTest.java
+++ b/athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/CustomSchemaRowWriterTest.java
@@ -36,6 +36,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 import static org.junit.Assert.assertEquals;
@@ -49,6 +50,17 @@ import static org.junit.Assert.assertTrue;
  */
 public class CustomSchemaRowWriterTest
 {
+    private static final String FLAG = "flag";
+    private static final String TIMESTAMP = "timestamp";
+    private static final String NUMBER_FIELD = "numberField";
+    private static final String BIG_INT_FIELD = "bigIntField";
+    private static final String FLOAT_FIELD = "floatField";
+    private static final String DOUBLE_FIELD = "doubleField";
+    private static final String NAME = "name";
+    private static final String TAGS = "tags";
+    private static final String ALL_PROPERTIES = "all_properties";
+    private static final String ID = "id";
+
     private BlockAllocatorImpl allocator;
 
     @Before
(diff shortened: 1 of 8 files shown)


ejeffrli (dismissed):

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/CustomSchemaRowWriterTest.java
```diff
@@ -0,0 +1,329 @@
...
+            reader.setPosition(0);
+            if (!reader.isSet()) {
+                return null;
+            }
+            return reader.readObject();
+        }
+    }
+
+    @Test
+    public void writeRowTemplate_varchar_valueMapList_writesFirstElement() throws Exception
```
fal-bharadwaj, line 108: Test coverage for CustomSchemaRowWriter is ~57.7%, can we add more test classes to account for the missing code paths? Let's have test cases for scalar input, single element list input, null/empty etc.

fal-bharadwaj (changes requested):

athena-neptune/src/test/java/com/amazonaws/athena/connectors/neptune/propertygraph/rowwriters/CustomSchemaRowWriterTest.java
VenkatasivareddyTR: Hi @fal-bharadwaj, thanks for the review. The Neptune connector currently has low baseline coverage, which is reflected in this class. We have covered all new and updated code and added tests for missing paths here (coverage is 84.44% now). Since there is a separate task dedicated to increasing Neptune's overall coverage, we will add more tests to the remaining classes as part of that task.

fal-bharadwaj (approved):

burhan94 (commented):
## Code Review

### ✅ Looks Good

New `FieldValueNormalizer` utility handles both valueMap() and project().by() results cleanly. Good refactoring of row writers to use List interface. Comprehensive tests. Good null handling with early returns.

hareeshkumar-t (approved):

PR #3303, opened 2026-03-02
Substrait Implementation for Oracle & SqlServer + Catalog Casing for JDBC Connectors + Resolve Too Many Connections error

stelanof88:
*Description of changes:*

- Adds Substrait/Calcite query plan support for Oracle and SQL Server connectors
- Introduces a CatalogCasingFilter mechanism for all four JDBC connectors (PostgreSQL, MySQL, Oracle, SQL Server) to handle case-sensitive identifier quoting. Also enabled for RedShift as it inherits MetadataHandler and QueryBuilder from Postgres
- A shared JdbcCasingSqlDialect base class consolidates the quoting logic, with each connector specifying only its database-specific quote character.
- When GLUE_MANAGED_CONNECTION=true is set in AthenaProperties, JDBC connectors use direct DriverManager connections instead of HikariCP pooling to address Too Many Connections error


ZhitongYan:
leveraging Hikari Pooled connections seems always to be the preferred choice v.s. going back to a more risk / raw implementation of DriverManager and manage on our own. 

this approach has several drawbacks:
1. connection re-open and re-close takes significant time compared to pooled approach
2. we are hand over the connection limit management to downstream database which might cause the database itself overwhelmed. 
3. we lost the benefit of Hikari rich features managing connections for us. 

Can we try just tweak the connections like 

```
HikariConfig config = new HikariConfig();
config.setMaximumPoolSize(50);  // adjust based on your workload
config.setMinimumIdle(10);
config.setConnectionTimeout(30000);  // 30s - fail fast if pool is exhausted
config.setIdleTimeout(600000);       // 10min - clean up idle connections
config.setMaxLifetime(1800000);      // 30min - recycle connections

```

stelanof88:
yeah I tried using config values to get hikari to work but


we actually want `setMaximumPoolSize` to be a lower value as this dictates how many connections each lambda holds.
For connection re-open and re-close, through performance tests there wasn't too much overhead difference, only about ~1 second max


The issue with HikariCP is that even with most aggressive settings allowed (pool size 1, idle timeout 10s, max lifetime 30s), connections still linger for ~4 minutes  on RDS. HikariCP also enforces minimum values on these timeouts so we can't go lower to try and have them close after use. As the lambdas scale up with concurrent queries, each holding connections that don't release for minutes, we quickly exhaust the RDS max_connections limit.

I tried getting HikariCP to work as it does have better features and performance then just the DriverManagersharing a single connection pool between metadata and record handlers within each Lambda instead of each creating their own individually but it looks like the multiplexing creates each pool independently meaning we'd have to have a large codebase refactor to get this to work

burhan94 (commented):
## Code Review

### 🔴 Critical Issues

1. **`getPooledConnection()` returns null on error** (GenericJdbcConnectionFactory.java) — After catching SQLException, returns `null` instead of throwing. All callers will get NPE.

2. **WHERE clause detection matches inside string literals** (JdbcSplitQueryBuilder.java) — `sql.toUpperCase().contains("WHERE")` will match WHERE inside string literals and comments, producing invalid SQL. Use regex word boundary: `Pattern.compile("\\bWHERE\\b", Pattern.CASE_INSENSITIVE)`.

3. **ResultSet consumed before reading min/max** (JdbcMetadataHandler.java) — `minMaxResultSet.getLong(1/2)` read after ResultSet may be consumed by `splitterFactory.getSplitter()`, causing incorrect split values.

4. **Missing class-level Javadoc** on new classes: JdbcCasingSqlDialect, MysqlSqlDialect, OracleSqlDialect, PostgresqlSqlDialect, SqlServerDialect.

### 🟡 Medium Issues

- Duplicate code: catalog casing filter propagation copied across 5 metadata handlers. Should extract to base class helper.
- Missing Javadoc on new public methods: `getDirectConnection()`, `getPooledConnection()`, `getSqlDialect(boolean)`
- INFO level logging on every split is noisy; should be DEBUG

### Missing Test Coverage

- `GenericJdbcConnectionFactory` direct connection path
- `JdbcSplitQueryBuilder` split clause injection
- `JdbcMetadataHandler.getSplitClauses()` range-too-small logic
- Edge cases: null configOptions, empty split clauses

### 🟢 What's Good

- Catalog casing filter is a needed feature for managed connectors
- Direct JDBC connection path for Glue managed connections is well-motivated
- Custom SQL dialects follow the established pattern

Please address the critical issues before merge. The null return from `getPooledConnection()` is the highest priority — it will cause NPEs in production.

stelanof88:
@burhan94 addressed these issues

burhan94 (commented):
## Re-Review (April 3)

Good progress — 3 of 4 original issues fixed.

### ✅ Fixed
- `getPooledConnection()` now throws instead of returning null ✓
- Min/max values read before `getSplitter()` ✓
- Class-level Javadoc added to all dialect classes ✓
- WHERE detection improved to regex word boundary ✓

### 🔴 New Issue

1. **NPE risk in handleSQLException** (GenericJdbcConnectionFactory.java) — `e.getMessage().contains(...)` will NPE if `getMessage()` returns null. Add:
```java
String message = e.getMessage();
if (message != null && message.contains("Name or service not known")) { ... }
```

### 🟡 Minor
- Catalog casing filter check is copy-pasted across 4 metadata handlers — consider extracting to base class.

Almost there. Fix the NPE and this is good to go.

burhan94 (approved):
## Re-Review (April 3 - Round 3)

### ✅ NPE Fix Verified

The `handleSQLException()` fix is correct — properly checks `message != null` before `.contains()`.

### All Critical Issues Resolved
- getPooledConnection() throws instead of returning null ✓
- Min/max values read before getSplitter() ✓
- WHERE detection uses regex word boundary ✓
- Class-level Javadoc on all new dialect classes ✓
- NPE in handleSQLException fixed ✓

### 🟢 Minor (non-blocking)
- Catalog casing filter check is duplicated across 5 locations — consider extracting to a helper in JdbcMetadataHandler in a follow-up.

LGTM. Approving.

ejeffrli (approved):

PR #3335, opened 2026-03-23
Add Query Federation Support for SAP HANA

DilipReddyGaddam:
Issue #, if available:

Description of changes:
Changes in this are required to support Query federation for SAP Hana connector. We have to implement custom logic of SAPHanaSqlDialect to support SQL query generation using Substrait libraries.

### DIFF
--- a/athena-saphana/src/main/java/com/amazonaws/athena/connectors/saphana/SAPHanaSqlDialect.java
+++ b/athena-saphana/src/main/java/com/amazonaws/athena/connectors/saphana/SAPHanaSqlDialect.java
@@ -1,0 +1,165 @@
+/*-
+ * #%L
+ * athena-saphana
+ * %%
+ * Copyright (C) 2019 - 2026 Amazon Web Services
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package com.amazonaws.athena.connectors.saphana;
+
+import org.apache.calcite.avatica.util.Casing;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.sql.SqlBasicCall;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.dialect.PostgresqlSqlDialect;
+import org.apache.calcite.sql.fun.SqlLibraryOperators;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * Calcite SQL dialect for SAP HANA (Calcite 1.40.0).
+ *
+ * Behavior:
+ *  - Unquoted identifiers are folded to UPPERCASE
+ *  - Quoted identifiers preserve case
+ *  - BOOLEAN MAX/MIN rewritten to BOOL aggregations
+ *  - OFFSET/FETCH rendered as LIMIT/OFFSET
+ */
+public class SAPHanaSqlDialect extends PostgresqlSqlDialect
+{
+    public static final SqlDialect.Context DEFAULT_CONTEXT =
+            SqlDialect.EMPTY_CONTEXT
+                    .withDatabaseProduct(DatabaseProduct.UNKNOWN)
+                    .withIdentifierQuoteString("\"")
+                    .withUnquotedCasing(Casing.TO_UPPER)
+                    .withQuotedCasing(Casing.UNCHANGED)
+                    .withCaseSensitive(false);
+
+    public static final SAPHanaSqlDialect DEFAULT =
+            new SAPHanaSqlDialect(DEFAULT_CONTEXT, false);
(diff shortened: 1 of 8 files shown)


burhan94 (commented):
## Code Review

### 🔴 Critical Issues

1. **quoteIdentifier doesn't escape embedded quotes** (SAPHanaSqlDialect.java) — If an identifier contains `"`, the generated SQL breaks. Should escape with `identifier.replace("\"", "\"\"")`. This is also a potential SQL injection vector.

2. **Behavioral change in credential fetching** (SaphanaMetadataHandler.java) — Changed from `getDatabaseConnectionConfig().getSecret()` to `secretName` parameter. Needs verification this is intentional.

3. **Dead code** — `rewriteMaxMinExpr` is public but never called. Either integrate it or remove it.

### 🟡 Medium Issues

- Tests don't verify actual SQL output for LIMIT/OFFSET
- Mixing JUnit 4 and JUnit 5 annotations in tests
- Uses `Mockito.any()` matchers instead of specific captures
- `log.info` on every query could be noisy; consider `log.debug`

### Potential Edge Cases

- `isCatalogCasingFilterApplied`: NPE risk if `split.getProperties()` returns null
- `quoteIdentifier` with `upperCaseFilterApplied=true` returns unquoted identifier — breaks on reserved words

### 🟢 What's Good

- Custom Calcite dialect approach is correct for HANA
- CHAR_LENGTH→LENGTH rewrite is a nice touch

Please address the critical issues before merge.

athena-saphana/src/main/java/com/amazonaws/athena/connectors/saphana/SaphanaRecordHandler.java
```diff
@@ -122,7 +122,8 @@ public CredentialsProvider createCredentialsProvider(String secretName, AwsReque
         return CredentialsProviderFactory.createCredentialProvider(
                 getDatabaseConnectionConfig().getSecret(),
```
ejeffrli, line 123: should this be using secretName instead?

athena-saphana/src/main/java/com/amazonaws/athena/connectors/saphana/SAPHanaSqlDialect.java
```diff
@@ -0,0 +1,165 @@
...
+    {
+        // SAP HANA does not support approximate distinct aggregates
+        return false;
+    }
+
+    @Override
+    public StringBuilder quoteIdentifier(StringBuilder buf, String identifier)
+    {
+        if (upperCaseFilterApplied) {
+            return buf.append(identifier);
```
ejeffrli, line 159: should this wrap identifier in double quotes like the lowercase implementation? The tests are expecting different results for some reason

DilipReddyGaddam: Default case in SAP Hana is uppercase and we don't need to wrap the identifier in quote. Only lowercase table names and column names needs to be quoted.

athena-saphana/src/main/java/com/amazonaws/athena/connectors/saphana/SaphanaRecordHandler.java
DilipReddyGaddam: Validated the both secretName and  getDatabaseConnectionConfig().getSecret() they hold same value. this is not newly added and it works fine. I think that should be okay

ZhitongYan (approved):

athena-saphana/src/main/java/com/amazonaws/athena/connectors/saphana/SAPHanaSqlDialect.java
ejeffrli: what about escaping characters? The quotes for lowercase would allow complex names. Why can't both casing be quoted here?

athena-saphana/src/main/java/com/amazonaws/athena/connectors/saphana/SAPHanaSqlDialect.java
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * Calcite SQL dialect for SAP HANA (Calcite 1.40.0).
+ *
+ * Behavior:
+ *  - Unquoted identifiers are folded to UPPERCASE
+ *  - Quoted identifiers preserve case
+ *  - BOOLEAN MAX/MIN rewritten to BOOL aggregations
+ *  - OFFSET/FETCH rendered as LIMIT/OFFSET
+ */
+public class SAPHanaSqlDialect extends PostgresqlSqlDialect
```
ejeffrli, line 46: Why are we extending PostgresqlSqlDialect and not SqlDialect? Their dialects can differ

athena-saphana/src/main/java/com/amazonaws/athena/connectors/saphana/SAPHanaSqlDialect.java
DilipReddyGaddam: Column names in the Substrait plan are always lowercase. When upperCaseFilterApplied=true, identifiers in the catalog are uppercase, so we emit them unquoted — HANA automatically folds unquoted identifiers to uppercase, matching the catalog. If we quoted them, it would preserve the lowercase casing from the plan, causing a mismatch with the actual column names in HANA.

For the lowercase path, we quote to preserve the exact casing.

Regarding escaping: the uppercase path only applies when the catalog casing filter confirms all identifiers are simple uppercase names. Any special characters in identifiers would go through the lowercase/quoted path where double-quote wrapping handles them. Reserved words are not a concern here since HANA's catalog metadata returns actual column names, not reserved keywords.

athena-saphana/src/main/java/com/amazonaws/athena/connectors/saphana/SAPHanaSqlDialect.java
DilipReddyGaddam: PostgresqlSqlDialect provides the double-quote identifier quoting and catalog casing filter logic that HANA also needs. SAPHanaSqlDialect overrides quoteIdentifier and unparseCall for HANA-specific
  behavior (uppercase unquoted identifiers, CHAR_LENGTH→LENGTH rewrite, BOOLEAN MAX/MIN rewrite). If we want to clean up the hierarchy to make the shared logic more explicit (e.g., extracting a common base), I can send a follow-up PR refactoring all connectors.

ejeffrli: sounds good. we should extend based off shared logic

ejeffrli (approved):

PR #3353, opened 2026-04-03
StringTemplate enhancement – BigQuery Connector

Trianz-Akshay:
*Issue #, if available:*

*Description of changes:*

This PR enhances stringTemplate through the following improvements:
- **Identifier Quoting**: Schema, table, and column identifiers are now wrapped in backticks (e.g., `table_name`).
- **String literals**: All user-provided text and date values are now correctly escaped.

Please find the attached detailed description and functional test documentation.
[BIGQUERY_FUNCTIONAL_TEST.xlsx](https://github.com/user-attachments/files/26460393/BIGQUERY_FUNCTIONAL_TEST.xlsx)

### DIFF
--- a/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryFederationExpressionParser.java
+++ b/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryFederationExpressionParser.java
@@ -49,8 +49,6 @@ import static com.amazonaws.athena.connector.lambda.domain.predicate.expression.
  */
 public class BigQueryFederationExpressionParser extends FederationExpressionParser
 {
-    private static final String quoteCharacter = "'";
-
     public String writeArrayConstructorClause(List<String> arguments)
     {
         return BigQuerySqlUtils.renderTemplate("comma_separated_list_with_parentheses", Map.of("items", arguments));
@@ -96,6 +94,12 @@ public class BigQueryFederationExpressionParser extends FederationExpressionPars
         return mapFunctionToDataSourceSyntax(functionName, functionCallExpression.getType(), arguments);
     }
 
+    @Override
+    public String parseVariableExpression(VariableExpression variableExpression)
+    {
+        return BigQuerySqlUtils.backtickQuotedIdentifier(variableExpression.getColumnName());
+    }
+
     public String parseConstantExpression(ConstantExpression constantExpression)
     {
         Block values = constantExpression.getValues();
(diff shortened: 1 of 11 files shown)


ejeffrli:
From Claude:
**`escapeForSingleQuotedStringLiteral` uses ANSI quote-doubling (`''`), but BigQuery string literals use backslash escaping (`\'`).**

`backtickQuotedIdentifier` in this PR correctly uses backslash escaping — but the string-literal helper uses doubling, which in GoogleSQL:
- **doesn't prevent injection** — a trailing backslash (`foo\` → `'foo\'`) escapes the closing quote, so trailing input runs as SQL (reachable via Storage API `row_restriction` and constant pushdown — both non-parameterized);
- **breaks valid values** — `O'Brien` → `'O''Brien'` parses as two adjacent strings → syntax error.

Fix (matches the identifier helper):
```java
return value.replace("\\", "\\\\").replace("'", "\\'");
```
The `?`-parameterized `BigQueryPredicateBuilder` path is unaffected.

ritiktrianz:
> From Claude: **`escapeForSingleQuotedStringLiteral` uses ANSI quote-doubling (`''`), but BigQuery string literals use backslash escaping (`\'`).**
> 
> `backtickQuotedIdentifier` in this PR correctly uses backslash escaping — but the string-literal helper uses doubling, which in GoogleSQL:
> 
>     * **doesn't prevent injection** — a trailing backslash (`foo\` → `'foo\'`) escapes the closing quote, so trailing input runs as SQL (reachable via Storage API `row_restriction` and constant pushdown — both non-parameterized);
> 
>     * **breaks valid values** — `O'Brien` → `'O''Brien'` parses as two adjacent strings → syntax error.
> 
> 
> Fix (matches the identifier helper):
> 
> ```java
> return value.replace("\\", "\\\\").replace("'", "\\'");
> ```
> 
> The `?`-parameterized `BigQueryPredicateBuilder` path is unaffected.

Hi @ejeffrli , thanks for the review.
We verified that BigQuery string literals use backslash escaping rather than ANSI quote-doubling, and updated `escapeForSingleQuotedStringLiteral` to align with `backtickQuotedIdentifier`.
This handles values like `O'Brien` and trailing backslashes correctly in BigQuery. Unit tests have been updated accordingly.

ritiktrianz:
> Are we continuing the use of stringtemplate in bigquery?

Hi @ejeffrli , This PR hardens the existing StringTemplate usage with consistent identifier quoting and string-literal escaping to improve correctness in the current query-building paths. The ?-parameterized BigQueryPredicateBuilder path remains unchanged.
 
If the team would prefer to move away from StringTemplate entirely, please let us know-we can revert StringTemplate changes.

ejeffrli (approved):

PR #3380, opened 2026-04-15
Added unit tests in athena-hbase connector.

ritiktrianz:
*Issue #, if available:*

*Description of changes:*
 Added unit tests in athena-hbase connector.


ejeffrli:
Thanks for the coverage work — the given/when/then migration is consistent and the Kerberos S3-path / QPT / error-branch additions are real, valuable coverage. A few test-quality items before merge (none are functional/security risks):

**1. Tautological `>= 0` assertions (`HbaseRecordHandlerTest`).** `getRecordCount()` returns `records.getRowCount()`, which is backed by Arrow's non-negative row counter, so it is always `>= 0` on the success path; on error the handler throws (it never returns a negative count). That makes `assertTrue(response.getRecordCount() >= 0)` pass for every possible execution, including a regression that returns the wrong number of rows — so these tests exercise the code path (good for coverage) but can't actually fail. Affected:
- `doReadRecords_withNoConstraints_returnsAllRecords` (feeds `makeResults(10)`, name says "all")
- `doReadRecords_withQueryPassthroughWithFilter_returnsFilteredRecords`
- `doReadRecords_withRowKeyConstraint_returnsRecordsMatchingRowKey`
- `doReadRecords_withSpecificSchema_returnsRecordsForNativeSchema`
- `doReadRecords_withStructField_returnsRecordsWithStructField`
- `doReadRecords_withOnlyRowColumn_returnsRecordsWithRowColumnOnly`

Suggest asserting the exact expected count (e.g. `assertEquals(10, response.getRecordCount())`) where the result is deterministic, or at least `> 0` for the "returns records" cases.

**2. Misnamed test (`HbaseMetadataHandlerTest`).** `doGetTable_withGlueException_fallsBackToHBaseSchema` — the body comment notes it can't easily mock `super.doGetTable` and tests the null-Glue case instead, so it exercises the `origSchema == null` path and effectively duplicates `doGetTable_withNullGlue_fallsBackToHBaseSchema`; it never exercises a Glue exception. Suggest renaming it to reflect the null-Glue path, or dropping it as a duplicate.

**3. Redundant override (`HbaseEnvironmentProperties`, `main/`).** The added `getEnvMap()` override is a pure `return super.getEnvMap();` at the same visibility. The base `EnvironmentProperties.getEnvMap()` is already `protected @VisibleForTesting`, so the test subclass can override it directly — this redeclaration doesn't add a seam and can be removed. (The `getEnvironmentProperties()` seams added to the two `HbaseConnectionFactory` classes are fine — those are the ones doing the actual work; keep them.)

athena-hbase/src/test/java/com/amazonaws/athena/connectors/hbase/HbaseRecordHandlerTest.java
```diff
@@ -213,26 +231,112 @@ public void doReadRecordsNoSpill()
...
+        logger.info("doReadRecords_withQueryPassthroughWithFilter_returnsFilteredRecords: enter");
+        setupMockScanner();
+        Map<String, String> qptArguments = createQptArguments("RowFilter(=,'substring:test')");
+        ReadRecordsRequest request = createReadRecordsRequestWithQpt(schemaForRead, qptArguments);
 
+        RecordResponse rawResponse = handler.doReadRecords(allocator, request);
+        assertTrue("Response should be ReadRecordsResponse", rawResponse instanceof ReadRecordsResponse);
         ReadRecordsResponse response = (ReadRecordsResponse) rawResponse;
-        logger.info("doReadRecordsNoSpill: rows[{}]", response.getRecordCount());
+        assertTrue("Record count should be greater than or equal to 0", response.getRecordCount() >= 0);
```
christov: What is the value of having equality here as well? Do we want this assertion to pass if 0 records were returned? Same question for the other instances of this assertion in this file.

christov (changes requested):

anushatrianz:
> Thanks for the coverage work — the given/when/then migration is consistent and the Kerberos S3-path / QPT / error-branch additions are real, valuable coverage. A few test-quality items before merge (none are functional/security risks):
> 
> **1. Tautological `>= 0` assertions (`HbaseRecordHandlerTest`).** `getRecordCount()` returns `records.getRowCount()`, which is backed by Arrow's non-negative row counter, so it is always `>= 0` on the success path; on error the handler throws (it never returns a negative count). That makes `assertTrue(response.getRecordCount() >= 0)` pass for every possible execution, including a regression that returns the wrong number of rows — so these tests exercise the code path (good for coverage) but can't actually fail. Affected:
> 
> * `doReadRecords_withNoConstraints_returnsAllRecords` (feeds `makeResults(10)`, name says "all")
> * `doReadRecords_withQueryPassthroughWithFilter_returnsFilteredRecords`
> * `doReadRecords_withRowKeyConstraint_returnsRecordsMatchingRowKey`
> * `doReadRecords_withSpecificSchema_returnsRecordsForNativeSchema`
> * `doReadRecords_withStructField_returnsRecordsWithStructField`
> * `doReadRecords_withOnlyRowColumn_returnsRecordsWithRowColumnOnly`
> 
> Suggest asserting the exact expected count (e.g. `assertEquals(10, response.getRecordCount())`) where the result is deterministic, or at least `> 0` for the "returns records" cases.
> 
> **2. Misnamed test (`HbaseMetadataHandlerTest`).** `doGetTable_withGlueException_fallsBackToHBaseSchema` — the body comment notes it can't easily mock `super.doGetTable` and tests the null-Glue case instead, so it exercises the `origSchema == null` path and effectively duplicates `doGetTable_withNullGlue_fallsBackToHBaseSchema`; it never exercises a Glue exception. Suggest renaming it to reflect the null-Glue path, or dropping it as a duplicate.
> 
> **3. Redundant override (`HbaseEnvironmentProperties`, `main/`).** The added `getEnvMap()` override is a pure `return super.getEnvMap();` at the same visibility. The base `EnvironmentProperties.getEnvMap()` is already `protected @VisibleForTesting`, so the test subclass can override it directly — this redeclaration doesn't add a seam and can be removed. (The `getEnvironmentProperties()` seams added to the two `HbaseConnectionFactory` classes are fine — those are the ones doing the actual work; keep them.)

Hi @ejeffrli, thanks for the review and for the detailed feedback.

We've addressed all three items:

1. **Tautological `>= 0` assertions** — Replaced them with exact `assertEquals(...)` counts in `HbaseRecordHandlerTest`.
2. **Misnamed Glue exception test** — Removed `doGetTable_withGlueException_fallsBackToHBaseSchema` as a duplicate of the null-Glue fallback path.
3. **Redundant `getEnvMap()` override** — Removed the no-op override from `HbaseEnvironmentProperties`; tests continue to override `getEnvMap()` via the base `EnvironmentProperties` seam.

Please review and let us know if you have any further feedback.

athena-hbase/src/test/java/com/amazonaws/athena/connectors/hbase/HbaseRecordHandlerTest.java
anushatrianz: > What is the value of having equality here as well? Do we want this assertion to pass if 0 records were returned? Same question for the other instances of this assertion in this file.

Hi @christov, thanks for the review.

Agreed — `getRecordCount() >= 0` is not a meaningful assertion here, and these tests should not pass when 0 records are returned unexpectedly.

We've updated the affected tests in `HbaseRecordHandlerTest` to assert exact expected counts .
Please review and let us know if you have any further feedback.

PR #3382, opened 2026-04-16
Fix PermissionsBoundary placement in connector SAM templates

AbdulR3hman:
### Description:

   Corrected PermissionsBoundary configuration in 5 connector templates (mysql, oracle, postgresql, sqlserver, clickhouse) so that it is applied to the
   FunctionRole IAM role resource rather than the AWS::Serverless::Function resource. Removed the redundant property from the function resource.

   Affected files (.yaml and -package.yaml for each):

   - athena-mysql
   - athena-oracle
   - athena-postgresql
   - athena-sqlserver
   - athena-clickhouse

   No new parameters or conditions introduced. Backward compatible — no behavior change when PermissionsBoundaryARN is not provided.

### DIFF
--- a/athena-clickhouse/athena-clickhouse-package.yaml
+++ b/athena-clickhouse/athena-clickhouse-package.yaml
@@ -77,7 +77,6 @@ Resources:
       Runtime: java11
       Timeout: !Ref LambdaTimeout
       MemorySize: !Ref LambdaMemory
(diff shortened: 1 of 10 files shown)


AbdulR3hman:
#todo the `XXXX-package.yaml` are no longer required; we need to clean them up in a different PR.

aimethed (approved):

chngpe (approved):

PR #3430, opened 2026-05-18
Add Backtick Quote Identifier for Impala Connector

anushatrianz:
*Issue #, if available:*

*Description of changes:*

Replaced PreparedStatement (no bind parameters) with Statement and backtick-quoted identifiers for metadata commands.

**Identifier quoting:** Schema, table, and column names are wrapped in backticks

Please find the attached functional test documentation.
[CLOUDERAIMPALA_FUNCTIONAL_TEST_2026-05-18.xlsx](https://github.com/user-attachments/files/28012982/CLOUDERAIMPALA_FUNCTIONAL_TEST_2026-05-18.xlsx)

### DIFF
--- a/athena-cloudera-impala/src/main/java/com/amazonaws/athena/connectors/cloudera/ImpalaMetadataHandler.java
+++ b/athena-cloudera-impala/src/main/java/com/amazonaws/athena/connectors/cloudera/ImpalaMetadataHandler.java
@@ -62,7 +62,6 @@ import software.amazon.awssdk.services.athena.AthenaClient;
 import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
 
 import java.sql.Connection;
-import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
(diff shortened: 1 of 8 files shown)


ejeffrli (approved):

athena-cloudera-impala/src/main/java/com/amazonaws/athena/connectors/cloudera/ImpalaMetadataHandler.java
```diff
@@ -295,9 +294,9 @@ protected Schema getSchema(Connection jdbcConnection, TableName tableName, Schem
         SchemaBuilder schemaBuilder = SchemaBuilder.newBuilder();
         try (ResultSet resultSet = getColumns(jdbcConnection.getCatalog(), tableName, jdbcConnection.getMetaData());
              Connection connection = getJdbcConnectionFactory().getConnection(getCredentialProvider())) {
-            try (PreparedStatement psmt = connection.prepareStatement(
-                GET_METADATA_QUERY + tableName.getQualifiedTableName().toUpperCase())) {
-                Map<String, String> hashMap = getMetadataForGivenTable(psmt);
+            try (Statement stmt = connection.createStatement()) {
```
AbdulR3hman, line 297: why have we stopped used prepared statements instead? those are usually better against sql injection?

AbdulR3hman (changes requested):

athena-cloudera-impala/src/main/java/com/amazonaws/athena/connectors/cloudera/ImpalaMetadataHandler.java
anushatrianz: > why have we stopped used prepared statements instead? those are usually better against sql injection?

Hi @AbdulR3hman, thanks for the review.
 
Agreed that PreparedStatement is usually preferred for values. In this case it did not help: the old code still concatenated the table name into the SQL with no `?` placeholders, so it only looked safer.
 
JDBC bind parameters apply to values, not identifiers. `DESCRIBE` and `SHOW FILES IN` require a table identifier, which cannot be bound with `?`. The safe approach is to quote/escape the identifier, which we do via `ImpalaUtils.quoteIdentifier()` / `qualifiedTableForMetadataSql()` (wrapped in backticks, with any embedded backticks escaped by doubling)  , and run that SQL with a `Statement`.

AbdulR3hman (approved):

PR #3431, opened 2026-05-18
Add Backtick Quote Identifier for HIVE Connector

anushatrianz:
*Issue #, if available:*

*Description of changes:*

Replaced PreparedStatement (no bind parameters) with Statement and backtick-quoted identifiers for metadata commands.

**Identifier quoting:** Schema, table, and column names are wrapped in backticks

**String literals:** Table names in SHOW TABLE EXTENDED ... LIKE '...' are single-quoted with escaped apostrophes (' → '') through likePatternLiteral.

Please find the attached functional test documentation.
[CLOUDERAHIVE_FUNCTIONAL_TEST_2026-05-18.xlsx](https://github.com/user-attachments/files/28013016/CLOUDERAHIVE_FUNCTIONAL_TEST_2026-05-18.xlsx)

### DIFF
--- a/athena-cloudera-hive/src/main/java/com/amazonaws/athena/connectors/cloudera/HiveUtils.java
+++ b/athena-cloudera-hive/src/main/java/com/amazonaws/athena/connectors/cloudera/HiveUtils.java
@@ -20,6 +20,8 @@
 package com.amazonaws.athena.connectors.cloudera;
 
 import com.amazonaws.athena.connector.lambda.domain.TableName;
+import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.dialect.HiveSqlDialect;
 
 import static com.amazonaws.athena.connectors.cloudera.HiveConstants.HIVE_QUOTE_CHARACTER;
 
@@ -29,6 +31,8 @@ import static com.amazonaws.athena.connectors.cloudera.HiveConstants.HIVE_QUOTE_
  */
 public class HiveUtils
 {
+    private static final SqlDialect HIVE_SQL_DIALECT = HiveSqlDialect.DEFAULT;
+
     private HiveUtils()
     {
     }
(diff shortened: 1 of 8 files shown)


athena-cloudera-hive/src/main/java/com/amazonaws/athena/connectors/cloudera/HiveUtils.java
```diff
@@ -0,0 +1,64 @@
...
+        return quoteIdentifier(tableName.getSchemaName().toUpperCase())
+                + "." + quoteIdentifier(tableName.getTableName().toUpperCase());
+    }
+
+    /**
+     * Single-quoted pattern for {@code SHOW TABLE EXTENDED IN ... LIKE '...'}. Hive uses {@code *}
+     * (not {@code %}) as the wildcard in {@code SHOW} commands; valid table names do not include
+     * wildcards, so only single quotes are escaped and the name is upper-cased to match prior behavior.
+     */
+    public static String likePatternLiteral(String pattern)
```
ejeffrli, line 65: From AI below
```
A SQL string literal is delimited by single quotes: '...'. The hard question is always: how do you put a literal single-quote character inside the string without it ending the string? There are two conventions:

Convention A — doubling (ANSI SQL standard):
code
'it''s here'     → the value:  it's here
Two quotes in a row mean "one literal quote, keep going."

Convention B — backslash escaping (C-style):
code
'it\'s here'     → the value:  it's here
A backslash before a quote means "this quote is literal, keep going."

Hive supports both. Its string literals honor C-style backslash escapes (\', \\, \n, …) in addition to quote-doubling. (This is the same behavior MySQL has, and it's the classic source of this bug class.) The likePatternLiteral helper only defends against Convention A. It never touches backslashes, so Convention B is wide open.
```

How are backslashes escaped?

ejeffrli: is there a better way to do this escaping? maybe a function from Hive or a dependency we can use?

anushatrianz: > From AI below
> 
> ```
> A SQL string literal is delimited by single quotes: '...'. The hard question is always: how do you put a literal single-quote character inside the string without it ending the string? There are two conventions:
> 
> Convention A — doubling (ANSI SQL standard):
> code
> 'it''s here'     → the value:  it's here
> Two quotes in a row mean "one literal quote, keep going."
> 
> Convention B — backslash escaping (C-style):
> code
> 'it\'s here'     → the value:  it's here
> A backslash before a quote means "this quote is literal, keep going."
> 
> Hive supports both. Its string literals honor C-style backslash escapes (\', \\, \n, …) in addition to quote-doubling. (This is the same behavior MySQL has, and it's the classic source of this bug class.) The likePatternLiteral helper only defends against Convention A. It never touches backslashes, so Convention B is wide open.
> ```
> 
> How are backslashes escaped?

Hi @ejeffrli, thanks for the review.
We’ve updated the code to handle both quote-doubling and backslash escaping. It now escapes backslashes first and then uses Calcite’s HiveSqlDialect.quoteStringLiteral() to handle single quotes correctly.
We also added unit tests covering cases with backslashes and verified that those escape sequences are handled properly.

anushatrianz: > is there a better way to do this escaping? maybe a function from Hive or a dependency we can use?

There isn’t a built-in Hive or Calcite helper that handles both ANSI quoting and backslash escaping together. quoteStringLiteral() from Calcite only handles quote doubling (' → ''), so we implemented the backslash escaping additionally.

aimethed (approved):

hareeshkumar-t (approved):

PR #3448, opened 2026-05-26
Fix Vertica connection timeout issue due to resource leakage.

ritiktrianz:
*Issue #, if available:*

*Description of changes:*
While testing the Vertica connector, we observed that after running some queries it throws a connection timeout error due to resource leakage. Fixed this by wrapping JDBC connections in try-with-resources to ensure connections are always closed and returned to the pool after use.
PFA Functional Testing and analysis document for reference.
[VERTICA_FUNCTIONAL_TEST_2026-05-26.xlsx](https://github.com/user-attachments/files/28254250/VERTICA_FUNCTIONAL_TEST_2026-05-26.xlsx)
[Vertica Connection Issue Analysis.docx](https://github.com/user-attachments/files/28877693/Vertica.Connection.Issue.Analysis.docx)

### DIFF
--- a/athena-vertica/src/main/java/com/amazonaws/athena/connectors/vertica/VerticaMetadataHandler.java
+++ b/athena-vertica/src/main/java/com/amazonaws/athena/connectors/vertica/VerticaMetadataHandler.java
@@ -104,6 +104,7 @@ public class VerticaMetadataHandler
     private static final String EMPTY_STRING = StringUtils.EMPTY;
     private static final String TABLE_SCHEMA = "TABLE_SCHEM";
     private static final String[] TABLE_TYPES = {"TABLE"};
+    private static final int VERTICA_STALE_CONNECTION_ERROR = 100023;
     private final QueryFactory queryFactory = new QueryFactory();
     private final VerticaSchemaUtils verticaSchemaUtils;
     private S3Client amazonS3;
@@ -153,12 +154,11 @@ public class VerticaMetadataHandler
             return new ListSchemasResponse(request.getCatalogName(), fetchSchemaNames());
         }
         catch (SQLException e) {
-            // VJDBC(100023) means a stale pooled connection was reused after a Lambda freeze/resume.
-            // Lambda freezes the JVM between requests; during that time Vertica may close its end of
-            // the idle pooled connection. When Lambda resumes, HikariCP hands back that dead connection
-            // and getTables() fails mid-protocol. Retry once to get a fresh connection.
-            if (e.getMessage() != null && e.getMessage().contains("100023")) {
-                logger.warn("doListSchemaNames: stale connection detected (VJDBC 100023), retrying with fresh connection");
+            // Error code 100023 indicates a stale pooled connection may have been
+            // reused after a Lambda freeze/resume cycle. Retry once to obtain a fresh connection.
+            if (e.getErrorCode() == VERTICA_STALE_CONNECTION_ERROR) {
+                logger.warn("doListSchemaNames: stale connection detected (errorCode={}), retrying with fresh connection",
+                        VERTICA_STALE_CONNECTION_ERROR);
                 return new ListSchemasResponse(request.getCatalogName(), fetchSchemaNames());
             }
             throw e;
--- a/athena-vertica/src/main/java/com/amazonaws/athena/connectors/vertica/VerticaMetadataHandler.java
+++ b/athena-vertica/src/main/java/com/amazonaws/athena/connectors/vertica/VerticaMetadataHandler.java
@@ -37,6 +37,7 @@ import com.amazonaws.athena.connector.lambda.metadata.GetTableRequest;
 import com.amazonaws.athena.connector.lambda.metadata.GetTableResponse;
 import com.amazonaws.athena.connector.lambda.metadata.ListSchemasRequest;
 import com.amazonaws.athena.connector.lambda.metadata.ListSchemasResponse;
+import com.amazonaws.athena.connector.lambda.exceptions.AthenaConnectorException;
 import com.amazonaws.athena.connector.lambda.metadata.optimizations.OptimizationSubType;
 import com.amazonaws.athena.connectors.jdbc.connection.DatabaseConnectionConfig;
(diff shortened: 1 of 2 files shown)


aimethed (dismissed):

athena-vertica/src/main/java/com/amazonaws/athena/connectors/vertica/VerticaMetadataHandler.java
```diff
@@ -146,20 +150,36 @@ public ListSchemasResponse doListSchemaNames(BlockAllocator allocator, ListSchem
             throws Exception
     {
         logger.info("doListSchemaNames: {}", request.getCatalogName());
+        try {
+            return new ListSchemasResponse(request.getCatalogName(), fetchSchemaNames());
+        }
+        catch (SQLException e) {
+            // Error code 100023 indicates a stale pooled connection may have been
+            // reused after a Lambda freeze/resume cycle. Retry once to obtain a fresh connection.
+            if (e.getErrorCode() == VERTICA_STALE_CONNECTION_ERROR) {
```
ejeffrli: why is only doListSchemas getting this error code retry? What about the other functions

ritiktrianz: Hi @ejeffrli , thanks for the review. 
We added the 100023 retry only in doListSchemaNames since that’s where we consistently reproduced the stale connection error after Lambda freeze/resume. Other methods mainly had pool exhaustion issues, which we fixed with try‑with‑resources. We kept the retry scoped to the observed path to limit changes, but can extend it to other methods if consistency is preferred.

ejeffrli:
The core fix — wrapping the JDBC connections in `doGetTable`, `getPartitions`, and `doGetSplits` in try-with-resources — is the right change and is what actually addresses the leak. 👍

My one piece of feedback is on the stale-connection retry in `doListSchemaNames`. I'd suggest a cleaner shape, and I agree with keeping this at the connector layer rather than pushing it into HikariCP:

**Why not rely on the pool.** Hikari's on-borrow validation isn't a reliable guard for this specific failure. The error the retry keys on (`100023` / `"Unexpected message type: RowDescription"`) surfaces with SQLState `S1000`, which is a *general* state, not the `08xxx` connection-exception class — so it won't reliably mark the connection broken on close. Combined with the default `aliveBypassWindow` (500ms) skipping `isValid()` on a quick re-borrow, a pool-validation approach can hand the same desynced physical connection right back. Doing the detect-and-retry in the connector is sturdier and keeps the behavior explicit.

**Two issues with the current form:**

1. **It's only on one of five call sites.** `doGetTable`, `getPartitions`, `doGetSplits`, and `doGetQueryPassthroughSchema` all borrow from the same pool and are equally exposed to a stale connection after a Lambda freeze/resume. If the retry is worth having, it should cover all of them, not just `doListSchemaNames`.

2. **The inline catch duplicates the response construction** (`new ListSchemasResponse(...)` appears twice), which is what makes it read awkwardly.

**Suggested cleaner approach — one reusable helper, applied uniformly:**

```java
@FunctionalInterface
private interface SqlSupplier<T> { T get() throws Exception; }

/**
 * Executes a JDBC-backed action, retrying once on Vertica error 100023, which indicates a
 * stale pooled connection reused after a Lambda freeze/resume. The retry borrows a new
 * connection, so it must be paired with the try-with-resources cleanup above (a leaked/unclosed
 * connection would just be re-borrowed in the same broken state).
 */
private <T> T withStaleConnectionRetry(SqlSupplier<T> action) throws Exception
{
    try {
        return action.get();
    }
    catch (SQLException e) {
        if (e.getErrorCode() != VERTICA_STALE_CONNECTION_ERROR) {
            throw e;
        }
        logger.warn("Stale connection detected (errorCode={}), retrying once with a fresh connection",
                VERTICA_STALE_CONNECTION_ERROR);
        return action.get();
    }
}
```

Call sites then stay flat with no duplicated response construction, and every connection-using method gets the same guard:

```java
public ListSchemasResponse doListSchemaNames(BlockAllocator allocator, ListSchemasRequest request) throws Exception
{
    logger.info("doListSchemaNames: {}", request.getCatalogName());
    return withStaleConnectionRetry(() ->
            new ListSchemasResponse(request.getCatalogName(), fetchSchemaNames()));
}

public GetTableResponse doGetTable(BlockAllocator allocator, GetTableRequest request) throws Exception
{
    return withStaleConnectionRetry(() -> {
        Set<String> partitionCols = new HashSet<>();
        try (Connection connection = getJdbcConnectionFactory().getConnection(getCredentialProvider())) {
            Schema schema = verticaSchemaUtils.buildTableSchema(connection, request.getTableName());
            return new GetTableResponse(request.getCatalogName(), request.getTableName(), schema, partitionCols);
        }
    });
}
```

The important part is that the retry's success depends on the connection actually being closed and returned before it re-borrows — so it only works layered on top of the try-with-resources fixes in this PR. Worth noting `"RowDescription"` desync is classically caused by reusing a connection with an unclosed statement/ResultSet, so once the leaks are fixed the condition may largely disappear on its own; the retry is then a nice defense-in-depth for the freeze/resume edge rather than the primary fix.

Minor: in `doGetSplits`, now that the whole body is inside the `try`, the trailing `catch (Exception e)` labels *any* checked exception `"connection failed: ..."`, which can be misleading for a non-connection failure deeper in the method. The `catch (RuntimeException e) { throw e; }` above it correctly preserves the `testAccess` / `executeQueriesOnVertica` exceptions, so that part's fine — just the checked-exception message that's a touch broad.

ritiktrianz:
Hi @ejeffrli , thanks for the review. As suggested, We have updated the changes to:  
1. Extract a reusable `withStaleConnectionRetry` helper (`SqlSupplier`).  
2. Apply it uniformly to `doListSchemaNames`, `doGetTable`, `getPartitions`, `doGetSplits`, and `doGetQueryPassthroughSchema`.  
3. Keep try‑with‑resources as the primary leak fix; the retry is defense‑in‑depth for the Lambda freeze/resume edge case.  
4. Soften the `doGetSplits` checked‑exception message to “Failed to get splits: …”.

PR #3489, opened 2026-06-24
Add paginated ListSchemas support for BigQuery connector

VenkatasivareddyTR:
*Issue #, if available:*

*Description of changes:*
Adds pagination support for schema (dataset) listing in the BigQuery federation connector, aligned with the existing ListTables pagination pattern.

**Changes:**
Extended federation SDK ListSchemas request/response with nextToken and pageSize
Introduce PaginatedSchemaLister for schema list pagination implementation and added serde V7.

**Behavior:**
Legacy requests (no pagination fields): Unchanged; retrieves all datasets using iterateAll() with a MAX_RESULTS guard
Paginated requests: Returns a single page of datasets along with a nextToken for subsequent calls


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListSchemasResponse.java
```diff
@@ -92,12 +117,13 @@ public boolean equals(Object o)
 
         return CollectionsUtils.equals(this.schemas, that.schemas) &&
                 Objects.equal(this.getRequestType(), that.getRequestType()) &&
-                Objects.equal(this.getCatalogName(), that.getCatalogName());
```
chngpe, line 95: We should add a serde V"7" to include this not modify at V2.. we need to bump the sdk version....

athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
```diff
@@ -138,27 +138,36 @@ public ListSchemasResponse doListSchemaNames(BlockAllocator blockAllocator, List
     {
         logger.info("doListSchemaNames called with Catalog: {}", listSchemasRequest.getCatalogName());
 
+        String nextToken = null;
```
chngpe: We should introduce an interface into based sdk or jdbc-sdk, with a based implementation.

https://github.com/awslabs/aws-athena-query-federation/blob/master/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/manager/JdbcMetadataHandler.java#L215-L270

each connector then complete the interface to support "pagination"

chngpe (changes requested):

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/lambda/metadata/ListSchemasResponse.java
VenkatasivareddyTR: Hi @chngpe, thank you for your review and feedback. Addressed as follows:
- Reverted SerDe V2 changes.
- Added SerDe V7 for paginated ListSchemas.
- Bumped SerDeVersion.SERDE_VERSION to 7 and wired V7 into VersionedObjectMapperFactory.

athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryMetadataHandler.java
VenkatasivareddyTR: Hi @chngpe, Addressed with a hook-based pagination contract in the federation SDK:

- Added PaginatedSchemaLister with default implementations for:
       - executeListSchemaNames — unlimited vs paginated branching (pageSize == -1 and nextToken == null)
       - listPaginatedSchemas — default in-memory pagination via PaginationHelper.manualSchemasPagination.

- BigQueryMetadataHandler implements PaginatedSchemaLister:
       - listSchemas — unlimited path (iterateAll())
       - listPaginatedSchemas — overridden for native BigQuery dataset pagination (pageToken / pageSize from the BigQuery API)

JDBC schema pagination in JdbcMetadataHandler is not in this PR (not required for BigQuery); it can follow the same listPaginatedSchemas hook pattern in a follow-up.

Issue #2409, opened 2024-11-19
[FEATURE] Support advanced logging controls for Lambda functions


sergei-ivanov:
**Is your feature request related to a problem? If yes, please describe.**

Lambda function [configuration](https://docs.aws.amazon.com/lambda/latest/dg/monitoring-cloudwatchlogs-advanced.html) now allows setting log level and switch between plain text and JSON logging. But unfortunately it's not currently being handled by the federation connectors. If I configure JSON logging, only Lambda runtime picks up that configuration. The connector still logs in plain text.

**Describe the solution you'd like**

Update Log4j2 configuration for all connectors according to the [AWS guide](https://docs.aws.amazon.com/lambda/latest/dg/java-logging.html#java-logging-log4j2).

For example, the [logging configuration for Athena connector](https://github.com/awslabs/aws-athena-query-federation/blob/258e1df8dc3cbbeee41780a7d259a34301a63b58/athena-postgresql/src/main/resources/log4j2.xml#L22-L37) should be changed to:

```xml
<Configuration packages="com.amazonaws.services.lambda.runtime.log4j2">
    <Appenders>
        <Lambda name="Lambda" format="${env:AWS_LAMBDA_LOG_FORMAT:-TEXT}">
            <LambdaTextFormat>
                <PatternLayout>
                    <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1} - %m%n </pattern>
                </PatternLayout>
            </LambdaTextFormat>
            <LambdaJSONFormat>
                <JsonTemplateLayout eventTemplateUri="classpath:LambdaLayout.json" />
            </LambdaJSONFormat>
        </Lambda>
    </Appenders>
    <Loggers>
        <Logger name="com.amazonaws.athena.connector.lambda" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-warn}" />
        <Logger name="com.amazonaws.athena.connectors.postgresql" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-warn}" />
        <Root level="${env:AWS_LAMBDA_LOG_LEVEL:-INFO}">
            <AppenderRef ref="Lambda" />
        </Root>
    </Loggers>
</Configuration>
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

macohen:
@sergei-ivanov would you be up to submitting this as a PR? good idea.

PR #3494, opened 2026-06-30
Support advanced logging controls for Lambda functions

anushatrianz:
*Issue #, if available:* 
[#2409](https://github.com/awslabs/aws-athena-query-federation/issues/2409)

*Description of changes:* 
Adds support for AWS Lambda Advanced Logging Controls so connector logs follow Lambda logging settings.

**Changes:**
Updated connector Log4j2 configurations to use the AWS Lambda Log4j2 appender.
Added support for Lambda log format selection through: AWS_LAMBDA_LOG_FORMAT
1. TEXT
2. JSON

Added support for Lambda log level configuration through: AWS_LAMBDA_LOG_LEVEL (controls the root logging level for the Lambda function)
ATHENA_FEDERATION_SDK_LOG_LEVEL controls the logging level for Athena Federation SDK and connector-specific loggers.

Preserved backward compatibility with existing text-based logging behavior.

Please find attached analysis document.
[Support_Advanced_Logging_Controls_For_Lambda_Functions.docx](https://github.com/user-attachments/files/30588298/Support_Advanced_Logging_Controls_For_Lambda_Functions.docx)

### DIFF
--- a/athena-clickhouse/src/main/resources/log4j2.xml
+++ b/athena-clickhouse/src/main/resources/log4j2.xml
@@ -21,16 +21,21 @@
 
 <Configuration packages="com.amazonaws.services.lambda.runtime.log4j2">
     <Appenders>
-        <Lambda name="Lambda">
-            <PatternLayout>
-                <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1}:%L - %m%n</pattern>
-            </PatternLayout>
+        <Lambda name="Lambda" format="${env:AWS_LAMBDA_LOG_FORMAT:-TEXT}">
+            <LambdaTextFormat>
+                <PatternLayout>
+                    <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1} - %m%n</pattern>
+                </PatternLayout>
+            </LambdaTextFormat>
+            <LambdaJSONFormat>
+                <JsonTemplateLayout eventTemplateUri="classpath:LambdaLayout.json" />
+            </LambdaJSONFormat>
         </Lambda>
     </Appenders>
     <Loggers>
         <Logger name="com.amazonaws.athena.connector.lambda" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-warn}" />
         <Logger name="com.amazonaws.athena.connectors.mysql" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-warn}" />
-        <Root level="info">
+        <Root level="${env:AWS_LAMBDA_LOG_LEVEL:-INFO}">
             <AppenderRef ref="Lambda" />
         </Root>
     </Loggers>
--- a/athena-clickhouse/src/main/resources/log4j2.xml
+++ b/athena-clickhouse/src/main/resources/log4j2.xml
@@ -21,7 +21,23 @@
 
 <Configuration packages="com.amazonaws.services.lambda.runtime.log4j2">
     <Appenders>
+        <Lambda name="LambdaDebug" format="${env:AWS_LAMBDA_LOG_FORMAT:-TEXT}">
+            <Filters>
+                <LevelRangeFilter minLevel="DEBUG" maxLevel="TRACE" onMatch="ACCEPT" onMismatch="DENY"/>
+            </Filters>
+            <LambdaTextFormat>
+                <PatternLayout>
+                    <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1}:%L - %m%n</pattern>
+                </PatternLayout>
+            </LambdaTextFormat>
+            <LambdaJSONFormat>
+                <JsonTemplateLayout eventTemplateUri="classpath:LambdaLayout.json" />
+            </LambdaJSONFormat>
+        </Lambda>
         <Lambda name="Lambda" format="${env:AWS_LAMBDA_LOG_FORMAT:-TEXT}">
+            <Filters>
+                <ThresholdFilter level="INFO" onMatch="ACCEPT" onMismatch="DENY"/>
+            </Filters>
             <LambdaTextFormat>
                 <PatternLayout>
                     <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1} - %m%n</pattern>
@@ -36,6 +52,7 @@
         <Logger name="com.amazonaws.athena.connector.lambda" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-warn}" />
         <Logger name="com.amazonaws.athena.connectors.mysql" level="${env:ATHENA_FEDERATION_SDK_LOG_LEVEL:-warn}" />
         <Root level="${env:AWS_LAMBDA_LOG_LEVEL:-INFO}">
+            <AppenderRef ref="LambdaDebug" />
             <AppenderRef ref="Lambda" />
         </Root>
     </Loggers>
--- a/athena-jdbc/pom.xml
+++ b/athena-jdbc/pom.xml
@@ -122,6 +122,12 @@
             <version>${log4j2Version}</version>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-layout-template-json</artifactId>
+            <version>${log4j2Version}</version>
+            <scope>runtime</scope>
+        </dependency>
         <!-- https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-redshift -->
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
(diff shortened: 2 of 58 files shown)


athena-clickhouse/src/main/resources/log4j2.xml
```diff
@@ -21,16 +21,21 @@
...
 <Configuration packages="com.amazonaws.services.lambda.runtime.log4j2">
     <Appenders>
-        <Lambda name="Lambda">
-            <PatternLayout>
-                <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1}:%L - %m%n</pattern>
-            </PatternLayout>
+        <Lambda name="Lambda" format="${env:AWS_LAMBDA_LOG_FORMAT:-TEXT}">
+            <LambdaTextFormat>
+                <PatternLayout>
+                    <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1} - %m%n</pattern>
```
fal-bharadwaj, line 43: is this "%L" removal intentional? let's not remove it as its a good indicator of where the logline is at and that's how we've had it all this time (unless there's a good reason to get rid of it).

Let's apply this change to all connectors.

athena-jdbc/pom.xml
```diff
@@ -122,6 +122,12 @@
             <version>${log4j2Version}</version>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
```
fal-bharadwaj, line 125: the dependency change here is sufficient for all other connectors that use jdbc correct? (like mysql, oracle, clickhouse etc or do they need specific updates in pom as well?)

fal-bharadwaj (changes requested):

athena-clickhouse/src/main/resources/log4j2.xml
anushatrianz: > is this "%L" removal intentional? let's not remove it as its a good indicator of where the logline is at and that's how we've had it all this time (unless there's a good reason to get rid of it).
> 
> Let's apply this change to all connectors.

Hi @fal-bharadwaj, thanks for the review.
                                                                                                                                                                                          %L requires Log4j2 to resolve caller-location information for each log event, which is expensive in performance-sensitive environments ([see the Log4j2 performance guidance](https://logging.apache.org/log4j/2.x/manual/async.html#Location)).

We retained %c{1} to preserve the logger/class name and removed %L to avoid that overhead. The proposed solution for issue #2409 also does not include %L, so this removal is consistent with that approach and may have been intentional.

if you prefer to keep %L for easier debugging and consistency with the previous pattern, I’m happy to add it back. Please let me know your preference.

athena-jdbc/pom.xml
anushatrianz: > the dependency change here is sufficient for all other connectors that use jdbc correct? (like mysql, oracle, clickhouse etc or do they need specific updates in pom as well?)

Yes. JDBC connectors (MySQL, Oracle, ClickHouse, etc.) depend on athena-jdbc, so the pom update is sufficient. We also validated this during testing, and the details are included in the analysis document attached to the PR description.

athena-clickhouse/src/main/resources/log4j2.xml
anushatrianz: > > is this "%L" removal intentional? let's not remove it as its a good indicator of where the logline is at and that's how we've had it all this time (unless there's a good reason to get rid of it).
> > Let's apply this change to all connectors.
> 
> Hi @fal-bharadwaj, thanks for the review. %L requires Log4j2 to resolve caller-location information for each log event, which is expensive in performance-sensitive environments ([see the Log4j2 performance guidance](https://logging.apache.org/log4j/2.x/manual/async.html#Location)).
> 
> We retained %c{1} to preserve the logger/class name and removed %L to avoid that overhead. The proposed solution for issue #2409 also does not include %L, so this removal is consistent with that approach and may have been intentional.
> 
> if you prefer to keep %L for easier debugging and consistency with the previous pattern, We are happy to add it back. Please let us know your preference.

fal-bharadwaj: Thanks for taking a look, the performance improvement makes sense since the %L doing a caller-location stack walk on every event is costly. #2409 though is a feature request and its pattern is just an example, not a spec. Since this will change an existing pattern in logs (a bit of a regression even if small) I'm okay with removing it but erring on the side of caution @AbdulR3hman @chngpe thoughts?

AbdulR3hman: ok I've confirmed few things; 


from here [location cost](https://logging.apache.org/log4j/2.x/manual/async.html#Location):

> [Computing the location information (i.e., the caller class, method, file, and line number) of a log event is an expensive operation.](https://logging.apache.org/log4j/2.x/manual/layouts.html#LocationInformation) The impact on asynchronous loggers and appenders is even higher since the component must decide whether to compute it or not before crossing the asynchronous barrier. Location information is turned off by default for asynchronous loggers and appenders. To enable it for a certain logger, set its [includeLocation](https://logging.apache.org/log4j/2.x/manual/configuration.html#logger-attributes-includeLocation) attribute to true.

and from the [Log4j2 Pattern Layout manual ](https://logging.apache.org/log4j/2.x/manual/pattern-layout.html) it states: 

>"Capturing the source location information to generate the line number of the caller is an expensive operation" — and it "is not garbage-free."

so what we could do is enable it for debug mode:

```
<Appenders>
    <!-- DEBUG/TRACE: carry %L, pay stack walk only when debug is on -->
    <Lambda name="LambdaDebug" format="${env:AWS_LAMBDA_LOG_FORMAT:-TEXT}">
        <LambdaTextFormat>
            <PatternLayout>
                <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1}:%L - %m%n</pattern>
            </PatternLayout>
        </LambdaTextFormat>
        <LevelRangeFilter minLevel="DEBUG" maxLevel="TRACE" onMatch="ACCEPT" onMismatch="DENY"/>
    </Lambda>

    <!-- INFO and above: no %L, never pays -->
    <Lambda name="LambdaMain" format="${env:AWS_LAMBDA_LOG_FORMAT:-TEXT}">
        <LambdaTextFormat>
            <PatternLayout>
                <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1} - %m%n</pattern>
            </PatternLayout>
        </LambdaTextFormat>
        <ThresholdFilter level="INFO" onMatch="ACCEPT" onMismatch="DENY"/>
    </Lambda>
</Appenders>
```

AbdulR3hman: the above is only an example; we'd have to test.

AbdulR3hman (changes requested):

athena-clickhouse/src/main/resources/log4j2.xml
anushatrianz: > ok I've confirmed few things;
> 
> from here [location cost](https://logging.apache.org/log4j/2.x/manual/async.html#Location):
> 
> > [Computing the location information (i.e., the caller class, method, file, and line number) of a log event is an expensive operation.](https://logging.apache.org/log4j/2.x/manual/layouts.html#LocationInformation) The impact on asynchronous loggers and appenders is even higher since the component must decide whether to compute it or not before crossing the asynchronous barrier. Location information is turned off by default for asynchronous loggers and appenders. To enable it for a certain logger, set its [includeLocation](https://logging.apache.org/log4j/2.x/manual/configuration.html#logger-attributes-includeLocation) attribute to true.
> 
> and from the [Log4j2 Pattern Layout manual ](https://logging.apache.org/log4j/2.x/manual/pattern-layout.html) it states:
> 
> > "Capturing the source location information to generate the line number of the caller is an expensive operation" — and it "is not garbage-free."
> 
> so what we could do is enable it for debug mode:
> 
> ```
> <Appenders>
>     
>     <Lambda name="LambdaDebug" format="${env:AWS_LAMBDA_LOG_FORMAT:-TEXT}">
>         <LambdaTextFormat>
>             <PatternLayout>
>                 <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1}:%L - %m%n</pattern>
>             </PatternLayout>
>         </LambdaTextFormat>
>         <LevelRangeFilter minLevel="DEBUG" maxLevel="TRACE" onMatch="ACCEPT" onMismatch="DENY"/>
>     </Lambda>
> 
>     
>     <Lambda name="LambdaMain" format="${env:AWS_LAMBDA_LOG_FORMAT:-TEXT}">
>         <LambdaTextFormat>
>             <PatternLayout>
>                 <pattern>%d{yyyy-MM-dd HH:mm:ss} %X{AWSRequestId} %-5p %c{1} - %m%n</pattern>
>             </PatternLayout>
>         </LambdaTextFormat>
>         <ThresholdFilter level="INFO" onMatch="ACCEPT" onMismatch="DENY"/>
>     </Lambda>
> </Appenders>
> ```

Hi @AbdulR3hman, thanks for the suggestion and for confirming the cost associated with location information.
We've implemented the dual-pattern approach as suggested:

- **DEBUG/TRACE** use a pattern that includes %L (line number).
- **INFO and higher levels** use a lighter pattern without %L.

We tested the changes and confirmed they work as expected. We've also updated the analysis document with the test results at the end.
Please review and let us know if you have any further feedback.

PR #3526, opened 2026-07-15
athena-synapse: managed FAS credentials, Substrait rendering

smanikoz:
*Issue #, if available:* N/A

*Description of changes:*

Adds Athena query-federation (Substrait) support and Glue-managed-connector credential handling to the Azure Synapse connector (`athena-synapse`), and fixes a read failure on serverless SQL pools.

### Summary

1. **Substrait query-plan rendering.** When a request carries a Substrait query plan (`Constraints.getQueryPlan() != null`), `JdbcSplitQueryBuilder` renders SQL through the Calcite `SqlDialect` returned by `getSqlDialect()`. `athena-synapse` did not override it and fell back to `AnsiSqlDialect`, producing ANSI SQL for a T-SQL engine. This adds `SynapseDialect` (Calcite `MssqlSqlDialect` with `[ ]` identifier quoting, catalog-casing aware) and overrides `getSqlDialect()` / `getSqlDialect(boolean)`. Azure Synapse is accessed over the Microsoft SQL Server JDBC driver (port 1433) and speaks T-SQL, so `MssqlSqlDialect` with `[ ]` quoting is the correct dialect.

2. **Catalog casing filter for Substrait.** `SynapseMetadataHandler.doGetSplits` propagates the `CATALOG_CASING_FILTER` config property into each split so the casing-aware dialect is selected during Substrait rendering.

3. **Managed-connector credentials (FAS).** `SynapseRecordHandler` / `SynapseMetadataHandler` forward the request's `AwsRequestOverrideConfiguration` via `getCredentialProvider(getRequestOverrideConfig(request))` (and `makeEncryptionKey(...)`), enabling Federated Identity (FAS) request-override credentials and customer-KMS spill in managed-connector mode. This includes the metadata `getSchema` path.

4. **Serverless `@@TRANCOUNT` fix.** `readWithConstraint` set `connection.setAutoCommit(false)` unconditionally. On an Azure Synapse **serverless (on-demand)** SQL pool — which does not support transactions — running reads with `autoCommit=false` manifests as `"'@@TRANCOUNT' is not supported."`, failing serverless reads. The fix detects the serverless environment once (`SynapseUtil.checkEnvironment(url) == SynapseConstants.SQL_POOL`) and guards **both** `setAutoCommit(false)` and `commit()`: on serverless the connection stays at `autoCommit=true` and `commit()` is skipped, while result streaming (driven by the configured statement fetch size) is unaffected. Dedicated SQL pools (which support transactions) are unchanged.

### Notably NOT changed: LIMIT pushdown

`SynapseQueryStringBuilder.appendLimitOffset` still returns an empty string (no LIMIT pushdown). This is deliberate: Azure Synapse **dedicated** SQL pool does not support the `OFFSET/FETCH` clause, so pushing `FETCH` down would fail against a dedicated pool; the engine applies the limit.

### Changes

- **New** `athena-synapse/.../synapse/SynapseDialect.java` — Calcite `MssqlSqlDialect`, catalog-casing-aware identifier quoting.
- `athena-synapse/.../synapse/SynapseQueryStringBuilder.java` — override `getSqlDialect()` / `getSqlDialect(boolean)`.
- `athena-synapse/.../synapse/SynapseMetadataHandler.java` — propagate `CATALOG_CASING_FILTER` in `doGetSplits`; FAS request-override credentials + customer-KMS spill.
- `athena-synapse/.../synapse/SynapseRecordHandler.java` — serverless-guarded `setAutoCommit(false)` / `commit()` (`@@TRANCOUNT` fix); FAS request-override credentials.
- **New** `athena-synapse/.../synapse/SynapseDialectTest.java`.
- `athena-synapse/.../synapse/SynapseRecordHandlerTest.java`, `SynapseMetadataHandlerTest.java` — serverless/dedicated and request-override credential cases.

### Testing

- `mvn -pl athena-synapse -am -DskipTests package` — build succeeds.
- `mvn -pl athena-synapse test` on Corretto 17 — unit tests pass, including the new `SynapseDialectTest` and the serverless/dedicated `SynapseRecordHandlerTest` cases.
- End-to-end: verified in a gamma environment that a serverless Synapse read (previously failing with `'@@TRANCOUNT' is not supported.`) succeeds after the fix, including filter/order-by pushdown.

### Notes

- No `pom.xml` / SDK-version changes: the shared `athena-jdbc` and `athena-federation-sdk` already provide the Substrait machinery; this change wires the connector-specific dialect, split property, credential forwarding, and serverless autoCommit guard.
- Rebased onto the latest `master`.

### DIFF
--- a/athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseDialect.java
+++ b/athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseDialect.java
@@ -1,0 +1,53 @@
+/*-
+ * #%L
+ * athena-synapse
+ * %%
+ * Copyright (C) 2019 - 2026 Amazon Web Services
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package com.amazonaws.athena.connectors.synapse;
+
+import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.dialect.MssqlSqlDialect;
+
+import java.util.Locale;
+
+/**
+ * Azure Synapse-specific SQL dialect with catalog casing filter support. Synapse uses the Microsoft SQL Server
+ * (T-SQL) engine and JDBC driver, so it renders Substrait-derived query plans with the Calcite MSSQL dialect and
+ * uses square brackets ({@code []}) for identifier quoting.
+ */
+public class SynapseDialect extends MssqlSqlDialect
+{
+    public static final SqlDialect DEFAULT = MssqlSqlDialect.DEFAULT;
+
+    private final boolean catalogCasingFilter;
+
+    public SynapseDialect(boolean catalogCasingFilter)
+    {
+        super(DEFAULT_CONTEXT);
+        this.catalogCasingFilter = catalogCasingFilter;
+    }
+
+    @Override
+    public StringBuilder quoteIdentifier(StringBuilder buf, String identifier)
+    {
+        if (catalogCasingFilter) {
+            String upper = identifier.toUpperCase(Locale.ROOT);
+            return buf.append("[").append(upper.replace("]", "]]")).append("]");
+        }
+        return super.quoteIdentifier(buf, identifier);
+    }
+}
--- a/athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
+++ b/athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
@@ -387,7 +387,7 @@ public class SynapseMetadataHandler extends JdbcMetadataHandler
         SchemaBuilder schemaBuilder;
         HashMap<String, List<String>> columnNameAndDataTypeMap = new HashMap<>();
 
-        try (Connection connection = getJdbcConnectionFactory().getConnection(getCredentialProvider());
+        try (Connection connection = getJdbcConnectionFactory().getConnection(getCredentialProvider(requestOverrideConfiguration));
              PreparedStatement stmt = connection.prepareStatement(dataTypeQuery)) {
             // fetch data types of columns and prepare map with column name and datatype information.
             stmt.setString(1, tableName.getSchemaName() + "." + tableName.getTableName());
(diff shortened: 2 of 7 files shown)


ZhitongYan (dismissed):

athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
```diff
@@ -311,16 +312,22 @@ public GetSplitsResponse doGetSplits(BlockAllocator blockAllocator, GetSplitsReq
...
             }
             else {
-                splitBuilder = Split.newBuilder(spillLocation, makeEncryptionKey())
+                splitBuilder = Split.newBuilder(spillLocation, makeEncryptionKey(getRequestOverrideConfig(getSplitsRequest)))
                         .add(PARTITION_NUMBER, partInfo);
             }
+
+            if (getSplitsRequest.getIdentity().getConfigOptions() != null && getSplitsRequest.getIdentity().getConfigOptions().containsKey(EnvironmentConstants.CATALOG_CASING_FILTER)) {
+                LOGGER.info("Catalog Casing Filter found: {}", getSplitsRequest.getIdentity().getConfigOptions().get(EnvironmentConstants.CATALOG_CASING_FILTER));
+                splitBuilder.add(EnvironmentConstants.CATALOG_CASING_FILTER, getSplitsRequest.getIdentity().getConfigOptions().get(EnvironmentConstants.CATALOG_CASING_FILTER));
```
DilipReddyGaddam: nit : create a constant for getSplitsRequest.getIdentity().getConfigOptions().get(EnvironmentConstants.CATALOG_CASING_FILTER) and use it

athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseDialect.java
```diff
@@ -0,0 +1,53 @@
...
+    public SynapseDialect(boolean catalogCasingFilter)
+    {
+        super(DEFAULT_CONTEXT);
+        this.catalogCasingFilter = catalogCasingFilter;
+    }
+
+    @Override
+    public StringBuilder quoteIdentifier(StringBuilder buf, String identifier)
+    {
+        if (catalogCasingFilter) {
```
DilipReddyGaddam, line 47: please make sure we have done testing this with lower and upper case values. catalogCasingFilter = false and true

athena-synapse/src/test/java/com/amazonaws/athena/connectors/synapse/SynapseDialectTest.java
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ * #L%
+ */
+package com.amazonaws.athena.connectors.synapse;
+
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class SynapseDialectTest
```
DilipReddyGaddam, line 26: Nit: Consider adding a test case for identifiers containing ] to verify the ]] escape in the casing path. This will replace("]", "]]") logic that prevents SQL injection through crafted identifiers.

athena-synapse/src/main/java/com/amazonaws/athena/connectors/synapse/SynapseMetadataHandler.java
smanikoz: Done in b6e0d9c5b — extracted the map into a local `identityConfigOptions` and the value into `catalogCasingFilter`, so the `getIdentity().getConfigOptions()` chain isn't repeated. Behavior-identical (same null-check → containsKey → get → split add).

athena-synapse/src/test/java/com/amazonaws/athena/connectors/synapse/SynapseDialectTest.java
smanikoz: Added `testQuoteIdentifierWithFilterEscapesClosingBracket` in b6e0d9c5b: input `emp]loyees` → `[EMP]]LOYEES]`, verifying the `replace("]", "]]")` escaping on the casing path.

DilipReddyGaddam (approved):

ZhitongYan (approved):

PR #3530, opened 2026-07-16
Fix issue where wrong SQL query generated for JDBC connectors when ca…

pearlswan:
Fix issue where SQL query were translated differently when customer configure catalog casing filter. 
The constructor with catalog casing filter produces default implementation instead of dialect specific implementation.

*Issue #, if available:*
Customer reported issue that SqlServer connector is translating query plan with limit operation into wrong query with "FETCH NEXT x.." instead of "TOP x"

*Description of changes:*


DilipReddyGaddam (dismissed):

ejeffrli:
Nice, targeted fix — re-parenting to the real Calcite dialects is the right call, since the old hierarchy dropped all the dialect-native `unparse` logic (LIMIT→`TOP`, etc.) and only kept `quoteIdentifier`.

One follow-up on **`JdbcCasingSqlDialect`**: after this change nothing in production extends it anymore. The four connector dialects (`MysqlSqlDialect`, `OracleSqlDialect`, `PostgresqlSqlDialect`, `SqlServerDialect`) now extend the corresponding `org.apache.calcite.sql.dialect.*` classes directly, so `athena-jdbc/.../manager/JdbcCasingSqlDialect.java` is only referenced by its own unit test (`JdbcCasingSqlDialectTest`). Leaving it in place is a bit of a trap — it reads like the shared base class for these dialects but is actually dead, and a future connector might extend it again and reintroduce exactly the bug this PR fixes.

Suggest deleting `JdbcCasingSqlDialect.java` (and `JdbcCasingSqlDialectTest.java`) in this PR so the abstraction doesn't linger.

A couple of smaller notes on the new casing override, carried over from the old implementation (not regressions, just worth tidying while it's fresh):
- The casing branch appends the quote chars raw (`[` + upper + `]`), so an identifier containing the close-quote char isn't escaped, whereas the `super.quoteIdentifier(...)` branch does escape it. The two branches now differ in quoting safety.
- `identifier.toUpperCase()` uses the default locale; `toUpperCase(Locale.ROOT)` avoids surprises (e.g. Turkish `i`).

And since the same structural change applies to all four dialects, it'd be good to add the SQL Server-style Substrait test for MySQL/Oracle/Postgres too, so the correct LIMIT/quoting output is locked in for each.

ZhitongYan (approved):

ejeffrli (approved):

PR #3547, opened 2026-07-23
make jdbc handler closeable to release resources

ZhitongYan:
*Description of changes:*

athena-query-federation handlers was designed to be single-tenant lambda handlers. For JDBC use cases, as long as the handlers are singleton objects, the underlying resources like Hikari pools will remain singleton. However when the lambda handler is used in multi-tenant scenarios, the handlers will be created on per request basis. As a result, after a long running time period of single AWS Lambda instance, multiple Hikari pools will be created with requests from different tenants, which causing OOM exceptions due to long running background threads managing the pools.

This change resolves this issue by making JDBC metadata and records handlers Autocloseable. All upstream callers of the handlers should managed the life cycle of athena-query-federation handlers by using try-with-resource or try finally blocks to eventually close the resources. As a results, all jvm resources used by SDK should be GCed correctly.

### DIFF
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/connection/GenericJdbcConnectionFactory.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/connection/GenericJdbcConnectionFactory.java
@@ -31,9 +31,6 @@ import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.services.glue.model.ErrorDetails;
 import software.amazon.awssdk.services.glue.model.FederationSourceErrorCode;
 
-import java.io.UnsupportedEncodingException;
-import java.net.URLEncoder;
-import java.nio.charset.StandardCharsets;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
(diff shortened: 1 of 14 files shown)


athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/connection/GenericJdbcConnectionFactory.java
```diff
@@ -159,29 +136,71 @@ private Connection getPooledConnection(String jdbcUrl) throws SQLException
...
+    public void close() 
     {
-        try {
-            return URLEncoder.encode(value, StandardCharsets.UTF_8.toString());
-        }
-        catch (UnsupportedEncodingException ex) {
-            throw new AthenaConnectorException("Unsupported Encoding Exception: ",
-                    ErrorDetails.builder().errorCode(FederationSourceErrorCode.OPERATION_NOT_SUPPORTED_EXCEPTION.toString()).errorMessage(ex.getMessage()).build());
+        if (ds != null) {
+            ds.close();
```
DilipReddyGaddam: do we make ds = null after closing ? something like this 

`  @Override
   public void close() {
       synchronized (GenericJdbcConnectionFactory.class) {
           if (ds != null) {
               ds.close();
               ds = null;
           }
       }
   }`

ZhitongYan: Thanks for flagging — I looked at this and don't think we want to null ds out here.

close() is terminal in this design: the factory has a 1:1 lifetime with its pool, and the only callers (JdbcMetadataHandler.close() / JdbcRecordHandler.close()) invoke it when the handler is no longer needed. Nulling ds would turn close() into a reset — any use-after-close would silently spin up a new HikariCP pool instead of surfacing the lifecycle bug. Leaving it as-is means such a call fails loudly with Hikari's IllegalStateException: HikariDataSource has been closed., which I'd rather keep.

DilipReddyGaddam (dismissed):

ejeffrli (approved):

DilipReddyGaddam (approved):

PR #3568, opened 2026-07-28
Athena-GCS Substrait support 

aashrith-madasu:
## Summary

Add Substrait plan support to the Athena GCS connector, enabling filter pushdown and limit pushdown directly within the connector. This allows Athena to push query predicates and LIMIT clauses down to the GCS record reader, reducing the number of rows materialized and improving query performance.

### Substrait Filter Pushdown (`SubstraitRowFilter`)
- Added a new `SubstraitRowFilter` class in `athena-federation-sdk` that parses Substrait `FilterRel` expressions from the query plan and evaluates them row-by-row against Arrow `VectorSchemaRoot` batches.
- Supports the full logical expression tree: `AND`, `OR`, `NOT`, as well as comparison operators (`=`, `!=`, `>`, `>=`, `<`, `<=`), `IS NULL`, `IS NOT NULL`, `NOT IN`, `NOR`, and `NAND`.
- Falls back gracefully to a no-op filter (pass all rows) if no Substrait plan is present or parsing fails.

### Substrait Limit Pushdown
- `GcsRecordHandler` now extracts `FetchRel` (LIMIT) from the Substrait plan when no `SortRel` (ORDER BY) is present, enabling early termination of file reads once the limit is reached.
- TOP-N pushdown is intentionally not supported to avoid costly in-memory sorting of large datasets.

### Data Source Capabilities
- `GcsMetadataHandler` now advertises `SUPPORTS_LIMIT_PUSHDOWN` (integer constant) and `SUPPORTS_COMPLEX_EXPRESSION_PUSHDOWN` (all `StandardFunctions`) via `doGetDataSourceCapabilities`.

### Credential Handling Refactored
- Moved GCS credential installation (`installCaCertificate`, `installGoogleCredentialsJsonFile`, `setupNativeEnvironmentVariables`) from `GcsCompositeHandler` into `GcsRecordHandler`'s constructor, so credentials are available at record-read time independent of handler composition.
- Introduced `GcsUtil.initDatasource()` factory method that creates `StorageMetadata` on-demand using `CachableSecretsManager`, removing the need to eagerly resolve credentials in `GcsMetadataHandler`'s constructor.
- Using the new config key `secret_arn` (`GCS_CREDS_SECRET_ARN`) instead of `secret_manager_gcp_creds_name`.

### Storage Listing Fix
- `StorageMetadata.getStorageSplits()` and `getPartitionFolders()` now ensure a trailing `/` on the GCS prefix to avoid matching unrelated blobs with overlapping prefixes.
- Simplified partition folder retrieval — currently returns all partition folders without constraint-based pruning (partition filtering via Substrait is deferred to future work).

### Assembly / Packaging
- Updated `athena-gcs/assembly.xml` to output dependencies into a `lib/` directory instead of unpacking them into the root. This avoids classpath conflicts and simplifies the Lambda deployment artifact.

### Build Compatibility
- Added a `java21+` Maven profile in the root `pom.xml` with `--add-opens` JVM flags required for Arrow/Netty/ByteBuddy on JDK 21+.

## Testing
- Added `GcsRecordHandlerTest` with comprehensive Substrait plan tests covering:
  - Equality, inequality, greater-than, less-than filters
  - AND/OR/NOT compound expressions
  - LIMIT pushdown
  - No-filter passthrough when Substrait plan is absent
- Added `GcsSubstraitPlanGenerator` test utility for programmatically building Substrait plans
- Added CSV test fixtures (`MOCK_DATA_male.csv`, `MOCK_DATA_female.csv`)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitRowFilter.java
```diff
@@ -0,0 +1,387 @@
...
+ * If the Substrait plan is not available or parsing fails, the filter passes all rows (no filtering).
+ */
+public class SubstraitRowFilter
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(SubstraitRowFilter.class);
+
+    private final LogicalExpression logicalExpression;
+    private final boolean hasFilter;
+
+    private SubstraitRowFilter(LogicalExpression logicalExpression, boolean hasFilter)
```
suyinlee: are we not supporting casing mode?

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitRowFilter.java
```diff
@@ -0,0 +1,387 @@
...
+        }
+        return true;
+    }
+
+    /**
+     * Evaluates a NOR predicate against the vector schema root.
+     * NOR means NOT(p1 OR p2 OR p3) — the row passes only if NONE of the children match.
+     * Each child predicate may reference its own column.
+     */
+    @SuppressWarnings("unchecked")
```
suyinlee: why unchecked here

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitRowFilter.java
```diff
@@ -0,0 +1,387 @@
...
+            }
+        }
+        return true;
+    }
+
+    /**
+     * Evaluates a NAND predicate against the vector schema root.
+     * NAND means NOT(p1 AND p2 AND p3) — the row passes if NOT ALL children match.
+     */
+    @SuppressWarnings("unchecked")
```
suyinlee: what is the reason to put @SuppressWarnings("unchecked")

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitRowFilter.java
```diff
@@ -0,0 +1,387 @@
...
+        String indent = "  " + "  ".repeat(depth);
+        if (expr == null) {
+            LOGGER.info("{}(null)", indent);
+            return;
+        }
+        if (expr.isLeaf()) {
+            ColumnPredicate pred = expr.getLeafPredicate();
+            LOGGER.info("{}LEAF: {} {} '{}' [type={}]",
+                indent, pred.getColumn(), pred.getOperator(), pred.getValue(),
+                pred.getArrowType() != null ? pred.getArrowType() : "unknown");
```
suyinlee: what is the case is would go to unknown?

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsRecordHandler.java
```diff
@@ -92,6 +100,15 @@ protected GcsRecordHandler(S3Client amazonS3, SecretsManagerClient secretsManage
     {
         super(amazonS3, secretsManager, amazonAthena, SOURCE_TYPE, configOptions);
         this.invoker = ThrottlingInvoker.newDefaultBuilder(EXCEPTION_FILTER, configOptions).build();
+
+        setupNativeEnvironmentVariables();
```
suyinlee, line 99: what is this used for

pom.xml
```diff
@@ -577,6 +577,43 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
```
suyinlee, line 580: why do we need this?

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
```diff
@@ -280,4 +286,27 @@ private int decodeContinuationToken(GetSplitsRequest request)
         //No continuation token present
         return 0;
     }
+
+    /*
+     * Currently only supports Filter and Limit pushdown.
+     * Does not support TOP-N pushdown as sorting huge number of rows in-memory can be costly.
+     */
+    @Override
+    public GetDataSourceCapabilitiesResponse doGetDataSourceCapabilities(BlockAllocator allocator, GetDataSourceCapabilitiesRequest request)
```
suyinlee: If the oroginal constraint path doesn't have doGetDataSourceCapabilities, how did it decided which filter are supported?

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsRecordHandler.java
```diff
@@ -149,17 +180,56 @@ protected void readWithConstraint(BlockSpiller spiller, ReadRecordsRequest recor
...
-                        // We will loop on batch records and consider each records to write in spiller.
+                        // We will loop on batch records and consider each records to write in spiller.                 
                         for (int rowIndex = 0; rowIndex < root.getRowCount(); rowIndex++) {
+                            if (rowFilter.hasFilter() && !rowFilter.evaluate(root, rowIndex)) {
+                                continue;
+                            }
                             // we are passing record to spiller to be written.
                             execute(spiller, invoker.invoke(root::getFieldVectors), rowIndex, partitionColumns, split);
+                            numRowsWritten++;
+                            if (limit > 0 && numRowsWritten >= limit) {
```
suyinlee: why would numRowsWritten ever > limit? If numRowsWritten > limit, doesn't it means that writing too many data?

pom.xml
suyinlee, line 580: why do you need to add java21 here

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsUtil.java
```diff
@@ -97,10 +97,13 @@ public static void installCaCertificate() throws IOException, NoSuchAlgorithmExc
      * Install/place Google cloud platform credentials from AWS secret manager to temp location
      * This is required for dataset api
      */
-    public static void installGoogleCredentialsJsonFile(java.util.Map<String, String> configOptions) throws IOException
+    public static void installGoogleCredentialsJsonFile(java.util.Map<String, String> configOptions, CachableSecretsManager secretsManager) throws IOException
     {
-        CachableSecretsManager secretsManager = new CachableSecretsManager(SecretsManagerClient.create());
-        String gcsCredentialsJsonString = secretsManager.getSecret(configOptions.get(GCS_SECRET_KEY_ENV_VAR));
+        if (configOptions.get(GCS_CREDS_SECRET_ARN) == null) {
```
suyinlee: in which case the secret arn can be null

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitRowFilter.java
aashrith-madasu: Will remove this and implement type checking in the method

athena-federation-sdk/src/main/java/com/amazonaws/athena/connector/substrait/SubstraitRowFilter.java
aashrith-madasu: Will remove this and implement type checking

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
aashrith-madasu: The original connector was not supported in Glue. After migration, I have added GetDataSourceCapabilities.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsRecordHandler.java
aashrith-madasu: numRowsWritten will never be greater than limit, this condition is essentially the same as numRowsWritten == limit.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsRecordHandler.java
aashrith-madasu: this function sets environment vars for
1. Google service account credentials file path
2. SSL certificate

pom.xml
aashrith-madasu: this adds a java21 profile for developers using jdk21 in their local system with exrtra add-opens for libraries -Mockito, Netty, Arrow

aashrith-madasu: Some libraries (Mockito's ByteBuddy, Arrow, Netty) need additional java internals which were restricted by JDK21. Hence, added a profile for java21+ adding opens to them:
java.base/sun.nio.ch
java.base/jdk.internal.misc
java.base/sun.misc
java.base/java.lang.reflect
java.base/jdk.internal.ref
java.base/java.lang.invoke

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsUtil.java
aashrith-madasu: this will only be null for the instance passed to AthenaConnectorCapabilities. But that wont be a problem because only doGetDataSourceCapabilities method is called and GCS_CREDS_SECRET_ARN key is not used.

athena-gcs/src/main/java/com/amazonaws/athena/connectors/gcs/GcsMetadataHandler.java
aashrith-madasu: removed filter-pushdown capability and only supporting limit pushdown

PR #3602, opened 2026-08-14
Add connection retry for Aurora Serverless v2 cold-start failures

Wenting-glue:
### Problem
When connecting to Aurora Serverless v2 with 0 ACU minimum capacity, the first connection attempt fails with 'Failed to initialize pool: The connection attempt failed' because the database is paused and needs ~25-45 seconds to resume.

*Refer to SIM Ticket - https://t.corp.amazon.com/V2153573163*

### Changes
Add retry logic (3 attempts, 15s backoff) in GenericJdbcConnectionFactory.getConnection() for transient connection failures. Only retries on:
- 'failed to initialize pool' (HikariCP pool init failure)
- 'connect timed out' (TCP timeout during database resume)

Non-retryable errors (auth failures, invalid host, etc.) fail without retry. 

### Testing
- Reproduced the error with gamma endpoint
- Deployed the fix to gamma VPC lambda and verified the issue is gone.

### DIFF
--- a/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/connection/GenericJdbcConnectionFactory.java
+++ b/athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/connection/GenericJdbcConnectionFactory.java
@@ -101,15 +101,67 @@ public class GenericJdbcConnectionFactory
         }
     }
 
+    private static final int MAX_CONNECTION_RETRIES = 3;
+    private static final long CONNECTION_RETRY_BACKOFF_MS = 15_000; // 15 seconds
+
     @Override
     public Connection getConnection(final CredentialsProvider credentialsProvider)
             throws Exception
     {
-        if (useDirectConnection) {
-            return createDirectConnection(credentialsProvider);
-        }
+        Connection connection = null;
+        Exception lastException = null;
 
-        return getConnectionFromManagedPool(credentialsProvider);
+        for (int attempt = 1; attempt <= MAX_CONNECTION_RETRIES; attempt++) {
+            try {
+                if (useDirectConnection) {
+                    connection = createDirectConnection(credentialsProvider);
+                }
+                else {
+                    connection = getConnectionFromManagedPool(credentialsProvider);
+                }
+                if (attempt > 1) {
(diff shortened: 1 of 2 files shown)


athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/connection/GenericJdbcConnectionFactory.java
```diff
@@ -101,15 +101,67 @@ public GenericJdbcConnectionFactory(final DatabaseConnectionConfig databaseConne
...
+            }
+            catch (Exception e) {
+                lastException = e;
+                if (attempt == MAX_CONNECTION_RETRIES || !isRetryableConnectionError(e)) {
+                    throw e;
+                }
+                LOGGER.warn("Connection attempt {} of {} failed: {}. Retrying in {}ms...",
+                        attempt, MAX_CONNECTION_RETRIES, e.getMessage(), CONNECTION_RETRY_BACKOFF_MS);
+                try {
+                    Thread.sleep(CONNECTION_RETRY_BACKOFF_MS);
```
ZhitongYan, line 135: 15 seconds is too long for metadata / query operations. Is it possible to handle the retry at a higher level (i.e. connector callers)?

athena-jdbc/src/main/java/com/amazonaws/athena/connectors/jdbc/connection/GenericJdbcConnectionFactory.java
```diff
@@ -101,15 +101,67 @@ public GenericJdbcConnectionFactory(final DatabaseConnectionConfig databaseConne
...
+    /**
+     * Determines if an exception represents a transient connection failure,
+     * typically caused by Aurora Serverless v2 resuming from a paused state (0 ACU).
+     */
+    private boolean isRetryableConnectionError(Throwable e)
+    {
+        Throwable current = e;
+        while (current != null) {
+            String message = current.getMessage();
+            if (message != null && (message.toLowerCase().contains("failed to initialize pool")
```
ZhitongYan, line 158: this might not be generic for only name matching. Same as above, maybe we can add retry logic in the connector callers instead inside connectors.

PR #3608, opened 2026-08-18
Fix BigQuery NameResolver Issue in Base Docker Image

anushatrianz:
*Issue #, if available:*

*Description of changes:*
While testing the BigQuery connector with the base Docker image, we identified a runtime issue:
**Delegate resolver not found, scheme: dns**

This occurs because Java SPI files under `META-INF/services/` can be overwritten when multiple archives contain files at the same path, resulting in the loss of service provider definitions.

This change updates the Docker image build process to **merge SPI service definition files** instead of overwriting them. As a result, all required service provider registrations are preserved, including the gRPC `NameResolverProvider` entries required by the BigQuery connector.

**Validation**
We tested the following connectors in addition to BigQuery to ensure that the fix does not impact other connectors:

- GCS
- DynamoDB
- Vertica
- Oracle

The results confirmed that the fix resolves the BigQuery issue without impacting other connectors.

Please refer to the attached functional testing and analysis document for additional details.
[BigQuery_NameResolver_Fix_For_Base_Docker_Image.docx](https://github.com/user-attachments/files/31178530/BigQuery_NameResolver_Fix_For_Base_Docker_Image.docx)
[BIGQUERY_FUNCTIONAL_TEST_2026-08-18.xlsx](https://github.com/user-attachments/files/31179184/BIGQUERY_FUNCTIONAL_TEST_2026-08-18.xlsx)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,8 +89,9 @@ RUN for file in ${LAMBDA_TASK_ROOT}/*.jar ${LAMBDA_TASK_ROOT}/*.zip; do \
         for svc in /tmp/extract/META-INF/services/*; do \
           [ -f "$svc" ] || continue; \
           base=$(basename "$svc"); \
-          cat "$svc" >> "${LAMBDA_TASK_ROOT}/META-INF/services/${base}" && \
-          sort -u "${LAMBDA_TASK_ROOT}/META-INF/services/${base}" -o "${LAMBDA_TASK_ROOT}/META-INF/services/${base}"; \
+          dest="${LAMBDA_TASK_ROOT}/META-INF/services/${base}"; \
+          touch "$dest"; \
+          awk '!seen[$0]++' "$dest" "$svc" > "$dest.tmp" && mv "$dest.tmp" "$dest"; \
         done && \
         rm -rf /tmp/extract/META-INF/services; \
       fi && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,7 +78,23 @@ COPY \
   athena-vertica/target/athena-vertica-${JAR_VERSION}.jar \
   ${LAMBDA_TASK_ROOT}/
 
-# Run a shell loop to iterate over all jar/zip files and extract each one.
+# Unpack all connector jars/zips into the same task root.
+# Merge META-INF/services so SPI providers from each artifact are preserved.
 RUN for file in ${LAMBDA_TASK_ROOT}/*.jar ${LAMBDA_TASK_ROOT}/*.zip; do \
-      jar xf "$file" && rm -f "$file"; \
+      [ -e "$file" ] || continue; \
+      mkdir -p /tmp/extract && \
+      (cd /tmp/extract && jar xf "$file") && \
+      if [ -d /tmp/extract/META-INF/services ]; then \
+        mkdir -p ${LAMBDA_TASK_ROOT}/META-INF/services && \
+        for svc in /tmp/extract/META-INF/services/*; do \
+          [ -f "$svc" ] || continue; \
+          base=$(basename "$svc"); \
+          cat "$svc" >> "${LAMBDA_TASK_ROOT}/META-INF/services/${base}" && \
+          sort -u "${LAMBDA_TASK_ROOT}/META-INF/services/${base}" -o "${LAMBDA_TASK_ROOT}/META-INF/services/${base}"; \
+        done && \
+        rm -rf /tmp/extract/META-INF/services; \
+      fi && \
+      cp -a /tmp/extract/. ${LAMBDA_TASK_ROOT}/ && \
+      rm -rf /tmp/extract && \
+      rm -f "$file"; \
 done


Dockerfile
```diff
@@ -78,7 +78,23 @@ COPY \
   athena-vertica/target/athena-vertica-${JAR_VERSION}.jar \
   ${LAMBDA_TASK_ROOT}/
 
-# Run a shell loop to iterate over all jar/zip files and extract each one.
+# Unpack all connector jars/zips into the same task root.
+# Merge META-INF/services so SPI providers from each artifact are preserved.
 RUN for file in ${LAMBDA_TASK_ROOT}/*.jar ${LAMBDA_TASK_ROOT}/*.zip; do \
-      jar xf "$file" && rm -f "$file"; \
+      [ -e "$file" ] || continue; \
```
AbdulR3hman, line 84: Nice fix — the root-cause analysis is spot on, and merging `META-INF/services` at the base-image layer is the right call since the flatten-into-`/var/task` step is what breaks the normal per-jar SPI isolation. Two things on the merge implementation itself that I'd want addressed before this lands, both in the same `RUN` step:

**1. (blocking) `cat >> …` can fuse two providers onto one line.**
`META-INF/services` files aren't guaranteed to end in a newline. If an earlier-appended file has no trailing `\n`, its last entry concatenates with the next archive's first entry:

```
# BigQuery file (no trailing newline):  io.grpc.internal.DnsNameResolverProvider
# GCS file appended after it:           com.google.cloud.grpc.GcpNameResolverProvider
# result in the merged file:
io.grpc.internal.DnsNameResolverProvidercom.google.cloud.grpc.GcpNameResolverProvider
```

`sort -u` then treats that as one "unique" line, and `ServiceLoader` throws `ServiceConfigurationError` trying to load a class by that fused name — dropping *both* providers. Since BigQuery and GCS both contribute to `io.grpc.NameResolverProvider`, this can hit the exact file the PR is fixing.

**2. `sort -u` alphabetically reorders providers.**
gRPC is fine here (it orders resolvers by `priority()`, not file order), but generic `ServiceLoader.iterator()` returns providers in file order, and some SPIs are "first one wins" (e.g. SLF4J's `SLF4JServiceProvider`). Alphabetizing can silently change which implementation is selected. Better to preserve first-seen order.

**Suggested change** — replace the `cat >> … ; sort -u` pair with a single `awk` that fixes both. `awk` resets records at each file boundary (so it never joins lines) and re-emits each line with a proper `\n`, while `!seen[$0]++` preserves order and de-dupes:

```sh
for svc in /tmp/extract/META-INF/services/*; do
  [ -f "$svc" ] || continue
  base=$(basename "$svc")
  dest="${LAMBDA_TASK_ROOT}/META-INF/services/${base}"
  touch "$dest"
  awk '!seen[$0]++' "$dest" "$svc" > "$dest.tmp" && mv "$dest.tmp" "$dest"
done
```

_(Minor: `ServiceLoader` already ignores duplicate class entries per spec, so the must-fix is really the newline safety + preserving the union — the dedup is just tidiness.)_

AbdulR3hman (changes requested):

Dockerfile
anushatrianz: > Nice fix — the root-cause analysis is spot on, and merging `META-INF/services` at the base-image layer is the right call since the flatten-into-`/var/task` step is what breaks the normal per-jar SPI isolation. Two things on the merge implementation itself that I'd want addressed before this lands, both in the same `RUN` step:
> 
> **1. (blocking) `cat >> …` can fuse two providers onto one line.** `META-INF/services` files aren't guaranteed to end in a newline. If an earlier-appended file has no trailing `\n`, its last entry concatenates with the next archive's first entry:
> 
> ```
> # BigQuery file (no trailing newline):  io.grpc.internal.DnsNameResolverProvider
> # GCS file appended after it:           com.google.cloud.grpc.GcpNameResolverProvider
> # result in the merged file:
> io.grpc.internal.DnsNameResolverProvidercom.google.cloud.grpc.GcpNameResolverProvider
> ```
> 
> `sort -u` then treats that as one "unique" line, and `ServiceLoader` throws `ServiceConfigurationError` trying to load a class by that fused name — dropping _both_ providers. Since BigQuery and GCS both contribute to `io.grpc.NameResolverProvider`, this can hit the exact file the PR is fixing.
> 
> **2. `sort -u` alphabetically reorders providers.** gRPC is fine here (it orders resolvers by `priority()`, not file order), but generic `ServiceLoader.iterator()` returns providers in file order, and some SPIs are "first one wins" (e.g. SLF4J's `SLF4JServiceProvider`). Alphabetizing can silently change which implementation is selected. Better to preserve first-seen order.
> 
> **Suggested change** — replace the `cat >> … ; sort -u` pair with a single `awk` that fixes both. `awk` resets records at each file boundary (so it never joins lines) and re-emits each line with a proper `\n`, while `!seen[$0]++` preserves order and de-dupes:
> 
> ```shell
> for svc in /tmp/extract/META-INF/services/*; do
>   [ -f "$svc" ] || continue
>   base=$(basename "$svc")
>   dest="${LAMBDA_TASK_ROOT}/META-INF/services/${base}"
>   touch "$dest"
>   awk '!seen[$0]++' "$dest" "$svc" > "$dest.tmp" && mv "$dest.tmp" "$dest"
> done
> ```
> 
> _(Minor: `ServiceLoader` already ignores duplicate class entries per spec, so the must-fix is really the newline safety + preserving the union — the dedup is just tidiness.)_

Hi @AbdulR3hman, thanks for the detailed review and the suggestion.
We've updated the Dockerfile to replace `cat >>` + `sort -u` with the recommended `awk`-based merge logic. We tested the changes with the BigQuery, GCS, and DynamoDB connectors and confirmed that they work as expected.
Please review the changes and let us know if you have any further feedback.

AbdulR3hman (approved):

aimethed (approved):

PR #3610, opened 2026-08-20
Enhance DB2/DB2 AS400 data-partition SQL with identifier quoting.

ritiktrianz:
*Issue #, if available:*

*Description of changes:*
Added quoting of the partitioning column in DATAPARTITIONNUM(...) for Athena DB2 and DB2 AS400 data-partition queries. 

### DIFF
--- a/athena-db2-as400/src/main/java/com/amazonaws/athena/connectors/db2as400/Db2As400QueryStringBuilder.java
+++ b/athena-db2-as400/src/main/java/com/amazonaws/athena/connectors/db2as400/Db2As400QueryStringBuilder.java
@@ -54,9 +54,9 @@ public class Db2As400QueryStringBuilder extends JdbcSplitQueryBuilder
             tableName.append(quote(catalog)).append('.');
         }
         if (!Strings.isNullOrEmpty(schema)) {
-            tableName.append(schema).append('.');
+            tableName.append(quote(schema)).append('.');
         }
-        tableName.append(table);
+        tableName.append(quote(table));
         return String.format(" FROM %s ", tableName);
     }
 
--- a/athena-db2-as400/src/main/java/com/amazonaws/athena/connectors/db2as400/Db2As400QueryStringBuilder.java
+++ b/athena-db2-as400/src/main/java/com/amazonaws/athena/connectors/db2as400/Db2As400QueryStringBuilder.java
@@ -72,8 +72,8 @@ public class Db2As400QueryStringBuilder extends JdbcSplitQueryBuilder
         String column = split.getProperty(Db2As400MetadataHandler.PARTITIONING_COLUMN);
         if (column != null) {
             LOGGER.debug("Fetching data using Partition");
-            //example query: select * from EMP_TABLE WHERE DATAPARTITIONNUM(EMP_NO) = 0
-            return Collections.singletonList(" DATAPARTITIONNUM(" + column + ") = " + split.getProperty(Db2As400MetadataHandler.PARTITION_NUMBER));
+            //example query: select * from EMP_TABLE WHERE DATAPARTITIONNUM("EMP_NO") = 0
+            return Collections.singletonList(" DATAPARTITIONNUM(" + quote(column) + ") = " + split.getProperty(Db2As400MetadataHandler.PARTITION_NUMBER));
         }
         else {
             LOGGER.debug("Fetching data without Partition");
--- a/athena-db2-as400/src/main/java/com/amazonaws/athena/connectors/db2as400/Db2As400QueryStringBuilder.java
+++ b/athena-db2-as400/src/main/java/com/amazonaws/athena/connectors/db2as400/Db2As400QueryStringBuilder.java
@@ -54,9 +54,9 @@ public class Db2As400QueryStringBuilder extends JdbcSplitQueryBuilder
             tableName.append(quote(catalog)).append('.');
         }
         if (!Strings.isNullOrEmpty(schema)) {
-            tableName.append(quote(schema)).append('.');
+            tableName.append(schema).append('.');
         }
-        tableName.append(quote(table));
+        tableName.append(table);
         return String.format(" FROM %s ", tableName);
     }
 
--- a/athena-db2-as400/src/test/java/com/amazonaws/athena/connectors/db2as400/Db2As400QueryStringBuilderTest.java
+++ b/athena-db2-as400/src/test/java/com/amazonaws/athena/connectors/db2as400/Db2As400QueryStringBuilderTest.java
@@ -38,8 +38,8 @@ public class Db2As400QueryStringBuilderTest {
     {
         Split split = Mockito.mock(Split.class);
         Db2As400QueryStringBuilder builder = new Db2As400QueryStringBuilder(QUOTE_CHARACTER);
-        Assert.assertEquals(" FROM \"default\".table ", builder.getFromClauseWithSplit("default", "", "table", split));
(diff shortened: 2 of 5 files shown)


athena-db2-as400/src/main/java/com/amazonaws/athena/connectors/db2as400/Db2As400QueryStringBuilder.java
```diff
@@ -72,8 +72,8 @@ protected List<String> getPartitionWhereClauses(Split split)
         String column = split.getProperty(Db2As400MetadataHandler.PARTITIONING_COLUMN);
         if (column != null) {
             LOGGER.debug("Fetching data using Partition");
-            //example query: select * from EMP_TABLE WHERE DATAPARTITIONNUM(EMP_NO) = 0
-            return Collections.singletonList(" DATAPARTITIONNUM(" + column + ") = " + split.getProperty(Db2As400MetadataHandler.PARTITION_NUMBER));
+            //example query: select * from EMP_TABLE WHERE DATAPARTITIONNUM("EMP_NO") = 0
+            return Collections.singletonList(" DATAPARTITIONNUM(" + quote(column) + ") = " + split.getProperty(Db2As400MetadataHandler.PARTITION_NUMBER));
```
AbdulR3hman, line 76: is there a utility from the driver itself we can use to quote? like how we did for postgresql? instead of relying on our own approach?

ritiktrianz: > is there a utility from the driver itself we can use to quote? like how we did for postgresql? instead of relying on our own approach?

Hi @AbdulR3hman ,
Thanks for the review. We verified this against the driver APIs.
For PostgreSQL, we used `org.postgresql.core.Utils.escapeIdentifier`, as it is a public static method provided by pgJDBC and documented in the official JDBC API:  
[pgJDBC Utils.escapeIdentifier](https://jdbc.postgresql.org/documentation/publicapi/org/postgresql/core/Utils.html#escapeIdentifier-java.lang.StringBuilder-java.lang.String-)

The IBM JCC and JT400 drivers used by `athena-db2` and `athena-db2-as400` do not provide an equivalent API for identifier quoting:

- **IBM JCC (DB2):** [DB2Connection interface – IBM Db2 12.1](https://www.ibm.com/docs/en/db2/12.1.x?topic=jdbc-db2connection-interface)  
- **JT400 (DB2 AS400):** [AS400JDBCConnectionImpl](https://jt400.sourceforge.net/doc/com/ibm/as400/access/AS400JDBCConnectionImpl.html)

Since neither driver exposes a similar quoting utility, we reused `JdbcSplitQueryBuilder.quote()`, which already handles quoting for catalog, schema, and table names within the same class.

athena-db2-as400/src/main/java/com/amazonaws/athena/connectors/db2as400/Db2As400QueryStringBuilder.java
```diff
@@ -54,9 +54,9 @@ protected String getFromClauseWithSplit(String catalog, String schema, String ta
             tableName.append(quote(catalog)).append('.');
         }
         if (!Strings.isNullOrEmpty(schema)) {
-            tableName.append(schema).append('.');
+            tableName.append(quote(schema)).append('.');
         }
-        tableName.append(table);
+        tableName.append(quote(table));
```
ejeffrli: Quoting `schema`/`table` here changes their matching semantics on Db2 for i. Unquoted identifiers are folded to uppercase and stored that way (`QSYS2.SYSTABLES`/`SYSSCHEMAS` return the folded name), so `FROM myschema.mytable` currently matches case-insensitively. Wrapping them in `"…"` makes them delimited identifiers that require an exact, case-sensitive match against the stored name.

That's safe for the DB2 (LUW) connector because `Db2MetadataHandler` installs `Db2JDBCCaseResolver`, which resolves schema/table to the exact catalog case (`SELECT TABNAME FROM SYSCAT.TABLES WHERE ... LOWER(TABNAME) = ?`) before the SQL is built. `Db2As400MetadataHandler` passes no case resolver (it uses the base `DefaultJDBCCaseResolver`), so the name can reach `getFromClauseWithSplit` in whatever case Athena hands over. If that's lowercase against an uppercase-stored table, quoting `"mytable"` would turn a currently-working query into `table not found`.

Note the partition-column and SELECT-column quoting isn't affected by this — those names are read straight from the catalog in their stored case (`SYSCOLUMNS.COLUMN_NAME`), so they're already exact.

Could you confirm what case schema/table arrive in for AS400 (an integ test against an uppercase-stored table, referenced in lowercase, would settle it), or add an AS400 case resolver mirroring `Db2JDBCCaseResolver`? Happy to be wrong if resolution already yields exact case.

ritiktrianz: > Quoting `schema`/`table` here changes their matching semantics on Db2 for i. Unquoted identifiers are folded to uppercase and stored that way (`QSYS2.SYSTABLES`/`SYSSCHEMAS` return the folded name), so `FROM myschema.mytable` currently matches case-insensitively. Wrapping them in `"…"` makes them delimited identifiers that require an exact, case-sensitive match against the stored name.
> 
> That's safe for the DB2 (LUW) connector because `Db2MetadataHandler` installs `Db2JDBCCaseResolver`, which resolves schema/table to the exact catalog case (`SELECT TABNAME FROM SYSCAT.TABLES WHERE ... LOWER(TABNAME) = ?`) before the SQL is built. `Db2As400MetadataHandler` passes no case resolver (it uses the base `DefaultJDBCCaseResolver`), so the name can reach `getFromClauseWithSplit` in whatever case Athena hands over. If that's lowercase against an uppercase-stored table, quoting `"mytable"` would turn a currently-working query into `table not found`.
> 
> Note the partition-column and SELECT-column quoting isn't affected by this — those names are read straight from the catalog in their stored case (`SYSCOLUMNS.COLUMN_NAME`), so they're already exact.
> 
> Could you confirm what case schema/table arrive in for AS400 (an integ test against an uppercase-stored table, referenced in lowercase, would settle it), or add an AS400 case resolver mirroring `Db2JDBCCaseResolver`? Happy to be wrong if resolution already yields exact case.

Hi @ejeffrli , thanks for the review.

This is a valid concern. On Db2 for i, quoted identifiers are case-sensitive, while unquoted names are folded to uppercase. Quoting the schema and table names in the FROM clause is safe in athena-db2 because Db2JDBCCaseResolver rewrites them to the catalog case first. However, Db2As400MetadataHandler uses DefaultJDBCCaseResolver (NONE), so Athena-provided names can reach the SQL unchanged.
Due to environment constraints, we cannot validate that behavior, so we have reverted the AS400 schema and table quoting in the FROM clause.

To avoid any potential case-resolution differences on AS400, the PR now only quotes the partitioning column in DATAPARTITIONNUM(...). That column name already comes from the catalog in its stored case, so quoting it does not affect name matching.

ejeffrli: Thanks. leaving comment here for future reference that this is not vulnerable to sql injection because the table and schema name here is resolved by metadata handler already

ejeffrli (approved):

AbdulR3hman (approved):

PR #3611, opened 2026-08-20
Fix NPE in BigQuery connector when STRUCT/RECORD field is null

suyinlee:
BigQueryUtils.getComplexObjectFromFieldValue() crashes with NPE when a BigQuery RECORD/STRUCT field has a null value. FieldValue.getRecordValue() calls Preconditions.checkNotNull internally, which throws when the value is null.

Add a null check at the top of getComplexObjectFromFieldValue() to return null early for null or isNull() field values.

Ticket: P457052656

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryUtils.java
+++ b/athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryUtils.java
@@ -387,6 +387,9 @@ public class BigQueryUtils
 
     public static Object getComplexObjectFromFieldValue(Field field, FieldValue fieldValue) throws ParseException
     {
+        if (fieldValue == null || fieldValue.isNull()) {
+            return null;
+        }
         Types.MinorType minorTypeForArrowType = getMinorTypeForArrowType(field.getFieldType().getType());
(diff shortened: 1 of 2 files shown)


athena-google-bigquery/src/main/java/com/amazonaws/athena/connectors/google/bigquery/BigQueryUtils.java
```diff
@@ -387,6 +387,9 @@ static Object getObjectFromFieldValue(String fieldName, FieldValue fieldValue, F
 
     public static Object getComplexObjectFromFieldValue(Field field, FieldValue fieldValue) throws ParseException
     {
+        if (fieldValue == null || fieldValue.isNull()) {
```
ZhitongYan, line 390: what is the diff between both null check here?

suyinlee: fieldValue == null checks the reference itself (no object at all — would NPE if we called .isNull() on it). fieldValue.isNull() checks that the object exists but wraps a SQL NULL value. The == null is first so short-circuit protects the .isNull() call

DilipReddyGaddam (approved):

ZhitongYan (approved):

